PR #28, opened 2025-11-14
Signpost and auto profiler

zoecarver:
This PR was written entirely autonomously by claude, I did not write any of this code. 

With this PR you can run any test/example normally after exporting `TTLANG_AUTO_PROFILE=1` and it will insert signposts around all ops, generate a flat buffer, pass it to `ttrt perf`, and then generate a nice little output to show how long each line took to run: 

Screenshot 2025-11-13 at 8 22 44 PM
Screenshot 2025-11-13 at 8 21 55 PM

### DIFF
--- a/lib/Dialect/TTL/Transforms/TTLDumpCBFlowGraph.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLDumpCBFlowGraph.cpp
@@ -14,10 +14,10 @@
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BuiltinAttributes.h"
-#include "ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.h"
 #include "ttlang/Dialect/TTL/IR/TTLOps.h"
 #include "ttlang/Dialect/TTL/IR/TTLOpsUtils.h"
 #include "ttlang/Dialect/TTL/Passes.h"
+#include "ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.h"
 
 #include "llvm/Support/JSON.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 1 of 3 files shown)


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03: Since this is for generating ttkernel -- would it be possible to have emitC be after in its own small TTLToEmitC pass? You can make the signpost op legal so it survives this pass. I think mixing more target dialects in this already huge pass is not great. Alternatively can add the op to ttkernel too at some point, should be straightforward.

python/ttlang/__init__.py
```diff
@@ -31,6 +32,7 @@
     "Semaphore",
     "CopyTransferHandler",
     "make_circular_buffer_like",
+    "signpost",
```
brnorris03: maybe as a todo, it would be good to have options (vs env vars) for some of these optional passes (including optimizations).

python/ttlang/ttl_api.py
```diff
@@ -70,6 +76,64 @@ def should_execute(self) -> bool:
...
+    # Dump profiler data from device
+    try:
+        ttnn.DumpDeviceProfiler(device)
+    except Exception as e:
+        print(f"[Auto-profile] Failed to dump device profiler: {e}")
+        return
+
+    # Find the profile CSV
+    profile_csv_path = os.environ.get(
+        "TTLANG_PROFILE_CSV", "/tmp/profile_log_device.csv"
```
brnorris03: should have a global tmpdir getter -- and suggest using `/tmp/$USER` for the directory because on qb or other shared machines, this breaks

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: Sure, I think that's a great point

PR #34, opened 2025-11-18
Add test/python to CI and configure ttrt to generate system desc

zoecarver:
- Add lit test execution to both CI workflows (call-build.yml and call-build-tt-mlir-fetchcontent.yml)
  - Generate system descriptor on Linux using ttrt query --save-artifacts
  - Include system descriptor file in repo for macOS builds (runtime not available on macOS)
  - Tests run after build with llvm-lit -sv test/python/

Note: won't work now, waiting for Boyana's TT HW config


zoecarver:
Boyana is taking this over

PR #54, opened 2025-12-05
[Discussion] `ttl` dialect proposal (plan)

brnorris03:
This draft PR is solely for discussion on a proposed `ttl` dialect (not intended to merge). See [TTL_Dialect_Plan.md](https://github.com/tenstorrent/tt-lang/blob/bnorris/ttl-dialect-plan/docs/TTL_Dialect_Plan.md)

### DIFF
--- a/docs/ttl/02_TTL_Type_System.md
+++ b/docs/ttl/02_TTL_Type_System.md
@@ -5,6 +5,60 @@
 This document specifies the type system for the TTL (TT-Lang) MLIR dialect,
 including core types, attributes, and TensorAccessor support strategy.
 
+## Implementation Status
+
+> **Legend**: ✅ Implemented | 🔄 Changed | ✨ Added | ⏳ In progress | 📋 Planned | ❌ Won't implement
+>
+> | Type/Feature | Status | Notes |
+> |--------------|--------|-------|
+> | **Types** |||
+> | `!ttl.cb` (CircularBuffer) | ✅🔄 | Implemented with mnemonic `cb` instead of `circular_buffer` |
+> | `!ttl.transfer_handle` | ✅ | With optional `<read>`/`<write>` direction |
+> | `!ttl.pipe` | 📋 | Inter-core communication |
+> | `!ttl.pipenet` | 📋 | Pipe container |
+> | `!ttl.semaphore` | 📋 | Synchronization primitive |
+> | `!ttl.accessor` | 📋 | TensorAccessor type |
+> | **Attributes** |||
+> | `#ttl.tensor_encoding` | 📋 | Currently uses `#ttnn.ttnn_layout` from tt-mlir |
+> | `#ttl.layout` | 📋 | Layout attribute |
+> | `#ttl.slice` | 📋 | For pipe dst_range |
+> | **Operations (Resource)** |||
+> | `ttl.bind_cb` | ✅✨ | Added: binds to existing CB slot (replaces `ttl.create_cb`) |
+> | `ttl.attach_cb` | ✅✨ | Added: associates tensor SSA value with CB |
+> | `ttl.create_cb` | ✅🔄 | Implemented as `ttl.bind_cb` |
+> | `ttl.create_pipe` | 📋 | Pipe creation |
+> | `ttl.create_semaphore` | 📋 | Semaphore creation |
+> | `ttl.tensor_accessor` | 📋 | TensorAccessor creation |
+> | **Operations (Data Movement)** |||
+> | `ttl.copy` | ✅🔄 | Simpler syntax (no `at [indices]`) |
+> | `ttl.wait` | ✅ | Wait on transfer handle |
+> | `ttl.dma_barrier` | 📋 | Global DMA barrier |
+> | **Operations (CB)** |||
+> | `ttl.cb_wait` | ✅ | Consumer acquire |
+> | `ttl.cb_pop` | ✅ | Consumer release |
+> | `ttl.cb_reserve` | ✅ | Producer acquire |
+> | `ttl.cb_push` | ✅ | Producer release |
+> | **Operations (Compute)** |||
+> | `ttl.compute` | ✅✨ | Added: structured compute with `indexing_maps`/`iterator_types` |
+> | `ttl.yield` | ✅✨ | Added: terminator for `ttl.compute` |
+> | `ttl.add`, `ttl.sub`, `ttl.mul`, `ttl.max` | ✅ | Binary elementwise (tensor-level) |
+> | `ttl.exp`, `ttl.log`, `ttl.sqrt`, `ttl.rsqrt` | ✅ | Unary elementwise (tensor-level) |
+> | `ttl.tanh`, `ttl.sigmoid`, `ttl.neg`, `ttl.abs`, `ttl.relu` | ✅ | Unary elementwise (tensor-level) |
+> | `ttl.tile_add`, `ttl.tile_sub`, `ttl.tile_mul`, `ttl.tile_max` | ✅✨ | Added: binary elementwise (tile-level) |
+> | `ttl.tile_exp`, `ttl.tile_log`, etc. | ✅✨ | Added: unary elementwise (tile-level) |
+> | `ttl.matmul` | 📋 | Matrix multiplication |
+> | `ttl.reduce_sum`, `ttl.reduce_max` | 📋 | Reductions |
+> | `ttl.bcast` | 📋 | Broadcast |
+> | `ttl.compute_region` | ✅🔄 | Implemented as `ttl.compute` |
+> | **Operations (Structural)** |||
+> | `ttl.program` | 📋 | Top-level kernel program |
+> | `ttl.kernel` | 📋 | Multi-thread kernel container |
+> | `ttl.compute_thread` | 📋 | Currently uses `func.func` with attr |
+> | `ttl.datamovement_thread` | 📋 | Currently uses `func.func` with attr |
+> | **Operations (Utility)** |||
+> | `ttl.core` | 📋 | Get core coordinates |
+> | `ttl.grid_size` | 📋 | Get grid dimensions |
+
 ## Table of Contents
 
 - [3.1 Core Types](#31-core-types)
@@ -77,6 +131,10 @@ def TTL_Tensor : TensorOf<[F32, F16, BF16]> {
 
 // TTL Types (TTLTypes.td)
 
+// 🔄 CHANGED: Implementation uses mnemonic "cb" instead of "circular_buffer"
+// Actual implementation in TTLOpsTypes.td:
+//   def TTL_CircularBuffer : TTL_Type<"CircularBuffer", "cb">
+//   assemblyFormat = "`<` `[` $shape `]` `,` $elementType `,` $bufferFactor `>`"
 def TTL_CircularBuffer : TTL_Type<"CircularBuffer", "circular_buffer"> {
   let summary = "Circular buffer for producer-consumer communication (L1 memory only)";
   let parameters = (ins
@@ -141,6 +199,7 @@ def TTL_CircularBuffer : TTL_Type<"CircularBuffer", "circular_buffer"> {
 }
 
 
+// ✅ IMPLEMENTED: TTLOpsTypes.td with optional direction parameter
 def TTL_TransferHandle : TTL_Type<"TransferHandle", "transfer_handle"> {
   let summary = "Handle for asynchronous transfer with transaction ID tracking";
   let description = [{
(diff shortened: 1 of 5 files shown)


docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+def TTL_TensorAccessor : TTL_Type<"TensorAccessor", "accessor"> {
+  let summary = "Handle for indexed tensor access";
+  let description = [{
+    Represents a tensor with indexing capability for slicing operations.
+    Carries layout metadata from python/ttlang/layouts.py (tiled, sharded, interleaved).
+    Used in DMA operations to specify source/destination with coordinates.
+  }];
+  let parameters = (ins
+    TTL_Tensor:$tensor,
+    TTL_LayoutAttr:$layout,              // Reuses create_metal_layout metadata
```
phizalev-TT: Should this layout be already present on `TTL_Tensor`?

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+    Each slice in dst_core_range is a tuple (start, stop, step) per TT-Lang spec.
+    The range is half-open: [start, stop). Step defaults to 1 if not specified.
+
+    Example from TT-Lang (multicast column):
+      ttl.Pipe(src_core=(x, 0), dst_core_range=(slice(x, x+1), slice(1, grid_y)))
+      Creates pipe from (x,0) to cores (x, 1), (x, 2), ..., (x, grid_y-1)
+  }];
+}
+
+def TTL_TensorAccessor : TTL_Type<"TensorAccessor", "accessor"> {
```
phizalev-TT: If we will have `TTKernel_TensorAccessor` I am not sure if `TTL_TensorAccessor` is needed.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+    # Tensors are sharded across 2x2 grid
+    # Each core processes 1 shard (1 tile in this example)
+
+    a_cb = ttl.make_circular_buffer_like(a, shape=(1, 1), buffer_factor=2)
+    b_cb = ttl.make_circular_buffer_like(b, shape=(1, 1), buffer_factor=2)
+    out_cb = ttl.make_circular_buffer_like(out, shape=(1, 1), buffer_factor=2)
+
+    @ttl.datamovement()
+    def dm_reader():
+        shard_id = ttl.core_linear()  # 0-3 for 2x2 grid
```
phizalev-TT: This should be `core(dim = 1)`

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+```
+
+#### 4.4.5 `DST` Register Management
+
+`DST` register operations are IR-level constructs for managing register
+allocation at the block abstraction level.
+
+Block-to-tile lowering context: TTL operations work on *blocks* (groups of
+tiles, e.g., a 2x1 block = 2 tiles). Hardware operations in TTKernel work on
+*individual tiles* (32x32 elements). During the `TTLLowerCompute` pass:
```
phizalev-TT: I think here we need to deal with block -> sub-block -> tile loop nest.

Say we do elwise chain (say 3 ops) on a block that that 8x8 tiles. 

Outer loop will iterate 8 times to have 8x1 sub-block in each iteration. 8x1 fits into DST.

Here we init elwise op 1.

Inner loop will iterate 8 times to have per-tile op 1 where each iteration targets different position in DST.

Here we uninit elwise op 1 and init elwise op 2.

Inner loop same as above for op 2.

Here we uninit elwise op 2 and init elwise op 3.

Inner loop same as above for op 3.

Here we uninit elwise op 3 and pack 8x1 sub-block to L1.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+- **Transform**:
+  - Tensor → CB: `ttkernel.tensor_accessor` +
+    `ttkernel.noc_async_read_shard/page/tile`
+  - CB → Tensor: `ttkernel.tensor_accessor` +
+    `ttkernel.noc_async_write_shard/page/tile`
+  - CB → Pipe → CB: `ttkernel.noc_async_write_multicast` or unicast
+  - Layout attribute determines which NOC API variant (shard/page/tile)
+- **Output**: TTKernel NOC operations using `TensorAccessor`
+
+### 5.3 Granularity and Block Shapes
```
phizalev-TT: I am not sure we should have granularity as a concept. It is entirely user-level.

brnorris03: Yeah, I struggled with it here... I will remove.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+   - More explicit and type-safe
+   - Aligns with TTKernel's requirement for exact tile counts
+
+3. **Python API**:
+   ```python
+   # Option A: Explicit shape (current)
+   cb = CircularBuffer(shape=(2, 1), buffer_factor=2)
+
+   # Option B: Kernel-level granularity parameter (future)
+   @ttl.kernel(granularity=2)  # Sets default for all CBs
```
phizalev-TT: I suggest we see if this really emerges as parameter everyone uses. Until then we should use explicit CB shapes.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+3. Port `examples/eltwise_add.py` to TTL
+4. Verify generated C++
+
+**Deliverable**: Working end-to-end TTL pipeline
+
+
+
+## 9. TTNN Runtime Integration
+
+### Runtime Architecture
```
phizalev-TT: Please see my points about TT-NN generic operation instead of dylib workflow.

docs/TTL_Dialect_Plan.md
brnorris03: Sounds good. I will leave it as a future maybe thing (or can remove entirely).

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+13. [References](#13-references)
+
+
+
+## Executive Summary
+
+The TTL dialect provides a tensor-level intermediate representation for TT-lang
+programs (defined in [TT-lang.md](TT-lang.md)), enabling multi-stage lowering
+with explicit compiler transformations before generating C++ kernels. Generated
+kernels can execute on either the TTNN runtime (via dylib workflow) or the
```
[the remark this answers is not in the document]
brnorris03: I guess I should have remembered/realized this. I was looking through the tt-metal docs for the best/easiest integration method, which gave me dylib, which is definitely worse for our purposes.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+
+### TTL Flow
+
+TTL supports two backend paths after TTKernel lowering:
+
+**Path 1: TTNN Dylib Workflow (Primary for MVP)**
+```
+Python Kernel → Python AST → TTL Dialect → TTL Passes → TTKernel → ConvertTTKernelToEmitC → C++ Source
+                                    ↓                                                            ↓
+                         Validation, Synchronization,                                    C++ Compiler
```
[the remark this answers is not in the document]
brnorris03: Great point! I am changing the runtime integration section completely and will update the workflows here.

docs/TTL_Dialect_Plan.md
```diff
@@ -0,0 +1,3793 @@
...
+| `ttl.cb_pop` | `ttkernel.cb_pop_front` | Low | **HIGH** |
+| `ttl.cb_reserve` | `ttkernel.cb_reserve_back` | Low | **HIGH** |
+| `ttl.cb_push` | `ttkernel.cb_push_back` | Low | **HIGH** |
+| `ttl.get_tile` | `ttkernel.copy_tile_init`, `ttkernel.copy_tile` | Medium | **HIGH** |
+| `ttl.pack_tile` | `ttkernel.pack_tile` | Low | **HIGH** |
+| **Compute (FPU)** ||||
+| `ttl.block_add` | `ttkernel.add_tiles_init`, `ttkernel.add_tiles` | Medium | **HIGH** |
+| `ttl.block_sub` | `ttkernel.sub_tiles_init`, `ttkernel.sub_tiles` | Medium | **HIGH** |
+| `ttl.block_mul` | `ttkernel.mul_tiles_init`, `ttkernel.mul_tiles` | Medium | **HIGH** |
+| `ttl.block_matmul` | `ttkernel.mm_init`, `ttkernel.matmul_tiles` | Medium | **HIGH** |
```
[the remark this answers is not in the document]
brnorris03: Yes, although that can also happen in a separate `ttkernel.matmul_tiles` -> `ttkernel.matmul_block` lowering. But this brings up a good point that if we do DST allocation in ttl (which I do want to do), then we need the equivalent op in `ttl`.

docs/TTL_Dialect_Plan.md
brnorris03: I decided to do this with linalg and tiling transform. I added some details in the plan, but will also add a separate example with more details.

docs/TTL_Dialect_Plan.md
brnorris03: This document is no longer being edited, it was split into more manageable parts in the docs/ttl directory.

docs/ttl/02_TTL_Type_System.md
```diff
@@ -0,0 +1,740 @@
...
+
+    Unicast: src=[x,y], dst=[x',y']
+    Multicast: src=[x,y], dst_range=[slice(x0,x1,step), slice(y0,y1,step)]
+
+    Each slice in dst_range is a tuple (start, stop, step) per TT-Lang spec.
+    The range is half-open: [start, stop). Step defaults to 1 if not specified.
+
+    Arity requirement: The dst_range tuple must have the same arity as the
+    grid rank to prevent ambiguity. For a 2D grid (grid_x, grid_y), both dimensions
+    must be specified explicitly. Use slice(x, x+1) for a single core in that dimension.
```
phizalev-TT: I think the language spec has a lesser constraint. Pipes within the same pipe net must have the same dimensionality. But this dimensionality can be arbitrary since we have this ability with `grid_size` and `core` functions. For example there can be 1D pipe net defined within a 2D grid.

docs/ttl/02_TTL_Type_System.md
```diff
@@ -0,0 +1,740 @@
...
+| Memory space | `ttcore::MemorySpaceAttr` | `TTCoreOpsEnums.td` |
+| Grid topology | `ttcore::GridAttr` | `TTCoreOpsAttrs.td` |
+| Tensor layout | `ttnn::LayoutAttr` | `TTNNOpsAttrs.td` |
+| Core ranges | `ttnn::CoreRangeSetAttr` | `TTNNOpsAttrs.td` |
+| Distribution strategy | `ttnn::TensorMemoryLayoutAttr` | `TTNNOpsEnums.td` |
+
+TTL-specific attributes are defined below:
+
+```tablegen
+def TTL_SliceAttr : AttrDef<TTL_Dialect, "Slice"> {
```
phizalev-TT: Do we have representation for slicing into tensor accessor?

docs/ttl/02_TTL_Type_System.md
```diff
@@ -0,0 +1,740 @@
...
+}
+```
+
+TTL lowering:
+
+```mlir
+// TTL IR (%tensor : tensor<..., #ttl.tensor_encoding<DeviceDRAM,
+//                      #ttl.layout<sharded, grid=[2,2]>>>)
+%accessor = ttl.tensor_accessor %tensor
+%xf = ttl.copy %accessor[%shard_id], %cb
```
phizalev-TT: Where this %shard_id comes from? Do we convert slices into it?

brnorris03: That's the high-level idea, but not completely sure about the syntax yet.

docs/ttl/02_TTL_Type_System.md
brnorris03: I will update to match.

docs/ttl/02_TTL_Type_System.md
brnorris03: Not at the moment, but ttnn tensors do support slice. We may need an extra op (that we define) to extract slices from the tensor accessor before it can be used as an arg in an other op.

phizalev-TT: Yes, TT-NN does have slicing, but I am referring to tensor slicing with ttl.copy. I guess we need a way to convert slices expression to shard id/page id in noc_async_xxx_shard/page.

PR #67, opened 2025-12-09
metal/tt-lang single and multi-core matmul

arichinsTT:
## Description
added metal and tt-lang (up to spec, not necessarily compiling) implementations of a single and multicore matmul that take in tilized inputs.
In particular looking to get feedback on tt-lang multicore implementation and how it handles work distribution.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Build/CI changes

## Testing
on team box
build metal, with ./build_metal.sh -c -e --build-dir build --development --cxx-compiler-path=clang++-18 --c-compiler-path=clang-18
create env (with metal script)
run pytests in ./examples/metal_examples/[single|multi]core_matmul/metal/[single|multi]core_matmul.py


examples/metal_examples/utils.py
```diff
@@ -0,0 +1,393 @@
...
+    )
+    assert ulp_passed, ulp_message
+    return ulp_passed, ulp_message
+
+    """
+Python implementation of get_large_matmul_params from bmm_op.hpp
+
+This module provides functions to compute optimal matrix multiplication parameters
+for distributing work across multiple cores on a Tenstorrent device.
+"""
```
brnorris03: nit: formatting, surprised black didn't fix

examples/metal_examples/utils.py
```diff
@@ -0,0 +1,393 @@
...
+    other_dim = (400 - 2 * in0_block_w * block_dim) // (2 * in0_block_w + block_dim)
+
+    if other_dim > 0:
+        return other_dim
+
+    return 0
+
+
+# Subblock choices in priority order
+SUBBLOCK_HW_CHOICES = [
```
brnorris03: how was this ordering determined? just curious
it does make sense since we want to use the cores more fully and keep it squarish, but interested why (4,2) is better than (2,4) -- would seem to be similar or at least dependent on the inputs shapes.

arichinsTT: Not sure, but I'd guess it was derived from either experimental testing in the llk's or from first principles of the hardware. tho really that one slipped by, meant to be part of next PR

examples/metal_examples/utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # hitting this oops
+    # if not allow_nonfinite and not torch.all(torch.isfinite(calculated)):
+    #     return False, "Calculated tensor contains non-finite values"
+
+    # if not _comp_nonfinite(golden, calculated):
+    #     return False, "Tensors are not finite at the same positions"
+    # nonfinite elements can interfere with ULP error calculation
+    # To avoid this, replace nan, +inf, -inf with 0
+    # (we have already checked that both tensors have the same nonfinite elements)
```
zoecarver: Can you explain this more clearly / not have commented out code?

examples/metal_examples/singlecore_matmul/ttlang/singlecore_matmul.py
```diff
@@ -0,0 +1,76 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
zoecarver: Does this test run with the sim? I think we should put a comment somewhere in here to make it very clear that this is a nice guiding example, but it's not expected to run yet.

examples/utils.py
```diff
@@ -112,3 +114,158 @@ def assert_allclose(
             f"Tensors not close: max_abs_error={max_abs_error:.6e}, "
             f"max_rel_error={max_rel_error:.6e}, rtol={rtol}, atol={atol}"
         )
+
+
+def ulp(x: torch.Tensor) -> torch.Tensor:
```
zoecarver: Can we import rather than duplicating the code? I think we have some other utils for asserting pretty close values, maybe we can unify all of them?

examples/metal_examples/singlecore_matmul/metal/kernels/mm_reader.cpp
```diff
@@ -0,0 +1,58 @@
...
+  const auto s0 = TensorAccessor(s0_args, src0_addr, get_tile_size(cb_id_in0));
+  constexpr auto s1_args =
+      TensorAccessorArgs<s0_args.next_compile_time_args_offset()>();
+  const auto s1 = TensorAccessor(s1_args, src1_addr, get_tile_size(cb_id_in1));
+
+  // Loop through the dimensions of the matrices. Read them and push to the
+  // circular buffers. Dimension names are called M, N and K. `t` in `mt` means
+  // tile.
+  for (uint32_t mt = 0; mt < Mt; mt++) {
+    uint32_t itileB = 0;
```
zoecarver: Not used?

examples/metal_examples/multicore_matmul/ttlang/multicore_matmul.py
```diff
@@ -0,0 +1,140 @@
...
+        f"core_grid: {core_grid}, num_output_tiles_total: {num_output_tiles_total}",
+        flush=True,
+    )
+    try:
+        (_, all_cores, core_group_1, core_group_2, work_per_core1, work_per_core2) = (
+            ttnn.split_work_to_cores(core_grid, num_output_tiles_total)
+        )
+    except Exception as e:
+        print(
+            f"Error splitting work to cores: {e}, trying again with row wise splitting"
```
zoecarver: In what cases do we fall back to this? Is this HW dependent? If one of these paths is always taken, can we refactor to just use that path?

examples/metal_examples/singlecore_matmul/ttlang/singlecore_matmul.py
arichinsTT: I haven't tried, but it's mentioned in the readme in metal_examples that the ttlang examples are no expected to run/compile. I'll add a comment at the top tho

zoecarver: Thanks!

examples/utils.py
arichinsTT: the spot where this exists in the metal repo made it difficult to import. would be interested in having unified numerical correctness util spot tho

zoecarver: Can we factor it into a separate module then? I'm confident we can share this utility with the right setup.

examples/metal_examples/multicore_matmul/ttlang/multicore_matmul.py
arichinsTT: just when the helper can't find a valid work split with it's internal constraints of divisibility (based on hw by the cores you pass in). I found that some shapes worked only on row-wise, so I added this here

zoecarver: OK interesting. Maybe we can have a TODO instead for when we can run this to investigate how to assign cores? I just feel like a try/catch with these two runtime functions that we can't test seems a little premature. Maybe leave one of them with a TODO or just the TODO? Or comment that we might want one of these variants?

examples/utils.py
arichinsTT: Like a helper available throughout the tt-lang repo? maybe a module in the tools directory of the repo?

examples/metal_examples/multicore_matmul/ttlang/multicore_matmul.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    x_size, y_size = ttl.grid_size(dims=2)
+    upper_bound_core = ttnn.CoreCoord(x_size - 1, y_size - 1)
+    core_grid = ttnn.CoreRangeSet(
+        [ttnn.CoreRange(ttnn.CoreCoord(0, 0), upper_bound_core)]
+    )
+    print(f"core_grid: {core_grid}, num_output_tiles_total: {num_output_tiles_total}")
+    (_, all_cores, core_group_1, core_group_2, work_per_core1, work_per_core2) = (
+        ttnn.split_work_to_cores(core_grid, num_output_tiles_total, row_wise=True)
```
phizalev-TT: Can we make TT-Lang specific variant of this function?

arichinsTT: We could, but I'm not really sure what benefit or variation could be provided by it being specific to tt-lang

arichinsTT: done!

examples/utils.py
arichinsTT: done!

examples/metal_examples/singlecore_matmul/ttlang/singlecore_matmul.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    @ttl.compute()
+    def mm_compute():
+        for _ in range(Mt):
+            for _ in range(Nt):
+                with out_cb.reserve() as out_blk:
+                    for _ in range(Kt):
+                        with a_cb.wait() as a_blk, b_cb.wait() as b_blk:
+                            out_blk.store(out_blk + a_blk @ b_blk)
```
phizalev-TT: This needs to be `out_blk.store(a_blk @ b_blk, acc = True)`

examples/utils.py
phizalev-TT: No material changes to this file?

Issue #78, opened 2025-12-12
[ttl] Emit real TTKernel CB handles instead of unrealized casts


brnorris03:
**Priority: HIGH**
**Parent Issue:** #77

## Description
Replace unrealized conversion casts with proper TTKernel CB handle materialization in the TTL to TTKernel lowering pass.

## Location
`lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp:182`

## Context
Currently using placeholder unrealized casts in `emitPlaceholderCB()`. Need to emit actual CB handles that the TTKernel dialect can work with.

## Related Subtasks
Part of the CB Handle Materialization group in #77.

PR #94, opened 2025-12-15
[ttl] Implement block lowering of ttl data movement threads

brnorris03:
## What?
Add multi-tile tensor DMA support to TTL→TTKernel lowering. When tensors exceed a single 32x32 tile, generate nested `affine.for` loops over the tile grid with correct linear offset computation for NOC async operations.

## Why?
The previous implementation only handled single-tile (32x32) tensor transfers. Any tensor larger than one tile would incorrectly DMA only the first tile repeatedly. Multi-tile tensors are essential for real workloads operating on non-trivial tensor shapes.

## How?
- Extract tile grid dimensions from TTNN layout's `memref<YxXx!ttcore.tile<32x32, f32>>` type.
- Add `emitTileLoop` helper function that generates nested `scf.for` loops for multi-tile cases or emits single-tile body for 1x1 grids.
- Compute linear tile offset inside loop body: `offset = iy * tilesX + ix` with proper index→i32 type conversion.
- Test cleanup: separate the operation tests from the conversion ones (they shouldn't have been together in the first place). Add more tests.

## Tests
- `llvm-lit test/ttlang/Conversion/TTLToTTKernel/dma_single_core.mlir` - 4 multi-tile test cases (2x2 and 3x2 tile grids)
- `llvm-lit test/ttlang/Translate/TTLToCpp/dma_single_read.mlir` - End-to-end C++ codegen verification

## TODO
- Add CB protocol operations (reserve, push, wait, pop) inside tile loop bodies once TTL dialect exposes these ops (Issue #78).
- Currently only NOC async read/write and barriers are emitted; proper CB lifecycle management is pending TTL→TTKernel lowering for CB ops. [#6320](https://github.com/tenstorrent/tt-mlir/pull/6320)
- Implement loop fusion for batched multi-tile DMA patterns: #95 

### Checklist:
*   [x] Self-reviewed (style, logic)
*   [x] Added tests (or justified none needed)
*   [ ] PR is small and focused (one task)

### DIFF
--- a/include/ttlang/Dialect/TTKernel/Transforms/TTKernelCleanupPatterns.h
+++ b/include/ttlang/Dialect/TTKernel/Transforms/TTKernelCleanupPatterns.h
@@ -1,0 +1,19 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTLANG_DIALECT_TTKERNEL_TRANSFORMS_TTKERNELCLEANUPPATTERNS_H
+#define TTLANG_DIALECT_TTKERNEL_TRANSFORMS_TTKERNELCLEANUPPATTERNS_H
+
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir::tt::ttkernel {
+
+/// Populate cleanup patterns for TTKernel ops. These patterns optimize
+/// TTKernel code by removing redundant operations (e.g., deduplicating
+/// consecutive barriers of the same type).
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -266,26 +333,37 @@ static LogicalResult lowerCBToTensor(CopyOp op, Value dstAccessor,
                                      const TypeConverter &typeConverter) {
   auto loc = op.getLoc();
 
+  int64_t tilesY = 1, tilesX = 1;
```
zoecarver: nit: helper function to share the logic with L301?

PR #97, opened 2025-12-15
[ttl] CB support (ops and ttkernel lowering)

zoecarver:
This PR adds circular buffer (CB) synchronization operations to the TTL dialect as a targeted, incremental step toward the full TTL-based kernel compilation path. It introduces four new ops (`ttl.cb_reserve`, `ttl.cb_push`, `ttl.cb_wait`, `ttl.cb_pop`). These ops lower to their TTKernel counterparts (ttkernel.cb_reserve_back, ttkernel.cb_push_back, ttkernel.cb_wait_front, ttkernel.cb_pop_front), which eventually emit to the hardware runtime APIs.

 The lowering uses `unrealized_conversion_cast` as a type-bridging mechanism during dialect conversion. Since `ttl.create_cb` produces a TTL CB type (`!ttl.cb<[shape], element_type, buffer_factor>`) but TTKernel ops require `!ttkernel.cb<total_elements, element_type>`, the conversion inserts casts to bridge these types. This pattern allows incremental lowering without requiring all type conversions to happen atomically. The CB ops with tensor results (reserve/wait) also cast the TTKernel CB back to a tensor view, maintaining SSA value flow for downstream consumers.

The `unrealized_conversion_cast` ops created during CB lowering are temporary scaffolding for bridging type differences between dialects (e.g., !ttl.cb → !ttkernel.cb). Lowering patterns trace through these casts to access the underlying TTKernel values, and once all TTL ops are converted to TTKernel, the casts become dead and a cleanup pass that will be introduced in the future (not yet up for review) simply erases any dead casts and placeholder ops, leaving clean TTKernel IR that proceeds to EmitC and finally C++ code generation.

### DIFF
--- a/test/ttlang/Conversion/TTLToTTKernel/cb_ops.mlir
+++ b/test/ttlang/Conversion/TTLToTTKernel/cb_ops.mlir
@@ -1,7 +1,11 @@
 // RUN: ttlang-opt --convert-ttl-to-ttkernel --canonicalize -cse --split-input-file %s | FileCheck %s
 
 // CHECK-LABEL: func.func @cb_reserve_single(
-// CHECK: ttkernel.cb_reserve_back
+// CHECK-DAG: %[[C1:.*]] = arith.constant 1 : i32
+// CHECK-DAG: %[[C0:.*]] = arith.constant 0 : i32
+// CHECK: %[[CB_TTL:.*]] = builtin.unrealized_conversion_cast %[[C0]] : i32 to !ttl.cb<[1, 1], f32, 2>
+// CHECK: %[[CB:.*]] = builtin.unrealized_conversion_cast %[[CB_TTL]] : !ttl.cb<[1, 1], f32, 2> to !ttkernel.cb<2, f32>
+// CHECK: ttkernel.cb_reserve_back(%[[CB]], %[[C1]]) : (!ttkernel.cb<2, f32>, i32) -> ()
 // CHECK: return
 module {
   func.func @cb_reserve_single() -> tensor<1x1xf32> attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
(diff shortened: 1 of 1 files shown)


brnorris03 (commented):
Overall looks great except for the number of pages arg to the new ops.

I think that it would be better if the new ops do not take an explicit `num_pages` argument, e.g., just have `ttl.cb_push()`, for example. All required information for computing the ttkernel inputs it is in the cb value produced by `ttl.create_cb`, so when lowering to ttkernel, you can find the defining op and compute that argument during the conversion, eliminating the need to do an extra pass to get rid of the unrealized conversion casts and simplifying the new `tt.cb_*` ops - no danger of passing inconsisten `num_pages` for example, no need to verify that arg, etc. WDYT?

What I am suggesting in a bit more detail:

Add a helper method to `CircularBufferType`:
```c++
// In TTLOpsTypes.td extraClassDeclaration
int64_t getNumPages() const {
  // num_pages = elements per block (tiles or scalars in one block)
  // This matches tt-mlir's DeviceAttr::getMemrefCBNumPages pattern:
  // num_pages = blockSizeBytes / pageSizeBytes
  // where pageSizeBytes = one tile size for tiled CBs
  return getElementsPerBlock();  // product(shape)
}
```
Then in lowering:
```c++
auto ttlCbTy = cast<CircularBufferType>(adaptor.getCb().getType());
int64_t numPages = ttlCbTy.getNumPages();  // = product(shape)
auto numPagesConst = rewriter.create<arith::ConstantIntOp>(loc, numPages, 32);
```

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -122,4 +122,101 @@ def TTL_WaitOp : TTL_Op<"wait", [MemoryEffects<[MemRead, MemWrite]>]> {
...
+    The num_pages operand must be a positive i32 value.
+
+    Example:
+    ```mlir
+    %view = ttl.cb_reserve %cb, %c1 : <[1, 1], !ttcore.tile<32x32, bf16>, 2> -> tensor<1x1x!ttcore.tile<32x32, bf16>>
+    ```
+  }];
+  let arguments = (ins
+    TTL_CircularBuffer:$cb,
+    I32:$num_pages
```
brnorris03: I left an overall comment about this, but I suggest to remove the `num_pages` arg from all cb ops and compute ttkernel-required values during lowering to ttkernel based on the cb shape, element type (and tile size); tt-mlir has such a helper that does something similar for memrefs (`getMemrefCBNumPages`). Alternatively, could extend the `!ttl.cb` type to include the number of pages so we only have to look that up (can be computed in the `ttl.create_cb` op builder).

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -149,6 +149,116 @@ struct CreateCBLowering : OpConversionPattern<CreateCBOp> {
...
+  // Create the corresponding TTKernel CB type.
+  Type tkCbTy =
+      ttk::CBType::get(ttlCbTy.getContext(), ttlCbTy.getTotalElements(),
+                       ttlCbTy.getElementType());
+  auto cast = rewriter.create<UnrealizedConversionCastOp>(loc, tkCbTy, cb);
+  return cast.getResult(0);
+}
+
+struct CBReserveLowering : OpConversionPattern<CBReserveOp> {
+  using OpConversionPattern::OpConversionPattern;
```
brnorris03: Most of thee new op conversions have a lot of common code, so it would be good to extract it into an abstract pattern templated over the specific CB op type (similar to `D2MCBOpRewriter`'s approach). For the ops with no return value, you can still use replace op in the common pattern implementation, e.g., something like `rewriter.replaceOp(op, newOp->getNumResults() > 0 ? newOp->getResult(0) : {});`.

brnorris03 (approved):
LGTM, thank you!!

PR #98, opened 2025-12-15
[ttl] kernel op support + regions to funcs

zoecarver:
Adds `ttl.kernel` as a container operation. The kernel op holds multiple regions: one per thread (datamovement or compute) that share circular buffer arguments for inter-thread communication. A new ttl-kernel-regions-to-funcs pass extracts each region into a standalone func.func with the appropriate `#ttkernel.thread` attribute, enabling downstream lowering to TTKernel.
```
  %result = ttl.kernel {grid = #ttcore.grid<1x1>, threads = [#ttl.thread<datamovement>, #ttl.thread<compute>]}
      ins(%lhs, %rhs : tensor<...>, tensor<...>)
      outs(%out : tensor<...>) {
    ^bb0(%lhs_cb: !ttl.cb<...>, %rhs_cb: !ttl.cb<...>, %out_cb: !ttl.cb<...>):
      // Data movement: read inputs into CBs
  }, {
    ^bb0(%lhs_cb: !ttl.cb<...>, %rhs_cb: !ttl.cb<...>, %out_cb: !ttl.cb<...>):
      // Compute: process data from CBs
  } : tensor<...>
```
 
The verifier enforces that all regions receive identical CB arguments and that thread count matches region count. After extraction, each region becomes a private function (e.g., @kernel_dm_0, @kernel_compute_1) and the kernel op's threads attribute is updated with symbol references to enable later dispatch.

### DIFF
--- a/include/ttlang/Dialect/TTL/IR/TTLOps.h
+++ b/include/ttlang/Dialect/TTL/IR/TTLOps.h
@@ -9,10 +9,10 @@
 #include "mlir/IR/Dialect.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
-#include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
 #include "ttlang/Dialect/TTL/IR/TTLOpsAttrs.h"
 #include "ttlang/Dialect/TTL/IR/TTLOpsEnums.h"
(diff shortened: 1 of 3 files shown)


zoecarver:
Instead of this, we are going to just emit funcs for each kernel and python can be responsible for collecting them together and assigning #ttkernel.thread attribute

PR #145, opened 2025-12-18
Update TTLTileUnaryToTTKernel and TTLTileBinaryToTTKernel patterns toemit copy_tile_init and copy_tile operations before the SFPU compute operations.

zoecarver:
Before
```
  ttkernel.exp_tile_init                                                                                                                           
  ttkernel.exp_tile(%dst_idx)                                                                                                                      
```                                                                                                                                                 

After
```
  ttkernel.copy_tile_init(%cb)                                                                                                                     
  ttkernel.copy_tile(%cb, %0, %dst_idx)                                                                                                            
  ttkernel.exp_tile_init                                                                                                                           
  ttkernel.exp_tile(%dst_idx)                                                                                                                      
```
                                                                                                                                                   
Before (binary):                                                                                                                              
```
  ttkernel.add_binary_tile_init                                                                                                                    
  ttkernel.add_binary_tile(%src0, %src1, %odst)                                                                                                    
```
```                                                                                                                                                   
After (binary):                                                                                                                               
  ttkernel.copy_tile_init(%lhs_cb)                                                                                                                 
  ttkernel.copy_tile(%lhs_cb, %0, %src0)                                                                                                           
  ttkernel.copy_tile_init(%rhs_cb)                                                                                                                 
  ttkernel.copy_tile(%rhs_cb, %0, %src1)                                                                                                           
  ttkernel.add_binary_tile_init                                                                                                                    
  ttkernel.add_binary_tile(%src0, %src1, %odst)  
```

### DIFF
--- a/lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
@@ -26,6 +26,7 @@
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "ttlang/Dialect/TTL/IR/TTLOps.h"
+#include "ttlang/Dialect/TTL/IR/TTLOpsTypes.h"
 #include "ttlang/Dialect/TTL/Passes.h"
 #include "ttmlir/Dialect/TTKernel/IR/TTKernelOps.h"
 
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
```diff
@@ -54,12 +47,92 @@ static int64_t getDstIdxForValue(Value v) {
...
+
+/// Trace from a tile operand (block argument) to its associated CB.
+/// The tile must be a block argument of a ttl.compute body, and the
+/// corresponding input must come from ttl.attach_cb.
+/// Returns nullptr if the CB cannot be found.
+static Value getCBForTileOperand(Value tile) {
+  auto blockArg = dyn_cast<BlockArgument>(tile);
+  if (!blockArg) {
+    return nullptr;
+  }
```
brnorris03: This tracing is not ideal, but even if left here, I think there shouldn't be multiple variants of `getCB*` methods in various places -- combine in a single utility function that can be used for different types of operations (so that if the IR design is improved to eliminate the need for this, it would be minimally intrusive).

brnorris03:
So my plan for dst is to insert `ttl.copy_tile` ops that are trivially converted to ttkernel equivalents later, so I am adding `ttl.copy_tile` (will make a separate PR for that so it's available in general). I am currently working on habing the tiling + dst assignment right after `convert-ttl-to-compute` and before the loop transformation. Just letting you know since it will change this part of the ttkernel lowering.

PR #159, opened 2025-12-23
[ttl] The one python bindings PR

zoecarver:
Port the python bindings from d2m to ttl. Refactor to remove existing runtime tests and replace them with parameterized pytest. Write new lit tests. 

Removed the metal runtime path and all d2m logic, renamed accordingly. 

Overhauled operators.py to auto generate almost everything (kind of cool). 

Filed a few issues for patterns we don't support, there is still more work to do:

- [x] get pytest running (hangs right now, maybe can defer)
- [x] get lit working natively (hacked right now because I lost the battle with lit config, required)
- [ ] add a large-ish fused kernel (follow up change likely)

### DIFF
--- a/python/ttlang/ttl_api.py
+++ b/python/ttlang/ttl_api.py
@@ -238,7 +238,11 @@ def _write_kernel_to_tmp(
             placeholder = int(m.group(1))
             local_idx = placeholder - 42
             # Map local accessor index to global tensor index
-            global_idx = tensor_indices[local_idx] if local_idx < len(tensor_indices) else local_idx
+            global_idx = (
+                tensor_indices[local_idx]
+                if local_idx < len(tensor_indices)
+                else local_idx
+            )
             actual_offset = global_idx + num_tensors
             return f"TensorAccessorArgs<{actual_offset}, 0>()"
 
@@ -351,7 +355,11 @@ def _compile_ttnn_kernel(
     # Write all kernels to /tmp for debugging
     for kernel_idx, (name, thread_type) in enumerate(kernel_info):
         cpp_source = ttkernel_to_cpp_by_name(module, name)
-        tensor_indices = thread_tensor_indices[kernel_idx] if kernel_idx < len(thread_tensor_indices) else []
+        tensor_indices = (
+            thread_tensor_indices[kernel_idx]
+            if kernel_idx < len(thread_tensor_indices)
+            else []
+        )
         _write_kernel_to_tmp(name, cpp_source, num_tensors, tensor_indices)
 
     kernel_paths = []
@@ -361,8 +369,14 @@ def _compile_ttnn_kernel(
 
     for kernel_idx, (name, thread_type) in enumerate(kernel_info):
         cpp_source = ttkernel_to_cpp_by_name(module, name)
-        tensor_indices = thread_tensor_indices[kernel_idx] if kernel_idx < len(thread_tensor_indices) else []
-        kernel_path = _write_kernel_to_tmp(name, cpp_source, num_tensors, tensor_indices)
+        tensor_indices = (
+            thread_tensor_indices[kernel_idx]
+            if kernel_idx < len(thread_tensor_indices)
+            else []
+        )
+        kernel_path = _write_kernel_to_tmp(
+            name, cpp_source, num_tensors, tensor_indices
+        )
         kernel_paths.append((kernel_path, thread_type))
 
         if thread_type == "compute":


lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -110,12 +110,10 @@ mlir::LogicalResult mlir::tt::ttl::AttachCBOp::verify() {
...
-  // Require the CB block shape rank to match the tensor rank (tile grid).
-  if (static_cast<int64_t>(cbTy.getShape().size()) != tensorTy.getRank()) {
-    return emitOpError() << "cb shape rank (" << cbTy.getShape().size()
-                         << ") must match tensor rank (" << tensorTy.getRank()
-                         << ")";
-  }
+  // TODO: Revisit shape rank validation for TTNN tensors.
+  // TTNN tensors have 4D device shape (grid + shard) while CBs have 2D shard
+  // shape. For now, only validate element types match. The relationship between
+  // tensor shape and CB shape needs further investigation.
```
zoecarver: @brnorris03 curious to hear your thoughts on this

python/gen_elementwise.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
```
zoecarver: Also curious to hear your thoughts on this

test/python/test_elementwise_ops.py
```diff
@@ -0,0 +1,291 @@
...
+Auto-generated tests for TTL elementwise operations.
+
+Tests all binary and unary elementwise ops against PyTorch equivalents,
+with both L1 and DRAM memory configurations.
+"""
+
+# UNSUPPORTED: system-darwin
+# RUN: %python -m pytest %s -v
+
+# NOT YET RUNNING this is just a placeholder for how this could look.
```
zoecarver: Also would love your thoughts on this pytest strategy.

python/ttlang/semaphore.py
phizalev-TT: Do we have corresponding `ttl` ops?

phizalev-TT (changes requested):

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03: would be good to create a separate PR for this bugfix

test/python/test_elementwise_ops.py
brnorris03: I put together some pytest scaffolding for the middle end (but re-usable for DSL inputs, too) at https://github.com/tenstorrent/tt-lang/pull/167 -- it should enable very minimalistic test definitions (and single ops fully auto-generated, as you already started here). I think it would be great to have a single test base that does all the logistics of compiling and running.

In any case, after you merge this, I should be able to consolidate both the DSL-based and middle end to end tests if the design works as expected.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -430,15 +435,18 @@ materializeTensorAccessor(Value tensor, Value bankBase,
...
-  auto rowStride =
-      rewriter.create<arith::ConstantIntOp>(loc, layout.rowStrideElems, 32);
-  auto colStride =
-      rewriter.create<arith::ConstantIntOp>(loc, layout.colStrideElems, 32);
+  // TODO(XX): Placeholder CTA offsets - Python regex replaces 42+argIdx
+  // with actual offsets from ttnn.TensorAccessorArgs.get_compile_time_args().
+  auto argIdx = getTensorFuncArgIndex(tensor);
+  int32_t ctaPlaceholder =
+      42 + (failed(argIdx) ? 0 : static_cast<int32_t>(*argIdx));
+  constexpr int32_t crtaPlaceholder = 0;
```
brnorris03: Regex replacement in the C++? Sounds kind of risky... Can you instead track the actual compile-time args offsets during the MLIR generation. So something like this:

1. Compute `args_per_tensor` early in `_compile_and_run_kernel` (where actual tensors are available):
   ```python
   args_per_tensor = len(ttnn.TensorAccessorArgs(args[0]).get_compile_time_args())
   ```

2. Pass it through the compilation pipeline as a kwarg to `TTLGenericCompiler.__init__` and store it.

3. In `ttl_ast.py` `_emit_entry`, compute CTA offsets as tensor arguments are collected:
   ```python
   cta_offset = tensor_index * self.args_per_tensor
   ```

4. Attach each CTA offset as a function argument attribute when creating the `func.FuncOp`:
   ```python
   self.func_entry.setArgAttr(i, "ttl.cta_offset", IntegerAttr.get(..., cta_offset))
   ```

5. When compiling, in `materializeTensorAccessor`, we'd just read the attribute from the function argument

brnorris03: I propose that this (hacky) fix is split out from this PR and merged first. In the meanwhile I am working on adding support for CTA and CRTA argument chaining as shown above (to tt-mlir). When that is merged, we can get rid of the hack and update the AST code generator to include the base cta and crta indices to the dm threads.

python/TTLModule.cpp
```diff
@@ -33,4 +36,28 @@ void populateTTLModule(nb::module_ &m) {
       .def_prop_ro("start", &SliceAttr::getStart)
       .def_prop_ro("stop", &SliceAttr::getStop)
       .def_prop_ro("step", &SliceAttr::getStep);
+
+  //===--------------------------------------------------------------------===//
+  // CircularBufferType
+  //===--------------------------------------------------------------------===//
+
+  tt_type_class<CircularBufferType>(m, "CircularBufferType")
```
brnorris03: thank you for adding it!

brnorris03 (approved):
Much offline discussion on the tensor accessors, but nothing that would stop merging this. I would like to get the current e2e working within some limited context (with known next steps to remove limitations).

test/python/simple_add_with_stmt.py
```diff
@@ -0,0 +1,226 @@
...
+    print("TTNN not available - exiting")
+    exit(0)
+
+
+@pykernel_gen(grid=(1, 1), block_factors=[(1, 1), (1, 1), (1, 1)])
+def add_with_kernel(lhs, rhs, out):
+    """Add kernel using 'with' pattern for automatic CB lifecycle."""
+    lhs_accessor = TensorAccessor(lhs)
+    rhs_accessor = TensorAccessor(rhs)
+    out_accessor = TensorAccessor(out)
```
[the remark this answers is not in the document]
zoecarver: I think we agreed that TAs can go away (in the spec at least)

brnorris03 (approved):
LGTM!

phizalev-TT (approved):

PR #162, opened 2025-12-23
Cooperative mode is now sole running mode

FConstantinos:
In addition, yields now bubble up to the scheduler.


phizalev-TT:
What is left to be able to simulate matmuls from @arichinsTT ?

FConstantinos:
> What is left to be able to simulate matmuls from @arichinsTT ?

Also, "with" syntax

phizalev-TT (approved):

Issue #193, opened 2025-12-30
Introduce make_circular_buffer_like


zoecarver:
Right now we create CBs implicitly from args, we should introduce `make_circular_buffer_like` to match the example/spec

PR #195, opened 2025-12-30
Introduce `make_circular_buffer_like` (#193)

zoecarver:
Introduce `make_circular_buffer_like` to replace implicit CB creation. Update CB index assignment logic (same mlir emitted) and tests accordingly.

Note: we now emit bind_cb ops in alphabetical order due to how we iterate over captures, but the indexes are still creation order and correctly mapped.

Test changes are all just updating to use new pattern and a few small IR checks for ordering (and two new invalid tests).


python/ttlang/circular_buffer.py
```diff
@@ -25,8 +44,33 @@ class CircularBuffer:
...
+    def __init__(
+        self,
+        tensor: Any,
+        shape: Tuple[int, int],
+        buffer_factor: int,
+    ):
+        if len(shape) != 2:
+            raise ValueError(f"shape must be a 2-tuple, got {shape}")
+        if buffer_factor < 1:
+            raise ValueError(f"buffer_factor must be >= 1, got {buffer_factor}")
```
brnorris03: nit: technically buffer factor also can't exceed cb capacity (32)

test/python/invalid/invalid_cb_buffer_factor.py
```diff
@@ -0,0 +1,120 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# RUN: not %python %s 2>&1 | FileCheck %s
+
+"""
+Validation test: CB buffer_factor must be >= 1.
+
+This test verifies that using buffer_factor=0 for make_circular_buffer_like()
+raises the expected ValueError.
+"""
```
brnorris03: I think this test can be even shorted if it's only meant to check the buffer factor (don't need compute for example and probably just one dm thread is enough?). That's probably true for some of the other invalid tests (the possibility to shorten without losing anything).

brnorris03 (commented):
Minor comments, mainly suggestion to try to simplify the invalid tests; looks great!

Maybe also needs rebase? I see tensor accessors that I think you removed in one of the previous PRs.

FConstantinos (approved):
LGTM, just a minor think I'm curious about

test/python/invalid/invalid_cb_buffer_factor.py
```diff
@@ -0,0 +1,15 @@
...
+
+# RUN: not %python %s 2>&1 | FileCheck %s
+
+"""
+Validation test: CB buffer_factor must be in range [1, 32].
+"""
+
+from ttlang import make_circular_buffer_like
+
+# CHECK: buffer_factor must be in range [1, 32]
```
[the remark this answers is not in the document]
zoecarver: Because we only have 32 CBs, so theoretically you could only buffer up to 32.

PR #196, opened 2025-12-30
Remove TensorAccessors (#189)

### LINKED ISSUE #189: Remove TensorAccessors
No reason we can't just use tensors directly, right now the TAs don't provide much except unneeded verbosity, we can remove them.

zoecarver:
No reason we can't just use tensors directly, right now the TAs don't provide much except unneeded verbosity, we can remove them. Matches the spec. 

Change is strait forward, tests are updated accordingly (no change to behavior or checks). 

fixes #189


zoecarver:
Need to remove tensor_accesor.py from cmake

PR #203, opened 2025-12-31
[TTL] Fix TensorAccesorArgs generation in thread functions

### LINKED ISSUE #168: [ttl] Update copy lowering to use the correct TensorAccessorArgs
This will likely require updating ttkernel to support `next_compile_time_args_offset` and piping that through. Right now we are using regex as a temporary hack. 

It's possible that `next_compile_time_args_offset` won't work for whatever reason (maybe doesn't work for multi-kernel) and we can pipe through the correct args using attributes.

brnorris03:
### What?
Fixes the generation of `TensorAccessorArgs` to be per-function, not per-copy site and using a base compile-time args index (number of CBs in the kernel) for the first TensorAccessor offset and incrementally computing subsequent indices. The arguments to thread functions are filtered to include only tensors used in the thread.

**Before (per-copy materialization with placeholders):**
```cpp
// At each copy site:
auto accessor = TensorAccessor(TensorAccessorArgs<42, 0>(), bankBase, pageSize);
// Python regex replaces 42 with actual offset
```

**After (pre-materialization with simple offsets):**
```cpp
// At function entry (once per function):
auto args_0 = TensorAccessorArgs<3, 0>();  // base_cta=3 (num CBs), runtime_offset=0
auto accessor_0 = TensorAccessor(args_0, bankBase_0, pageSize);
auto args_1 = TensorAccessorArgs<4, 1>();  // base_cta+1, runtime_offset+1
auto accessor_1 = TensorAccessor(args_1, bankBase_1, pageSize);

// At copy sites: reuse pre-materialized accessors
noc_async_read_tile(tile_idx, accessor_0, write_ptr);
```

### Why?
Tensor accessor construction was duplicated at copy sites. A modification of generated C++ code was required.

### How?
The lowering now pre-materializes `TensorAccessorArgs` and `TensorAccessor` values at function entry for NOC kernels and reuses them, allowing multiple tensor arguments to be handled via chained offsets instead of hardcoded indices.

### How to Test?
- the existing `check-ttlang-all` target

### Checklist:
*   [x] Self-reviewed (style, logic)
*   [x] Added tests (or justified none needed)
*   [x] PR is small and focused (one task)

Closes #168

### DIFF
--- a/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
@@ -451,9 +451,11 @@ materializeFunctionTensorAccessors(func::FuncOp funcOp, OpBuilder &builder) {
   builder.setInsertionPointToStart(&entryBlock);
   Location loc = funcOp.getLoc();
 
-  // Materialize accessors for tensor arguments with simple index offsets.
+  // Materialize accessors for tensor arguments with simple local index offsets.
   // All function arguments are pre-filtered in Python to only include used
-  // tensors.
+  // tensors. Host builds per-thread compile_time_args with TensorAccessorArgs
+  // config for only the tensors this noc thread function uses, so we can use
+  // simple local indexing.
   unsigned accessorIndex = 0;
 
   for (auto [argIdx, arg] : llvm::enumerate(funcOp.getArguments())) {
@@ -482,7 +484,9 @@ materializeFunctionTensorAccessors(func::FuncOp funcOp, OpBuilder &builder) {
     auto pageSize =
         builder.create<arith::ConstantIntOp>(loc, pageSizeBytes, 32);
 
-    // Materialize accessor with simple index offsets.
+    // Materialize accessor with simple local index offsets.
+    // baseCTA points to the start of this kernel's TensorAccessorArgs in
+    // compile_time_args.
     Value accessor =
         buildTensorAccessor(loc, builder, baseCTA + accessorIndex,
                             baseCRTA + accessorIndex, *bankBase, pageSize);
--- a/python/ttlang/_src/ttl_ast.py
+++ b/python/ttlang/_src/ttl_ast.py
@@ -93,6 +93,15 @@ class CompilerContext:
     tiled: bool
 
 
+@dataclass
+class TensorArgInfo:
+    """Metadata for a tensor function argument."""
+
+    name: str
+    global_index: int
+    arg_type: Type  # MLIR type
+
+
 class TTLGenericCompiler(TTCompilerBase):
     """Compiler that generates TTL dialect ops from Python AST."""
 
@@ -204,9 +213,7 @@ class TTLGenericCompiler(TTCompilerBase):
             )
 
         # Collect tensor captures for function arguments
-        self._tensor_accessor_names = []
-        self._tensor_accessor_global_indices = []
-        func_arg_types = []
+        all_tensor_args = []
         for name, val in self.captures.items():
             if is_ttnn_tensor(val):
                 tensor_type = _build_tensor_type(
@@ -216,28 +223,25 @@ class TTLGenericCompiler(TTCompilerBase):
                     self.context.tiled,
                     self.context.memory_space,
                 )
-                self._tensor_accessor_names.append(name)
-                self._tensor_accessor_global_indices.append(
-                    get_tensor_global_index(val)
+                all_tensor_args.append(
+                    TensorArgInfo(
+                        name=name,
+                        global_index=get_tensor_global_index(val),
+                        arg_type=tensor_type,
+                    )
                 )
-                func_arg_types.append(tensor_type)
 
         # Collect which tensors are actually used in the function body
-        used_tensors = self._collect_used_tensor_names(node)
+        tensor_names = {arg.name for arg in all_tensor_args}
+        used_tensors = self._collect_used_tensor_names(node, tensor_names)
 
-        # Now filter to only used tensors and create the actual function.
-        # This ensures only captured tensors that are actually referenced in the
-        # thread function body become function arguments for TensorAccessor materialization.
-        filtered_names = []
-        filtered_arg_types = []
-        filtered_indices = []  # Track original indices for mapping
-        for i, tensor_name in enumerate(self._tensor_accessor_names):
-            if tensor_name in used_tensors:
-                filtered_names.append(tensor_name)
-                filtered_arg_types.append(func_arg_types[i])
-                filtered_indices.append(i)
+        # Filter to only used tensors
+        filtered_tensor_args = [
+            arg for arg in all_tensor_args if arg.name in used_tensors
+        ]
 
         # Create the actual function with only used tensor arguments
+        filtered_arg_types = [arg.arg_type for arg in filtered_tensor_args]
         self.func_entry = func.FuncOp(name=node.name, type=(filtered_arg_types, []))
 
         # Set thread attribute: ttl.kernel_thread = #ttkernel.thread<compute/noc>
@@ -258,10 +262,11 @@ class TTLGenericCompiler(TTCompilerBase):
         # Emit function body with filtered tensor arguments
         with InsertionPoint(func_bb):
             # Map filtered TensorAccessor function arguments to symbol table
-            for filtered_idx, orig_idx in enumerate(filtered_indices):
-                name = self._tensor_accessor_names[orig_idx]
-                self.symbol_tables[-1][name] = func_bb.arguments[filtered_idx]
-                self.streams.add(name)
+            for filtered_idx, tensor_arg in enumerate(filtered_tensor_args):
+                self.symbol_tables[-1][tensor_arg.name] = func_bb.arguments[
+                    filtered_idx
+                ]
+                self.streams.add(tensor_arg.name)
 
             # Prepopulate other captures (non-tensor)
             from ..circular_buffer import CircularBuffer
@@ -289,9 +294,10 @@ class TTLGenericCompiler(TTCompilerBase):
 
         self.symbol_tables.pop()
 
-        # Update _tensor_accessor_global_indices to only include filtered tensors
+        # Store filtered tensor metadata for use in ttl_api.py
+        # (accessed via ct._tensor_accessor_global_indices in ttl_api.py:722)
         self._tensor_accessor_global_indices = [
-            self._tensor_accessor_global_indices[i] for i in filtered_indices
+            arg.global_index for arg in filtered_tensor_args
         ]
 
         # Note: base_cta_index is set later in ttl_api.py after all threads are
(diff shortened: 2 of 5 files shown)


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -110,15 +110,14 @@ static FailureOr<unsigned> getTensorFuncArgIndex(Value tensor) {
 /// Get the L1 buffer address from runtime args for a tensor function argument.
 /// Runtime args are indexed by the tensor's function argument position.
 static FailureOr<Value>
-getBufferAddressFromRuntimeArg(Value tensor, Location loc,
-                               ConversionPatternRewriter &rewriter) {
+getBufferAddressFromRuntimeArg(Value tensor, Location loc, OpBuilder &builder) {
```
brnorris03: Switched helper builders (this one and others below) to `OpBuilder` so they can be called from non-rewrite contexts and to avoid mutating pattern rewriter state outside a rewrite.

python/ttlang/_src/ttl_ast.py
```diff
@@ -126,31 +139,64 @@ def _emit_entry(self, node):
...
+                    "name": cb_name,
+                    "shard_shape": info["shard_shape"],
+                    "element_type": info["element_type"],
+                    "cb_index": cb_index,
+                }
+            )
+
         # Collect TensorAccessor captures for function arguments
         self._tensor_accessor_names = []
         self._tensor_accessor_global_indices = []
```
zoecarver: Dead now, right?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -520,8 +617,9 @@ static LogicalResult lowerTensorToCB(CopyOp op, Value srcTensor, Value dstCB,
         op, "tensor must be a function argument for runtime arg mapping");
   }
 
-  // Create tensor accessor with actual buffer address.
-  auto srcAccessor = materializeTensorAccessor(srcTensor, *bankBase, rewriter);
+  // Look up or create tensor accessor.
```
zoecarver: Nit: does not create, right?

zoecarver (commented):
As discussed offline, I'd really like to split at least the TA logic into its own PR. I think that's very important to get right and I'm having a hard time splitting it out of other changes here mentally. 

High level question regarding `next_compile_time_args_offset` (also discussed offline a week or two ago): given that each kernel gets written to a different file, is this going to be consistent across files? TAs are global, so I think we need it to be. Take this test for example: 

```
  @ttl.kernel()
  def test_multi_tensor_kernel():
      @ttl.datamovement()
      def reader(A: TensorAccessor, B: TensorAccessor, lhs_cb: CircularBuffer):
          copy(A, lhs_cb)
          copy(B, lhs_cb)

      @ttl.datamovement()  
      def writer(out_cb: CircularBuffer, C: TensorAccessor):
          copy(out_cb, C)
```

Will it produce different non type template parameters (when evaluated) in `TensorAccessorArgs` across the different files? My intuition says they should be somewhat overlapping which is obviously a bug.

zoecarver:
> I will wait for that to be implemented first since I guess I don't really know what ttl_api.py plans to generate and only considered CBs as arguments to dm threads.

I'm not planning anything else. We have CBs and TAs (both as captures now, so no explicit args from the user, but the explicit args make it easier to see what we codegen in my example). You can just think of captures as args. 

IIUC you're saying we set up only the args each kernel cares about so we write into the correct buffer, which, yes, that is correct. The second dm kernel's arg0 will actually be global TA idx=2, because the frontend does this mapping. 

But what I think we are missing in this PR is that `base_cta_index = len(CB_symbol_table)` for every dm function. So the metadata will be indexed incorrectly. Take the example above and make C have shape 64x64, and watch the runtime spit out -999 (just tried this locally).

python/ttlang/ttl_api.py
```diff
@@ -750,6 +821,8 @@ def _compile_and_run_kernel(
                 raise type(e)(formatted) from None
             compiled_threads.append(ct)
             thread_tensor_indices.append(ct._tensor_accessor_global_indices)
+            # Extract CB indices from _cb_info
+            thread_cb_indices.append([cb["cb_index"] for cb in ct._cb_info])
```
zoecarver: Looks like this is not defined

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -589,8 +690,18 @@ static LogicalResult lowerCBToTensor(CopyOp op, Value srcCB, Value dstTensor,
   return success();
 }
 
+/// Map from function to its pre-materialized tensor accessor map.
+/// Used to pass accessor lookup tables from pre-materialization phase to
+/// CopyLowering pattern. Functions are keyed by their Operation* to avoid the
+/// need for a DenseMapInfo specialization on FuncOp.
+using FuncAccessorMaps = DenseMap<Operation *, DenseMap<unsigned, Value>>;
+using FuncAccessorMapsPtr = std::shared_ptr<const FuncAccessorMaps>;
```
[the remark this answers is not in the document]
brnorris03: Shared pointer because:
 - the map is created once in `lowerTTLOpsToTTKernel` before patterns are added
 - multiple `CopyLowering` pattern instances may share the same map (read-only) -- so can't be unique ptr
 - the map must outlive all pattern instances during rewrite application

python/ttlang/ttl_api.py
```diff
@@ -792,11 +865,10 @@ def _compile_and_run_kernel(
             "symbol-dce",
         ]
 
-        pipeline = ",".join(pipeline_passes)
```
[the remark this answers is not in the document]
brnorris03: that was a partial merge, reverted to main

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -422,32 +406,143 @@ static std::optional<TransferKind> getTransferKindFromHandleType(Type t) {
...
+
+  return it->second;
+}
+
+/// Look up a pre-materialized TensorAccessor for the given tensor value.
+/// Returns failure if no pre-materialized accessor exists. All tensor accessors
+/// in NOC kernels must be pre-materialized at function entry with proper
+/// chaining via materializeFunctionTensorAccessors().
+[[nodiscard]] static FailureOr<Value>
+materializeTensorAccessor(Value tensor, Value /*bankBase*/,
```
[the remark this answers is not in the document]
brnorris03: removed

python/ttlang/_src/ttl_ast.py
```diff
@@ -126,31 +139,64 @@ def _emit_entry(self, node):
...
+        # Now build _cb_info for ALL CBs in the kernel (sorted by index).
+        # For CBs not in this thread's signature, use info from the first CB as template.
+        # All CBs in a kernel typically have the same shape/dtype in simple cases.
+        template_cb = (
+            next(iter(signature_cb_info.values())) if signature_cb_info else None
+        )
+
+        for cb_name, cb_index in sorted(
+            self._kernel_cb_symbol_table.items(), key=lambda x: x[1]
+        ):
```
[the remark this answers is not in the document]
brnorris03: removed symbol table.

python/ttlang/ttl_api.py
brnorris03: removed

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -628,8 +677,18 @@ static LogicalResult lowerCBToTensor(CopyOp op, Value srcCB, Value dstTensor,
...
+/// need for a DenseMapInfo specialization on FuncOp.
+using FuncAccessorMaps = DenseMap<Operation *, DenseMap<unsigned, Value>>;
+using FuncAccessorMapsPtr = std::shared_ptr<const FuncAccessorMaps>;
+
 struct CopyLowering : OpConversionPattern<CopyOp> {
-  using OpConversionPattern::OpConversionPattern;
+  CopyLowering(const TypeConverter &typeConverter, MLIRContext *context,
+               FuncAccessorMapsPtr funcAccessorMaps)
+      : OpConversionPattern(typeConverter, context),
+        funcAccessorMaps(std::move(funcAccessorMaps)) {}
```
zoecarver: To make my above point more concrete: every use of the shared_ptr moves it, so it must not need to be shared! std::move nulls out the pointer, so either this is a bug with sharing, or we don't actually need sharing.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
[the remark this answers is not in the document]
zoecarver: Here's my suggestion for the copylowering changes: 1) add a `MaterializeTensorAccessorsPass` that walks NOC functions, creates chained accessors at entry 2) build a really simple map (if still needed) at the start of this pass and let copylowering read it. 

This is already our most complicated pass. If we can make it more composable and self contained, I think that would be great. However, I don't want to make too much work for you, so if you really want to keep this general architecture, I think a local map is fine. 

Or, maybe the simplest: 
```
FailureOr<Value> findAccessorForTensor(Value tensor, func::FuncOp func) {
    auto blockArg = dyn_cast<BlockArgument>(tensor);
    if (!blockArg)
      return failure();

    unsigned argIdx = blockArg.getArgNumber();
    for (auto op : func.getBody().front().getOps<ttk::TensorAccessorOp>()) {
      if (auto defOp = op.getBankBase().getDefiningOp<ttk::GetCommonArgValOp>())
        if (defOp.getArgIndex() == argIdx)
          return op.getResult();
    }
    return failure();
  }
```
Yes, it's no longer O(1) lookups but almost certainly faster than heap memory :)

zoecarver:
> The host must set up the kernel with the correct tensors which could be anywhere and any subset of however many other tensors there might be.

If this is how you want to model it, then you need to update what is passed for CTA. But to be clear, it will need to be the exact subset of used tensors (same logic as crta).

zoecarver:
I still think we are making this more complicated than it needs to be. We are doing a lot of work just to use `next_compile_time_args_offset` which seems like a less clear API in general. It makes sense for a programmer writing a kernel to use `next_compile_time_args_offset`, but the compiler needs to calculate the indexes anyway, so why not just pass them directly? For a programmer, it's easy for them to get indexes wrong, for the compiler, it's not. 

My proposal here, now that we've agreed there is an issue (I hope), is to pass `base_cta_index` and `base_crta_index` as attributes, then just use that value directly in the lowering. I think that would be a lot simpler and smaller change.

With the approach in this PR, we still pass `base_cta_index`, then we have to materialize accessors with chaining, and the runtime must filter CTA per kernel (or C++ must use global indexing). 

With what I'm proposing we'd pass both `base_cta_index` and `base_crta_index` as function attributes and the compiler would compute exact indicies for each tensor accessor during lowering. So

```
  // Instead of chaining:
  auto args = TensorAccessorArgs<base_cta_index, 0>();  // Wrong if CTA is global

  // Compiler directly computes:
  auto args = TensorAccessorArgs<5, 0>();  // Exact index for 'out' tensor
```

**This also makes the tests very easy to verify because we can say "yep that's the 6th arg, so the literal 5 makes sense".**

PR #212, opened 2026-01-05
Fix multi-tile CB addressing and add elementwise shape sweep tests

zoecarver:
This PR fixes circular buffer addressing for multi-tile operations. Previously, DMA tile loops used the same CB base address for all tiles, and pack operations always used tile index 0. Now:
  - `NocAsyncReadTileOp`/`NocAsyncWriteTileOp` compute proper CB addresses as `cbPtr + tileOffset * pageSize`
  - `PackTileOp` derives CB tile index from enclosing `scf.for` loops via new `computeCBTileIndexFromLoops()` helper
  - `getTileGridShapeFromValue()` handles rank-4 device tensors `[grid_y, grid_x, shard_tiles_y, shard_tiles_x]`

  Adds `test_elementwise_shapes.py` to sweep unary/binary elementwise ops across tile configurations (1x1 to 3x1/1x3), with `buffer_factor = tile_rows * tile_cols` to ensure semaphore capacity matches tiles per block. Currently limited to 1D tile shapes (Nx1 or 1xN) due to resource constraints with 2D multi-tile shapes.

### DIFF
--- a/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
@@ -156,12 +156,14 @@ static Value computeCBTileIndexFromLoops(Operation *op, OpBuilder &builder) {
 
   // Validate assumptions: all loops have step=1 and lower bound=0.
   for (auto loop : loops) {
-    assert(loop.getConstantLowerBound() == 0 &&
+    auto lb = getConstantIntValue(loop.getLowerBound());
+    assert(lb && *lb == 0 &&
            "computeCBTileIndexFromLoops: expected lower bound of 0");
-    assert(loop.getConstantUpperBound() &&
-           "computeCBTileIndexFromLoops: expected constant upper bound");
-    auto step = loop.getStepAsInt();
-    assert(step == 1 && "computeCBTileIndexFromLoops: expected step of 1");
+    auto ub = getConstantIntValue(loop.getUpperBound());
+    assert(ub && "computeCBTileIndexFromLoops: expected constant upper bound");
+    auto step = getConstantIntValue(loop.getStep());
+    assert(step && *step == 1 &&
+           "computeCBTileIndexFromLoops: expected step of 1");
   }
 
   // Process in reverse order (innermost-first) without mutating the vector.


test/python/test_elementwise_shapes.py
```diff
@@ -0,0 +1,501 @@
...
+
+# =============================================================================
+# Kernel Factories - generate kernels with specific shapes
+# =============================================================================
+
+# Cache for generated kernels: (name, op, tile_rows, tile_cols) -> kernel
+_kernel_cache = {}
+
+
+def make_binary_kernel(name: str, op: str, tile_rows: int, tile_cols: int):
```
brnorris03: The `make_*_kernel` functions have quite a bit of duplicated code, may be able to consolidate in a more generic kernel function.

nit: missing `-> Callable` return type in all the `make_*_kernel` methods.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -495,14 +534,24 @@ getTileGridShape(const RankedTensorType &tensorTy) {
...
+  if (dims.size() == 2) {
     return getTileGridShape(tensorTy);
+  } else if (dims.size() == 4) {
+    // Rank-4 tensor: [grid_y, grid_x, shard_tiles_y, shard_tiles_x]
+    // The last two dimensions are already tile counts, not element counts.
+    return {dims[2], dims[3]};
   }
-  return {1, 1};
+
+  llvm_unreachable("expected rank-2 or rank-4 tensor");
```
brnorris03: I know it doesn't make much of a difference at runtime, but I believe assert rather than llvm_unreachable is more appropriate here.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -572,12 +619,37 @@ static LogicalResult lowerTensorToCB(CopyOp op, Value srcTensor, Value dstCB,
 
   auto [tilesY, tilesX] = getTileGridShapeFromValue(srcTensor);
 
+  // Get page size for CB pointer arithmetic from TTNNLayoutAttr.
+  auto tensorTy = llvm::cast<RankedTensorType>(srcTensor.getType());
+  auto layoutAttr =
+      mlir::dyn_cast_or_null<ttnn::TTNNLayoutAttr>(tensorTy.getEncoding());
+  assert(layoutAttr && "tensor must have TTNNLayoutAttr encoding");
```
brnorris03: nit: message can be a bit more informative, e.g., "lowerTensorToCB: srcTensor must have TTNNLayoutAttr encoding"

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -126,6 +126,47 @@ static bool isNocKernel(Operation *op) {
...
+
+  Value linearIdx = builder.create<arith::ConstantIndexOp>(loc, 0);
+  for (size_t i = 0; i < loops.size(); ++i) {
+    Value iv = loops[i].getInductionVar();
+
+    // Compute stride as product of remaining loop upper bounds.
+    Value stride = builder.create<arith::ConstantIndexOp>(loc, 1);
+    for (size_t j = i + 1; j < loops.size(); ++j) {
+      stride =
+          builder.create<arith::MulIOp>(loc, stride, loops[j].getUpperBound());
```
brnorris03: This assumes `getUpperBound()` always returns a valid value. Consider `getUpperBoundorDie()` or validate assumptions (that loops have step 1 and lb 0).

brnorris03 (commented):
Some more comments and suggestions after a more thorough reading than the first one. Looks good overall.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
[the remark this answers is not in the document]
zoecarver: I think it's more of an unreachable thing, we want to signal that there is no set of inputs that should take this path. Regardless, I don't think which one we use makes material/functional difference in this case.

brnorris03 (approved):
lgtm, thank you

PR #213, opened 2026-01-06
Implement tensor_slice op and dynamic tensor indexing

zoecarver:
Adds support for `tensor[row, col]` syntax in TTL kernels to access specific tiles within multi-tile tensors. Previously, tensor indexing was restricted to `[0, 0]`.

  - Add `TensorSliceType` and `TensorSliceOp` to the TTL dialect for representing tile-indexed tensor views
  - Update Python DSL to emit `ttl.tensor_slice` ops when tensor subscript syntax is used
  - Add lowering in `ConvertTTLToTTKernel` to compute correct linear tile offsets (`row * num_cols + col`) for NOC read/write operations
  - Add `simple_tensor_slice.py` lit test and `pytest_tensor_slice.py` parameterized test covering 1x1 through 16x16 tile shapes
  - Based on #212

### DIFF
--- a/include/ttlang/Dialect/TTL/IR/TTLOps.td
+++ b/include/ttlang/Dialect/TTL/IR/TTLOps.td
@@ -116,9 +116,8 @@ def TTL_CopyOp : TTL_Op<"copy", [MemoryEffects<[MemRead, MemWrite]>]> {
     corresponding `ttl.wait` has completed.
 
     Exactly one operand must be a circular buffer (`!ttl.cb`). The other operand
-    must be either:
-    - A `!ttl.tensor_slice` (a view into a tensor at specific tile indices)
-    - A ranked tensor with TTNN layout encoding (transfers all tiles via loop)
+    must be a ranked tensor (the result of a tensor slice) with TTNN layout
+    encoding (transfers all tiles via loop).
 
     TODO(ttl): Add an optional TRID attribute (range 0..15) when TRID-specific
     ttkernel noc ops land in tt-mlir. Issue: #87.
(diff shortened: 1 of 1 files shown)


test/python/pytest_tensor_slice.py
brnorris03: suggest using the usual `test_` prefix for pytest

include/ttlang/Dialect/TTL/IR/TTLOpsTypes.td
brnorris03: Thanks for adding the types; can you also please add a simple python bindings test (in `test/bindings/python`)?

brnorris03 (commented):
looks good in general, thank you

arichinsTT (changes requested):
no tensor_slice value type, have it return reduced tensor, which should remove a lot of cases.  I think it is worth while to have handling for multiple tile slices from the get go.

python/ttlang/operators.py
```diff
@@ -81,28 +81,35 @@ def wait(ast_self: CopyTransferHandler):
         return ttl.wait(ast_self)
 
 
+def _make_tensor_slice(tensor, indices):
+    """Create a ttl.tensor_slice from a tensor and tile indices."""
+    if len(indices) != 2:
```
arichinsTT: shouldn't this match the rank of the tensor and can we do bounds checks here?

zoecarver:
Oops accidentally pushed to the wrong branch 🫣 

All comments have been addressed or responded to. Thank you!

brnorris03 (approved):
lgtm, thank you for updating!

test/ttlang/Translate/TTLToCpp/compute_with_data_movement.mlir
```diff
@@ -77,15 +77,18 @@
 // Reader kernel: reads A and B from DRAM, pushes to CB0 and CB1
 func.func @reader_binary(%a: tensor<64x64xf32, #layout>, %b: tensor<64x64xf32, #layout>)
     attributes {ttl.base_cta_index = 2 : i32, ttl.crta_indices = [0, 1], ttl.kernel_thread = #ttkernel.thread<noc>} {
+  %c0 = arith.constant 0 : index
   %cb0 = ttl.bind_cb {cb_index = 0, buffer_factor = 2} : !ttl.cb<[2, 2], f32, 2>
   %cb1 = ttl.bind_cb {cb_index = 1, buffer_factor = 2} : !ttl.cb<[2, 2], f32, 2>
 
   // Copy A to CB0
-  %xf_a = ttl.copy %a, %cb0 : (tensor<64x64xf32, #layout>, !ttl.cb<[2, 2], f32, 2>) -> !ttl.transfer_handle<read>
+  %slice_a = ttl.tensor_slice %a[%c0, %c0] : tensor<64x64xf32, #layout> -> tensor<64x64xf32, #layout>
```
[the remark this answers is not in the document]
zoecarver: Yes, we should be allowed to copy 2x2 tiles at a time if the CB shape is 2x2.

PR #215, opened 2026-01-06
Early Elementwise Op Fusion

zoecarver:
Adds fusion support for chains of elementwise tensor ops (like exp(a) + sqrt(b)) into a single ttl.compute block. When elementwise ops are not directly attached to circular buffers, the pass traces backwards through the op chain to find CB-attached roots, then emits all the tile-level ops inside a single fused compute body.

  - Recursive tracing supports arbitrary depth chains (tested up to 20 ops)
  - Uses existing `TTLElementwiseOps.def` to automatically support all unary/binary ops
  - Strict fusion heuristics: all intermediate values must have single use, all roots must be CB-attached
  - Clear diagnostic errors when fusion fails (e.g., "intermediate value has multiple uses", "cannot trace through non-elementwise op")

This does not fuse or merge multiple ttl.compute ops together; it only places elementwise ops inside a single compute block when strict heuristics are met.

### DIFF
--- a/include/ttlang/Dialect/TTL/IR/TTL.h
+++ b/include/ttlang/Dialect/TTL/IR/TTL.h
@@ -38,8 +38,8 @@ class TTLTileComputeOpTrait
 /// Trait for unary elementwise tensor operations (exp, sqrt, etc.).
 template <typename ConcreteType>
 class TTLUnaryElementwiseOpTrait
-    : public mlir::OpTrait::TraitBase<ConcreteType, TTLUnaryElementwiseOpTrait> {
-};
+    : public mlir::OpTrait::TraitBase<ConcreteType,
+                                      TTLUnaryElementwiseOpTrait> {};
 
 /// Trait for binary elementwise tensor operations (add, mul, etc.).
 template <typename ConcreteType>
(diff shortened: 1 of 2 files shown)


include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -60,6 +60,158 @@ inline bool isTileComputeOp(mlir::Operation *op) {
...
+  MultipleUses,
+};
+
+/// Result of tracing through elementwise ops to CB-attached roots.
+struct ElementwiseTraceResult {
+  /// CB-attached input values that form the roots of the chain.
+  mlir::SmallVector<mlir::Value, 2> rootInputs;
+  /// Operations in the chain, topologically ordered (roots first, sink last).
+  mlir::SmallVector<mlir::Operation *, 4> opsInOrder;
+  /// Failure reason (Success if tracing succeeded).
```
brnorris03: consider using `SmallSetVector` or `DenseSet` (more efficient search).

zoecarver:
Thank you Boyana! Will fix all :)

PR #216, opened 2026-01-06
Use new big CI runner

zoecarver:


brnorris03 (commented):
I would confirm first that the exact same steps work on qb or any tt hw machine, I suspect the tt-mlir configure options in `cmake/modules/ExternTTMLIR.cmake` would need updating since we don't require ttrt and use ttnn jit.

brnorris03:
> I want to land this to validate that the added worker is usable, then I can look into build config issues.

You can bypass requirements then.

PR #221, opened 2026-01-07
Kostas/sim launcher

FConstantinos:
A simulator launcher that can be used to run any example of ttlang

Also, this PR makes utils a standalone module.

The examples under examples/sim have moved to examples/ since now they should work with either simulator or compiler

### DIFF
--- a/test/sim/test_examples.py
+++ b/test/sim/test_examples.py
@@ -77,6 +77,7 @@ def assert_success_output(code: int, out: str) -> None:
         "eltwise_pipe_core3.py",
         "singlecore_matmul.py",
         "multicore_matmul.py",
+        "multistep_op.py",
     ],
 )
 def test_example_cli(script_name: str) -> None:
--- a/examples/multistep_op.py
+++ b/examples/multistep_op.py
@@ -1,0 +1,141 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+# type: ignore
+
+import ttnn
+import torch
+
+import ttl
+
+TILE_SIZE = 32
+GRANULARITY = 4
+
+
+@ttl.kernel(grid=(1, 1))
+def add_with_kernel(a, b, c, y):
+    row_tiles = GRANULARITY
+    col_tiles = GRANULARITY
+
+    rows = a.shape[0] // TILE_SIZE // row_tiles
+    cols = a.shape[1] // TILE_SIZE // col_tiles
+
+    a_cb = ttl.make_circular_buffer_like(
+        a, shape=(row_tiles, col_tiles), buffer_factor=2
+    )
+    b_cb = ttl.make_circular_buffer_like(
+        b, shape=(row_tiles, col_tiles), buffer_factor=2
+    )
+    c_cb = ttl.make_circular_buffer_like(
+        c, shape=(row_tiles, col_tiles), buffer_factor=2
+    )
+    y_cb = ttl.make_circular_buffer_like(
+        y, shape=(row_tiles, col_tiles), buffer_factor=2
+    )
+
+    @ttl.compute()
+    def add_compute():
+        for _ in range(rows):
+            for _ in range(cols):
+                with (
+                    a_cb.wait() as a_block,
+                    b_cb.wait() as b_block,
+                    c_cb.wait() as c_block,
+                    y_cb.reserve() as y,
+                ):
+                    y.store(a_block * b_block + c_block)
+
+    @ttl.datamovement()
+    def add_read():
+        for row in range(rows):
+            for col in range(cols):
+                with (
+                    a_cb.reserve() as a_blk,
+                    b_cb.reserve() as b_blk,
+                    c_cb.reserve() as c_blk,
+                ):
+                    tx_a = ttl.copy(
+                        a[
+                            row * row_tiles : (row + 1) * row_tiles,
+                            col * col_tiles : (col + 1) * col_tiles,
+                        ],
+                        a_blk,
+                    )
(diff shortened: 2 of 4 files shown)


test/python/utils/test_block_allocation.py
```diff
@@ -10,6 +10,7 @@
 import pytest
 
 from ttlang.utils.block_allocation import split_work_to_cores, get_large_matmul_params
+from utils.block_allocation import split_work_to_cores, get_large_matmul_params
```
brnorris03: duplicate import?

python/sim/ttlsim.py
```diff
@@ -0,0 +1,146 @@
...
+    sys.argv = [str(file_path)] + argv
+
+    # Read and execute the file
+    with open(file_path) as f:
+        code = compile(f.read(), str(file_path), "exec")
+        # Get the shadowed modules from sys.modules so they're available in exec
+        exec_globals: dict[str, Any] = {
+            "__name__": "__main__",
+            "__file__": str(file_path),
+            # Include shadowed modules
```
brnorris03: and maybe just in case, `"__builtins__": __builtins__`

test/sim/test_examples.py
```diff
@@ -1,111 +1,101 @@
...
+        text=True,
+        check=False,
+    )
+    return proc.returncode, proc.stdout
+
+
+class TestExamplesCLI:
+    def assert_success_output(self, code: int, out: str) -> None:
+        assert code == 0, f"ttlsim exited with code {code}. Output:\n{out}"
+        assert "success" in out.lower(), f"Expected 'success' in output. Got:\n{out}"
```
brnorris03: Both `TestExamplesCLI` and `TestMetalExamplesCLI` define their own `assert_success_output` methods with slightly different logic. Consider extracting to a module-level helper function and possibly unifying the success detection logic (one checks for "success", the other for "success" or "passed").

Maybe something like
```python
def assert_success_output(code: int, out: str, patterns: list[str] = None) -> None:
    """Shared assertion helper for CLI tests."""
    patterns = patterns or ["success"]
    assert code == 0, f"ttlsim exited with code {code}. Output:\n{out}"
    assert any(p in out.lower() for p in patterns), f"Expected one of {patterns} in output. Got:\n{out}"
```

python/sim/ttlsim.py
```diff
@@ -0,0 +1,146 @@
...
+            "ttl": sys.modules.get("ttl"),
+            "ttnn": sys.modules.get("ttnn"),
+        }
```
brnorris03: these are already imported by `setup_simulator_imports`, right? So perhaps not needed here

brnorris03 (commented):
Looking forward to using this!

test/python/utils/test_block_allocation.py
FConstantinos: fixed

examples/metal_examples/multicore_matmul/ttlang/multicore_matmul.py
```diff
@@ -125,3 +130,9 @@ def test_multicore_matmul_tt_lang(M, K, N):
     assert_with_ulp(golden, result)
 
     ttnn.close_device(device)
+
+
+if __name__ == "__main__":
+    # TODO: This won't work with 256, 256, 256
```
zoecarver: Why doesn't it work?

examples/metal_examples/multicore_matmul/ttlang/multicore_matmul.py
```diff
@@ -4,15 +4,20 @@
...
+from ttl import Program, make_circular_buffer_like, copy
+
+from utils.correctness import assert_with_ulp
+from utils.block_allocation import split_work_to_cores
 
 
 def get_number_of_cores(grid_range):
     total_cores = 0
-    for start, end in grid_range:
+    if not len(grid_range) == 0:
```
zoecarver: Can you explain this change?

python/sim/ttlsim.py
FConstantinos: true! Removed

bin/ttlsim
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+# TT-Lang Simulator launcher (ttlsim)
+#
+# Runs tt-lang kernels on the simulator backend without code changes.
+# Usage: ttlsim <file.py|module> [args...]
```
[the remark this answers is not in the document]
FConstantinos: there is already a ttlsim -h option (see ttlsim.py that parses the arguments)

I'm not sure why we want an option to override the default python though.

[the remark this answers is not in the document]
brnorris03: Oh, I don't know that we want to -- I just noticed that you allow it, e.g., `PYTHON=/some/other/python ttsim ...`

test/sim/test_examples.py
```diff
@@ -51,51 +63,36 @@ def run_ttlsim_and_capture(script_path: Path) -> tuple[int, str]:
...
+@pytest.mark.parametrize(
+    "script_name",
+    [
+        "eltwise_add.py",
+        "eltwise_pipe.py",
+        "eltwise_pipe_core3.py",
+        "singlecore_matmul.py",
+        "multicore_matmul.py",
+    ],
+)
```
brnorris03: thank you! 🙂

brnorris03 (approved):
lgtm in general, thanks for the update. I would agree with the try/catch suggestion (around `exec(code, exec_globals)`) -- it would be better to have more information (cleaner errors) than the standard python stack dump.

zoecarver (approved):

PR #231, opened 2026-01-13
A few fixes for multitile hang:

zoecarver:
* remove pure from where it shouldn't be
* CB index uses only innermost loops
* move aquire and release dst reg ops into compute loop


zoecarver:
> Thanks for catching the incorrect placement of the register acquire/release.

To be fair I did not think it was correct and only via splitting into two loops and then merging back together did I realize this fixed the hang.

PR #235, opened 2026-01-14
Claude slash commands 🤖

zoecarver:
Our vision discusses AI tooling and ecosystem as a primary goal and selling point; here it is (or at least the start). 

Claude-code is already the way that programmers write code today, rather than building something bespoke, let's build off what works and is used. These slash commands will integrate nicely with existing developer workflows. They can be used inside of claude code, so the rest of claude's abilities can be combined with these skills, to improve scripts or read existing codebases, run tests, etc. 

Currently we have two commands implemented: 
```
/ttl-import <kernel>
    Translate a CUDA, Triton, or PyTorch kernel to TT-Lang DSL. Analyzes the
    source kernel, maps GPU concepts to Tenstorrent equivalents, and iterates
    on testing until the translated kernel matches the original behavior.
/ttl-export <kernel>
    Export a TT-Lang kernel to TT-Metal C++ code. Runs the compiler pipeline,
    extracts the generated C++, and beautifies it by improving variable names
    and removing unnecessary casts for readable, production-ready output.
```

The idea is that a programmer can open claude code inside their existing codebase and immediately get started with tt-lang.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ TT-Lang bridges this gap through progressive disclosure: simple kernels require 
 
 ## Getting Started with TT-Lang
 
+> ⚠️ Skills are an experimental feature under active development; skills currently reference in-flight functionality that may not be available such as the matmul operator.
+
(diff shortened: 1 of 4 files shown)


brnorris03 (approved):
really cool; should also add a pointer to its README from the top-level README (with necessary disclaimers)

PR #238, opened 2026-01-15
Demo one script

phizalev-TT:
- Added demo_one.py script to examples
- Removed multistep_op.py script as older version of demo_one.py
- Removed the need for examples to print PASSED to be included in sim tests


zoecarver (approved):
LGTM

FConstantinos:
Could you please add it as a test case to the simulator, under test/sim/test_examples.py similar to the others? You could also have versions with errors. Perhaps the compiler would also like to have a similar test case.

phizalev-TT:
> Could you please add it as a test case to the simulator, under test/sim/test_examples.py similar to the others? You could also have versions with errors. Perhaps the compiler would also like to have a similar test case.

Added one without errors in sim test. I suggest errors to be tested in more unit-test way.

PR #239, opened 2026-01-16
[TTL] Implement interval-based DST allocation with register reuse

brnorris03:
### What?
Replaces the simple DST allocator with an interval-based linear scan algorithm that enables register reuse and handles multi-consumer values correctly.

### Why?
The previous allocator implemented a simple reuse heurstic that did not work correctly in all cases (e.g., multiple users of an operand that include unary ops). The new algorithm follows the approach in `docs/development/DST_Allocation.md` based on Wimmer & Franz (CGO'10) linear scan register allocation on SSA form.

### How?
**Phase 1: Copy Insertion**
- Insert `ttl.copy_dst` for values with multiple consumers when any consumer is unary
- Prevents unary operations (which overwrite in-place) from corrupting values needed by other consumers

**Phase 2: Live Interval Analysis**
- Build lifetime intervals [start, end] for each tile value
- Merge intervals for unary operation chains (input and output share same DST register)
- Block arguments start at (first_use - 1) to enable proper sorting

**Phase 3: Linear Scan Allocation**
- Process intervals by start position
- Expire intervals when their end <= next interval's start (enables register reuse)
- Reuse freed registers for new values

Other changes:
- Added `ttl.copy_dst` operation for DST-to-DST copies -- used in Phase 1 to prevent unary ops from clobbering values needed by other consumers
- Replaced `TTLTileAndAssignDST.cpp` with `TTLAssignDST.cpp` implementing full algorithm
- Updated Python API pipeline to use `ttl-assign-dst` pass name

### How to Test?
```bash
ninja -C build check-ttlang-all
```


zoecarver (approved):
Still lgtm!

PR #251, opened 2026-01-20
Remove `return ttl.Program(...)(...)` as it's easy to infer

zoecarver:
When @ttl.compute() or @ttl.datamovement() decorators are applied to nested functions inside a kernel, they register themselves into a module-level list (_thread_registry). This registration happens at decoration time, when the kernel function body executes. These threads are then compiled into a Program internally. The kernel also caches compiled results keyed by (function_id, grid, tensor_shapes_and_dtypes) so repeated calls with the same configuration skip recompilation.


arichinsTT (approved):
Looks good, lotta test changes, thanks for doing that!

examples/eltwise_add.py
```diff
@@ -120,7 +120,6 @@ def dm1():
                 # out_cb.pop() # double pop should fail
 
     # Execute the program across all cores
```
zoecarver: These comments look silly now

FConstantinos:
Removed ttl.Program from simulator

PR #256, opened 2026-01-20
Simulator: Add checks for proper use of blocks

FConstantinos:
- When we ttl.copy from a block to something else, the block cannot be written to until the ttl.copy's wait is finished

 - When we ttl.copy to a block from something else, the block cannot be written or read until the ttl.copy's wait is finished.

### DIFF
--- a/examples/copy_lock_error.py
+++ b/examples/copy_lock_error.py
@@ -113,9 +113,6 @@ def eltwise_add(
                 tx.wait()
                 out_cb.pop()
 
-    # Execute the program across all cores
-    ttl.Program(compute_func, dm0, dm1)(a_in, b_in, out)
-
 
 def main() -> None:
     dim = 256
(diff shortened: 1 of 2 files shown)


python/sim/block.py
```diff
@@ -47,6 +47,11 @@ def __len__(self) -> Size:
 
     @validate_call
     def __getitem__(self, idx: Index) -> Tensor:
+        """Get item with lock checking."""
+        # Import here to avoid circular dependency
+        from .copy import check_can_read
```
zoecarver: hrm might introduce perf overhead just because it's on the subscript... maybe move to separate module?

FConstantinos: Ah good catch! I just moved everything to Block instead, which should be good enough for now.

phizalev-TT:
Will error be accompanied by pointer to source code where it originates?

FConstantinos:
> Will error be accompanied by pointer to source code where it originates?

yes, see this test:
https://github.com/tenstorrent/tt-lang/pull/256/files#diff-5926e92f7a81c8b2dc236db3e38919069e9b631b30529d6e8dc24e909dda3035R158

Issue #87, opened 2025-12-12
[ttl] Lower ttl.wait to TRID-specific barrier


brnorris03:
**Priority: HIGH**
**Parent Issue:** #77

## Description
Lower `ttl.wait` operation to a TRID-specific barrier keyed by the transfer handle instead of a global read barrier.

## Location
`lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp:359`

## Context
In `WaitLowering::matchAndRewrite()`, currently emitting a global `NocAsyncReadBarrierOp`. Need to:
1. Extract the transfer handle from the wait operation
2. Use it to key a TRID-specific barrier
3. Only synchronize on the specific transfer(s) being waited on

## Related Subtasks
Part of the TRID Barrier Support group in #77.

brnorris03:
Still open. Current lowering emits global barriers only (`NocAsyncReadBarrierOp`/`NocAsyncWriteBarrierOp` in `WaitLowering`, [ConvertTTLToTTKernel.cpp:1141-1170](https://github.com/tenstorrent/tt-lang/blob/ce20bdc309bdba6e022af59cf32fc0d79ef410dd/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp#L1141-L1170)); no TRID barrier ops, `use-trid-barriers` option, or `TridAllocator` exist on main yet. TRID support is implemented in PR #267 (behind a default-off `use-trid-barriers` option, with a 16-slot TRID allocator and evict barriers), currently approved but not yet merged. Keep open until #267 merges.

PR #267, opened 2026-01-23
[ttl] Make TRID DMA wait lowering selectable (default: global barriers)

shutovilyaep:
# PR #267: [ttl] Make TRID DMA wait lowering selectable (default: global barriers)

## Why

Current lowering uses global DMA barriers. TRID-scoped waits needed for #87 are not supported. Default behavior must stay unchanged for existing users.

## What

- New pass option `use-trid-barriers` (default off): choose global barriers (current) or TRID barriers.
- When on: copy lowering emits set_trid; wait lowering emits barrier_with_trid. TRID slots (16) are tracked; reuse of an in-flight slot triggers an evict barrier first.
- Pipeline and TTKernel cleanup forward the option; TRID dedup patterns run only in TRID mode.
- Tests: conversion, translation, ME2E, and Python lit updated so both modes are covered (default path kept; TRID path added or enabled where needed).

## How

- Pass and pipeline accept the option and pass it through. In TRID mode, transfer handles become i32; SCF type conversions keep regions legal.
- TridAllocator tracks 16 slots and direction; before reusing a busy slot the lowering emits the matching barrier. In global mode the allocator is unused (handle is 0).
- Cleanup: TRID dedup is registered only when the option is true; dedup compares TRID and NOC.
- Tests use the option explicitly where TRID output is checked; translation tests keep default RUN and add a second RUN with TRID checks where output differs.


brnorris03 (commented):
Looks great, thank you! The only more significant issue I see is the lack of runtime tests, I think the best approach for now is to parameterize (some of) the `test/me2e` tests with the new option, what do you think? I can help with more concrete suggestions on how to do that if you agree.dd

Some general questions, mainly stemming from my lack of deep knowledge of the low-level semantics of the metal ops.

1. Is the TRID value semantically meaningful, or just needs to be unique per copy? I am guessing order doesn't matter? As defined the generated TRIDs could be nondeterministic (but correctly unique) due to parallel pattern application.

2. With the new ops requiring explicit NOC, I see that NOC 0 is always used -- is this appropriate or something that needs to be generalized (perhaps later PR)? 

Again, thank you for contributing this!!

shutovilyaep:
@brnorris03 Hello, the fixes were ready about 3 weeks ago, haven't pushed due to being a little bit off due to sudden layoff from Tenstorrent. Will try to complete this today, if still relevant

lib/Dialect/TTKernel/Transforms/TTKernelCleanupPatterns.cpp
```diff
@@ -31,13 +32,54 @@ struct DeduplicateConsecutiveBarriers : OpRewritePattern<BarrierOp> {
...
   patterns.add<DeduplicateConsecutiveBarriers<NocAsyncReadBarrierOp>>(
       patterns.getContext());
   patterns.add<DeduplicateConsecutiveBarriers<NocAsyncWriteBarrierOp>>(
       patterns.getContext());
+  patterns
+      .add<DeduplicateConsecutiveTridBarriers<NocAsyncReadBarrierWithTridOp>>(
+          patterns.getContext());
+  patterns
+      .add<DeduplicateConsecutiveTridBarriers<NocAsyncWriteBarrierWithTridOp>>(
+          patterns.getContext());
```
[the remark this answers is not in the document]
shutovilyaep: fixed:
- populateTTKernelCleanupPatterns now takes useTridBarriers (default false in header); TRID dedup patterns are only added when true.
- Convert pass calls it with useTridBarriers so the option is forwarded to cleanup.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -539,14 +553,58 @@ static Value linearizeTileIndex(OpBuilder &builder, Location loc, Value row,
...
+    if (outstanding[trid]) {
+      result.evictDirection = direction_[trid];
+    }
+    outstanding[trid] = true;
+    direction_[trid] = direction;
+    ++nextTrid;
+    return result;
+  }
+
+  void releaseTrid(uint32_t trid) { outstanding[trid % kNumTrids] = false; }
```
brnorris03: Is it used anywhere? If not, remove

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -690,6 +762,38 @@ struct CopyLowering : OpConversionPattern<CopyOp> {
       });
     }
 
+    if (!tridAllocator) {
+      return rewriter.notifyMatchFailure(op, "missing TRID allocator");
+    }
```
brnorris03: nit: dead code -- the allocator currently is always constructed; I think an assert is more appropriate here or llvm_unreachable.

brnorris03 (commented):
Apologies for taking so long with the re-review. I think overall looks great, my main concerns at the moment are about the test coverage, should be easily addressable. So sorry to learn about the layoff (if you don't mind, can you email me so I have your contact info).

brnorris03 (approved):
Ready to land after some deprecated builder usage is updated (e.g., `rewriter.create<arith::ConstantIntOp>(loc, 0, 8);` should be `arith::ConstantIntOp::create(rewriter, loc, 0, 8);`, same for all rewriter.create calls). 
Thank you!

shutovilyaep:
@brnorris03 Pushed removing unneeded "makeZeroI32" with a history rewrite to remove it from git at all - noticed CI broke due to unused function, please take a look

shutovilyaep:
Hello @brnorris03, please take a look.


**FYI:** The four post-main-to-branch-merge-commit fix commits were prepared ~19 hours ago (Jun 18). `main` moved once again since then, so I merged it once more (`ff1f70c`).

---

Merged latest `main` into `feat/lower_copy_wait`. The only real conflict was in `ConvertTTLToTTKernel.cpp`: upstream added pipe-transfer lowering and related helpers; this branch adds TRID copy/wait lowering. Both paths coexist. Default behavior is unchanged (`use-trid-barriers=false` → global NOC barriers).

Four follow-up commits, all mechanical:

- **clang-format** after conflict resolution in `ConvertTTLToTTKernel.cpp`
- **remove `makeZeroI32`** — dead code left over from TRID handle lowering
- **update TRID lit tests** — input IR only, to match merged `main` TTL syntax (`#ttl.layout`, `block_count`, tile CB types); barrier CHECKs unchanged
- **restore `pack_tile<true>`** in five Python lit tests — the branch had a stale `false` from an old commit that wrongly tied TRID mode to `pack_tile` expectations; TRID affects copy/wait only, not store lowering. Main always had `true`

No compiler behavior change in those four commits.

PR #269, opened 2026-01-23
Simulator: Add ability to do broadcasts

FConstantinos:
- Adds necessary changes to support broadcast example
- intermediate results between blocks are now block objects as well (marked as temporary and not backed by CBs)


phizalev-TT (approved):
LGTM!

PR #293, opened 2026-01-29
Fix bcast indexing

zoecarver:
The code used linearized indexing (row * numCols + col) for all cases, but for shape expansion:                                                
  - Col broadcast with (N,1) input → (N,M) output should use just row index                                                                                  
  - Row broadcast with (1,M) input → (N,M) output should use just col index                                                                                  
  - Scalar broadcast with (1,1) input → (N,M) output should use constant 0


brnorris03 (approved):
I am not a fan of the 2D shape hardcoding but given there is no time to iterate, I am not going to hold this back.

zoecarver:
> I am not a fan of the 2D shape hardcoding but given there is no time to iterate, I am not going to hold this back.

I can try to improve it in the follow up PR that fixes your outstanding comments from the first bcast change.

PR #297, opened 2026-01-30
Simulator: add various methods for broadcasting

FConstantinos:
For ‘a' MxN and ‘b’ Mx1 the following should work:

1) y.store(a * b )

2) y.store(a * broadcast(b, dims=[1]))

3) y.store( a * broadcast(b, y, dims=[1]))

4) w.store( broadcast(b, y, dims=[1]) ); y.store(a*w)

y = anything will not work because the simulator does not support direct assignment


phizalev-TT (approved):
LGTM!

PR #304, opened 2026-02-02
Fix store lowering

zoecarver:
Add a frontend store op and lower it all the way to pack tile; a few related fixes. 

This is a temporary fix, Boyana may have a cleaner approach, using this branch to unblock claude skills.

### DIFF
--- a/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
@@ -1012,10 +1012,11 @@ lowerTTLOpsToTTKernel(ModuleOp mod, MLIRContext &ctx,
   });
 
   RewritePatternSet patterns(&ctx);
-  patterns.add<BindCBLowering, TensorSliceLowering, CopyLowering, WaitLowering,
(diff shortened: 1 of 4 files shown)


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -369,10 +369,15 @@ struct StoreLowering : OpConversionPattern<StoreOp> {
               defOp->getAttrOfType<IntegerAttr>(kDstIdxAttrName)) {
         dstIndex =
             rewriter.create<arith::ConstantIndexOp>(loc, dstIdxAttr.getInt());
+      } else if (auto copyTile = dyn_cast<CopyTileOp>(defOp)) {
+        // Fallback: get dst_index directly from copy_tile operand
+        dstIndex = copyTile.getDstIndex();
+      } else {
```
brnorris03: Is this necessary? The assign dst pass should be setting the dst index for copy_tile ops -- so if the attribute is missing, then that should be an error since it may indicate that DST allocation may be invalid (e.g., some other pass created a copy_tile op that was not considered during DST allocation).

test/pytest.ini
brnorris03: Was this intentionally deleted?

PR #314, opened 2026-02-13
Explicit stores, cb lifetime ops, and side-effect-only loops

zoecarver:
This reworks the lowering pipeline so that stores and cb lifetime ops are explicit from Python all the way down, and loops don't carry dead tensor dataflow.

  - `ttl.store` / `ttl.tile_store` split: The old `ttl.store` is renamed to `ttl.tile_store` (tile-level, lives inside compute bodies). A new tensor-level `ttl.store` is added, emitted directly by Python's `o.store(result)`. This means the user's store intent is in the IR from the start rather than being reconstructed later from yields.
  - `LowerStoreToCompute` pattern: `convert-ttl-to-compute` now has a pattern that finds each `ttl.store`, locates the compute op that produces its input, and sinks to a `ttl.tile_store` in a `ttl.compute` block (this is the same logic we use to fuse eltwise ops). 
  - `TTLInsertTileRegsSync` no longer walks yields, scans for reserves, or synthesizes stores. It just inserts acquire/commit/wait/release around the existing `tile_store` ops. Went from ~250 lines of store inference down to straightforward insertion logic.
  - Side-effect-only loops: `ConvertTTLComputeToSCF` generates `scf.for` without `iter_args` for outputs. No `tensor.insert`, no yielding updated tensors. Stores are explicit side effects via `tile_store`. This also simplifies `ConvertTTLToTTKernel` cleanup since there are no `tensor.insert` ops to tear down anymore.
  - DST assignment refactored: `TTLAssignDST` now inserts all `copy_tile` ops at the block start (reverse arg order) via a shared `createCopyTileForArg` utility, instead of inserting at first-use with duplicated allocation logic.

### DIFF
--- a/test/python/test_store_patterns.py
+++ b/test/python/test_store_patterns.py
@@ -26,11 +26,8 @@ def passthrough_kernel(inp, out):
 
     @ttl.compute()
     def compute():
-        x = inp_cb.wait()
-        o = out_cb.reserve()
-        o.store(x)
-        inp_cb.pop()
-        out_cb.push()
+        with inp_cb.wait() as x, out_cb.reserve() as o:
+            o.store(x)
 
     @ttl.datamovement()
     def dm_read():


brnorris03 (approved):
LGTM, thank you for the updates!

PR #320, opened 2026-02-19
Debug tools and profiler fix

zoecarver:
* adds perf/debug tools for dataflow, noc events, and pipes (when they land)
* fixes bug with signposts where they would have hash conflicts on larger kernels

### DIFF
--- a/test/python/auto_profile_signposts.py
+++ b/test/python/auto_profile_signposts.py
@@ -64,35 +64,39 @@ def signpost_test_kernel(inp, out):
 # CHECK: #include "tools/profiler/kernel_profiler.hpp"
 # CHECK: void kernel_main()
 
-# Check for cb_wait signpost (scoped around the operation)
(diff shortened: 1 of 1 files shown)


python/ttl/ttl_api.py
```diff
@@ -203,6 +203,48 @@ def _run_profiling_pipeline(
...
+    """
+    Run NOC profiler summary and print CB flow / pipe graph after execution.
+
+    Called after kernel execution when TTLANG_PERF_DUMP=1 is set.
+    Reads NOC traces from $TT_METAL_HOME/generated/profiler/.logs/,
+    CB flow graph from /tmp/ttlang_cb_flow_graph.json (written by
+    ttl-dump-cb-flow-graph pass), and pipe graph from
+    /tmp/ttlang_pipe_graph.json (copied from compiler temp file).
+    """
+    from ._src.noc_summary import run as noc_summary_run
```
phizalev-TT: So we support both standalone `noc_summary` script and it integrated with performance dump?

PR #324, opened 2026-02-19
Simulator Refactoring

FConstantinos:
Just some refactoring, mainly DRY and resolving circular dependencies

### DIFF
--- a/examples/eltwise_pipe.py
+++ b/examples/eltwise_pipe.py
@@ -6,7 +6,7 @@ import math
 
 import ttl
 import ttnn
-from sim.typedefs import Pipe
+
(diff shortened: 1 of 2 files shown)


FConstantinos:
> Is there any other cleanup that we need to do between running examples? Previously, we spawned a sub processes, so I think things like `_pipe_buffer` would be cleaned up automatically, but now the full module persists between runs, so we might accumulate state.

Thank you for mentioning this :) I think this is a deeper problem. I'd prefer if there was no need for cleanup in the first place, so removing global state and passing it around as arguments etc. For now I've simply reverted the change. Also, other tests are (probably) affected by this lack of cleanup

PR #340, opened 2026-02-26
Perfetto server to serve profiling results from signpost or auto profile 

zoecarver:
* works on server directly and inside docker container
* instructions to forward the server to your local mac
* gives URL to open to see results

Example:
```
======================================================================
TTLANG PERFETTO TRACE SERVER
======================================================================
  419 trace events ready
  Serving on port 48019

  From your local machine, run:
    ssh -N -L 48019:172.17.0.2:48019 user@<server>

  Then open:
    http://localhost:48019

  Press Enter to stop the server...
======================================================================
```


phizalev-TT (approved):
LGTM! This is surprisingly simple for what it does!

Let's document all current profiling options and env variables (separate PR).

PR #342, opened 2026-02-27
[ttl] Op classification infrastructure + AssignDST changes for FPU

brnorris03:
Downstream passes in the DST maximization series need to classify tile ops by kind (copy, binary, unary, broadcast, FPU binary) and query whether an op reads from CB vs DST. This PR adds the traits, enums, and utilities that make those queries possible without ad-hoc `isa<>` chains scattered across each pass.

The main behavioral addition is the FPU binary op detection in `TTLAssignDST`: binary add/sub/mul whose operands both come from CB block arguments are marked `ttl.fpu_binary`, which tells later passes they consume zero DST input slots. This is controlled by the `--enable-fpu-binary-ops` option (default true for the pass, but **disabled in all three pipeline entry points** until PR 6 adds the FPU lowering patterns), so existing tests and hardware runs are unaffected in this PR.

Other AssignDST improvements: mixed-dtype detection now hard-errors instead of warning, in-place terminology replaces stale "unary" naming, and `ttl.unroll_factor` is computed and attached for the subblocking pass to consume.

This is PR 1 of 8 splitting the [bnorris/max-dst](https://github.com/tenstorrent/tt-lang/pull/316) branch:
  1. Op classification infrastructure + AssignDST enhancements <- this PR
  2. ComputeOp TilingInterface + subblocking pass
  3. Init extraction refactor (move init emission to TTKernelInsertInits)
  4. DST sync insertion + loop lowering + CB indexing
  5. Scheduling pass (TTLScheduleOperations)
  6. FPU binary lowering patterns
  7. Pipeline wiring + Python API + integration tests
  8. Documentation ([DST_Allocation.md](https://github.com/tenstorrent/tt-lang/blob/bnorris/max-dst/docs/development/DST_Allocation.md), [DST_Utilization.md](https://github.com/tenstorrent/tt-lang/blob/bnorris/max-dst/docs/development/DST_Utilization.md))

### DIFF
--- a/docs/sphinx/specs/TTLangSpecification.md
+++ b/docs/sphinx/specs/TTLangSpecification.md
@@ -18,6 +18,7 @@
 * [9. Semaphore](#9-semaphore)
 * [Appendix A. Glossary](#appendix-a-glossary)
 * [Appendix B. Block operators and math functions](#appendix-b-block-operators-and-math-functions)
+* [Appendix C. Naming guidelines](#appendix-c-naming-guidelines)
 
 ## 0. Versions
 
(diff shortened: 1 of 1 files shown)


zoecarver (approved):
LGTM!

PR #427, opened 2026-03-25
add bf8 support

zoecarver:
Simple code change: just update the mapping

### DIFF
--- a/test/python/test_bfp8_dram_add.py
+++ b/test/python/test_bfp8_dram_add.py
@@ -13,6 +13,7 @@
 import torch
 import ttnn
 import ttl
+from utils.correctness import assert_allclose
 
 
 @ttl.kernel(grid=(1, 1))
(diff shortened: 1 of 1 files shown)


test/python/test_bfp8_dram_add.py
```diff
@@ -0,0 +1,111 @@
...
+    print(f"Expected ~5.0")
+
+    # bfp8 has lower precision, use relaxed tolerance
+    expected = lhs_torch + rhs_torch
+    if torch.allclose(result.float(), expected.float(), rtol=0.1, atol=0.5):
+        print("\nPASS: BFP8 DRAM interleaved add works!")
+        # CHECK: PASS
+    else:
+        max_diff = (result.float() - expected.float()).abs().max().item()
+        print(f"\nFAIL: max difference = {max_diff}")
```
brnorris03: Could we use our `correctness.py` allclose here (and update as needed for the new types)?

brnorris03 (commented):
looks great, mainly wondering whether our correctness.py also needs updating.

PR #435, opened 2026-03-28
[spec] Improve examples etc

phizalev-TT:
- Limit block shapes for tiled tensors to be at least two-dimensional
- Add pure TT-NN example showing shape translation from Torch shape to tile and row-major shapes.
- Redo elementwise example to show `broadcast` and `reduce_sum`
- Make matmul example use `I_BLOCK_SIZE` for `I`, `M_BLOCK_SIZE` for `M `and `N_BLOCK_SIZE` for `N`
- Fix matmul example to unsqueeze and broadcast `b_blk` and `c_blk`
- Rewamp specifications for `broadcast` and `reduce_xxx` to expect resulting shape, make them not-supported for row-major blocks (for now)
- Rewamp specification for `fill` to expect resulting shape
- Rewamp specification for `__matmul__` to support more that two-dimension inputs
- Add `squeeze` and `unsqueeze`
- Specify that `block_count` defaults to 2
- Do not use abbreviated loop bounds and variables
- Do not shadow tensor variables in example
- Add more explanation in comments
- Clarify scalar as scalar element or value
- Unique example headers
- Use naming guidelines


docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,202 +227,323 @@ def some_compute():
...
+        # then it broadcasts column of M_TILES tiles to get N_BLOCK_SIZE×M_TILES tiles
+        c_squared_bcast = ttl.math.broadcast(c_squared, dims=[0], shape=(N_BLOCK_SIZE, M_TILES))
+
+        # Broadcast d_squared along all dimensions (0 and 1) to N_BLOCK_SIZE×M_TILES;
+        # This first broadcasts single scalar value at position (0, 0) to fill a single tile
+        # then it broadcasts single tile to get N_BLOCK_SIZE×M_TILES tiles
+        d_squared_bcast = ttl.math.broadcast(d_squared, dims=[0, 1], shape=(N_BLOCK_SIZE, M_TILES))
+
+        # Zero-initialize the accumulator z before summing N_BLOCKS partial sums
+        z = ttl.math.fill(0, shape=(1, M_TILES))
```
FConstantinos: In the appendix, the order of arguments is fill(shape, value) but here it's fill(value, shape)

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,202 +227,323 @@ def some_compute():
...
+            for _ in range(N_BLOCKS):
 
+                # Reserve y_blk
                 with y_dfb.reserve() as y_blk:
 
-                    # acquire c_blk from c_dfb:
-
-                    y = ttl.math.fill(y_blk, 0)
+                    # Zero-initialize the accumulator y before summing K_BLOCKS partial products
+                    y = ttl.math.fill(0, shape=(I_BLOCK_SIZE, M_BLOCK_SIZE, N_BLOCK_SIZE))
```
FConstantinos: see above

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1098,11 +1248,17 @@ def matmul_read():
...
 | `ttl.math.mask_posinf(expr: ttl.BlockExpr, mask: ttl.BlockExpr) -> ttl.BlockExpr` | Mask a block with specified `mask` by replacing masked (corresponding mask element equals to 1) elements with positive infinity. |
 | `ttl.math.where(condition: ttl.BlockExpr, true_value: ttl.BlockExpr, false_value: ttl.BlockExpr) -> ttl.BlockExpr` | For each element in specified condition block return the corresponding element from `true_value` if true (condition element equals to 1) or the element from `false_value` if false (condition element equals to 0) |
 
+### Shape manipulation functions
+
+| Function | Description |
+| :---- | :---- |
+| `ttl.math.squeeze(expr: ttl.BlockExpr, dim: int) -> ttl.BlockExpr` | Remove shape dimension at position specified by `dim`. Removed shape dimension must be 1.<br><br>Example for squeeze over dimension 0 (outermost): `ttl.math.squeeze(a, dim=0)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(1, N, M)`.<br><br>Example for squeeze over dimension -1 (innermost): `ttl.math.squeeze(a, dim=-1)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(N, M, 1)`. |
```
FConstantinos: I think the examples of squeeze and unsqueeze need to be swapped.

Also it's worth noting for the unsqueeze to mention that a positive dim means "insert before" while a negative one means "insert after" the position indicated by the index

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,202 +227,323 @@ def some_compute():
...
 
+                        # Wait for a_blk and b_blk to be loaded and pushed by matmul_read
                         with (
                             a_dfb.wait() as a_blk,
                             b_dfb.wait() as b_blk,
                         ):
+                            # b_blk has shape K_BLOCK_SIZE×N_BLOCK_SIZE;
+                            # Unsqueeze it to 1×K_BLOCK_SIZE×N_BLOCK_SIZE and then
+                            # broadcast it over dim 0 to I_BLOCK_SIZE×K_BLOCK_SIZE×N_BLOCK_SIZE
+                            b = ttl.math.broadcast(ttl.math.unsqueeze(b_blk, dim=0), dims=[0], shape=(I_BLOCK_SIZE, M_BLOCK_SIZE, N_BLOCK_SIZE))
```
FConstantinos: This is not consistent with the comment that says it's K_BLOCK_SIZE at that dimension

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,202 +227,323 @@ def some_compute():
...
+
+a_dfb = ttl.make_dataflow_buffer_like(a, shape = (N_BLOCK_SIZE, M_TILES))
+
+# Tiled DFB shape needs to be at least two-dimensional;
+# When tiled the one-dimensional vector b is placed in row 0
+# of each tile in a row of N_TILES tiles
+b_dfb = ttl.make_dataflow_buffer_like(b, shape = (N_BLOCK_SIZE, 1))
+# When tiled the one-dimensional vector c is placed in column 0
+# of each tile in a column of M_TILES tiles
+c_dfb = ttl.make_dataflow_buffer_like(b, shape = (1, M_TILES))
```
FConstantinos: should the likeness_tensor be 'c' maybe?

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,202 +227,323 @@ def some_compute():
...
+# When tiled the scalar value d is placed at position (0, 0)
+# of a single tile
+d_dfb = ttl.make_dataflow_buffer_like(b, shape = (1, 1))
```
FConstantinos: Should the likeness_tensor be 'd' maybe?

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1075,10 +1225,10 @@ def matmul_read():
...
...ifies one or both of last (innermost) dimensions; <br><br>For row-major blocks reduction is not supported.<br><br>Example for reduction over dimension -1 (innermost): `ttl.math.reduce_sum(a, dims=[-1], shape=(N, 1))`. Here if `a` has shape of `(N, M)` and therefore the result will have the shape of `(N, 1)`. In step (1) M rows of tiles in `a` are elementwise reduced to a single row of tiles. In step (1) each tile has all of its rows reduced to row 0.<br><br>Example for reduction over dimension 0 (outermost): `ttl.math.reduce_max(a, dims=[0], shape=(1, M))`. Here if `a` has shape of `(N, M)` and therefore the result will have the shape of `(N, 1)`. In step (1) N columns of tiles in `a` are elementwise reduced to a single column of tiles. In step (1) each tile has all of its columns reduced to column 0.<br><br>Example for reduction over two innermost dimensions: `ttl.math.reduce_sum(a, dims=[-1, -2], shape=(1, 1))`. Here if `a` has shape of `(N, M)` and therefore the result will have the shape of `(1, 1)`.  In step (1) M rows and N columns of tiles in `a` are elementwise reduced to a single tiles In step (1) a tile has all of its scalar values reduced to a value in position (0, 0). |
```
FConstantinos: > Here if `a` has shape of `(N, M)` and therefore the result will have the shape of `(N, 1)`

remove "and therefore", same for dims=[0], dims=[-1, -2] cases

> In step (1) M rows of tiles in `a` are elementwise reduced to a single row of tiles. In step (1) each tile has all of its rows reduced to row 0

I think the second step refers to step (2), not (1)

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1075,10 +1225,10 @@ def matmul_read():
...
...ns; (2) whole tiles get broadcasted along specified dimensions.<br><br>For row-major blocks broadcast is not supported.<br><br>Example for broadcast over dimension -1  (innermost): `ttl.math.broadcast(a, dims=[-1], shape=(N, M))`. Here the shape of the result is `(N, M)` and therefore `a` must have the shape of `(N, 1)`. In step (1) each tile in `a` has its row 0 broadcasted over the rest of rows. In step (2) whole row of tiles is broadcasted M times.<br><br>Example for broadcast over dimension 0 (outermost): `ttl.math.broadcast(a, dims=[0], shape=(N, M))`. Here the shape of the result is `(N, M)` and therefore `a` must have the shape of `(1, M)`. In step (1) each tile in `a` has its column 0 broadcasted over the rest of columns. In step (2) whole column of tiles is broadcasted N times.<br><br>Example for broadcast over two innermost dimensions: `ttl.math.broadcast(a, dims=[-1, -2], shape=(N, M))`. Here the shape of the result is `(N, M)` and therefore  `a` must have the shape of `(1, 1)`. In step (1) each tile in `a` has a scalar value in the position (0, 0) broadcasted over the rest positions. In step (2) whole tile is broadcasted N times column-wise and M times row-wise. times.|
```
FConstantinos: > Input shape must contain 1 in dimensions specified for broadcast.

By "input shape",  should we clarify that we mean the shape of expr, to avoid confusion with the other "shape" argument?

Also, if expr.shape contains 1 at the positions described in dims, but the target shape also contains 1 at some of those same positions, is it the case that this is an error? I'm looking at it as an attempt to broadcast on non-broadcastable dimensions. This comes back to the discussion we had on failing early a while ago. If it is an error, maybe it's worth mentioning this?

FConstantinos: > (1) scalar values within tiles get broadcasted along specified dimensions

also:

> In step (1) each tile in `a` has its row 0 broadcasted over the rest of rows

It is conceivable that within the tile, some of the rows that are not row 0 have non-zero values within them. Are we comfortable with the fact that those values will be overwritten with whatever is the corresponding element in row 0?

FConstantinos: Also, row 0 could have heterogeneous values, e.g it could be [1, 2, 3,... 32]

broadcasting this to the tile would make the tile look like:

tile = [ [1, 2, 3,... 32]; ... [1, 2, ..., 32] ]

And then all tiles we broadcast to will look like the tile above. Is that the intention?

FConstantinos: > In step (2) whole tile is broadcasted N times column-wise and M times row-wise. times.|

Remove the last "times."

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1098,11 +1248,17 @@ def matmul_read():
...
 
+### Shape manipulation functions
+
+| Function | Description |
+| :---- | :---- |
+| `ttl.math.squeeze(expr: ttl.BlockExpr, dim: int) -> ttl.BlockExpr` | Remove shape dimension at position specified by `dim`. Removed shape dimension must be 1.<br><br>Example for squeeze over dimension 0 (outermost): `ttl.math.squeeze(a, dim=0)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(1, N, M)`.<br><br>Example for squeeze over dimension -1 (innermost): `ttl.math.squeeze(a, dim=-1)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(N, M, 1)`. |
+| `ttl.math.unsqueeze(expr: ttl.BlockExpr, dim: int) -> ttl.BlockExpr` | Add shape dimension of 1 at position specified by `dim`.br><br>Example for unsqueeze over dimension 0 (outermost): `ttl.math.unsqueeze(a, dim=0)`. Here if the shape of `a` is `(1, N, M)` the shape of the result will be `(N, M)`.<br><br>Example for unsqueeze over dimension -1 (innermost): `ttl.math.unsqueeze(a, dim=-1)`. Here if the shape of `a` is `(N, M, 1)` the shape of the result will be `(N, M)`. |
```
FConstantinos: > Add shape dimension of 1 at position specified by `dim`.br>

I think that last br> needs an opening "<"

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -153,13 +154,39 @@ x, y, z = ttl.node(dims = 3)
...
+shape_in_tiles(from_torch(torch.randn((2, 128, 32)))) #    prints [2, 4, 1]
+shape_in_tiles(from_torch(torch.randn((2, 2, 128, 32)))) # prints [2, 2, 4, 1]
+```
+
+If tensor has a row-major layout the shape unit is a scalar element. For the TT-NN tensor in the above example the corresponding block that fits this entire tensor will have shape of `(2, 128, 32)`.
 
 Shape determines the shape of a *block* returned by one of the *acquisition functions*. The size of a block in L1 memory is determined by shape, shape unit and data type. For example, for a block with shape `(2, 4, 1)`, shape unit of a tile and BF16 data type, its size in L1 will be `2 * 4 * 32 * 1 * 32 * 2 = 16384` bytes. The block count determines the total size of L1 memory allocated for a dataflow buffer. This size as a product of a block size and block count. For the most common case block count defaults to 2 to support double buffering. With double buffered dataflow buffer one kernel can write to a block while another is reading from a block thus enabling enabling the pipelining. For the example above, this means there will be a total of 32768 bytes of L1 memory allocated for the dataflow buffer.
```
FConstantinos: > reading from a block thus enabling enabling the pipelining

remove duplicate *enabling

docs/sphinx/specs/TTLangSpecification.md
FConstantinos: Maybe "copy blk_to_send to the pipe"

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: > Also, if expr.shape contains 1 at the positions described in dims, but the target shape also contains 1 at some of those same positions, is it the case that this is an error? I'm looking at it as an attempt to broadcast on non-broadcastable dimensions. This comes back to the discussion we had on failing early a while ago. If it is an error, maybe it's worth mentioning this?

No, this is not an error. In this case only in-tile (step (1)) broadcast will be done. I clarified for both step (2) in broadcast and step (1) in reduce.

phizalev-TT: Yes.

phizalev-TT: This is the intention. Row 0 first gets copied to all other rows. Then the tile get copied to match the shape. If we look at the block all columns will contain same values across tiles.

FConstantinos (approved):

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -153,43 +154,70 @@ x, y, z = ttl.node(dims = 3)
...
+shape_in_tiles(from_torch(torch.randn((128)))) #           prints [1, 4]
+shape_in_tiles(from_torch(torch.randn((1, 128)))) #        prints [1, 4]
+shape_in_tiles(from_torch(torch.randn((32, 128)))) #       prints [1, 4]
+shape_in_tiles(from_torch(torch.randn((128, 1)))) #        prints [4, 1]
+shape_in_tiles(from_torch(torch.randn((128, 32)))) #       prints [4, 1]
+shape_in_tiles(from_torch(torch.randn((2, 128, 32)))) #    prints [2, 4, 1]
+shape_in_tiles(from_torch(torch.randn((2, 2, 128, 32)))) # prints [2, 2, 4, 1]
+```
+
+If tensor has a row-major layout the shape unit is a scalar element. For the TT-NN tensor in the above example the corresponding block that fits this entire tensor will have shape of `(2, 128, 32)`.
```
brnorris03: There are quite a few tensors in the above example, which one is this referring to?

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -153,43 +154,70 @@ x, y, z = ttl.node(dims = 3)
...
+
+Shape determines the shape of a *block* returned by one of the *acquisition functions*. The size of a block in L1 memory is determined by shape, shape unit and data type. For example, for a block with shape `(2, 4, 1)`, shape unit of a tile and BF16 data type, its size in L1 will be `2 * 4 * 32 * 1 * 32 * 2 = 16384` bytes. The block count determines the total size of L1 memory allocated for a dataflow buffer. This size as a product of a block size and block count. For the most common case block count defaults to 2 to support double buffering. With double buffered dataflow buffer one kernel can write to a block while another is reading from a block thus enabling the pipelining. For the example above, this means there will be a total of 32768 bytes of L1 memory allocated for the dataflow buffer.
```
brnorris03: Suggest "returned by one of the *acquisition functions*" -> "returned by one of the *acquisition functions* (`wait`, `reserve`)"

"With double buffered dataflow buffer" -- we use different terminology in the actual methods (block count), so may be good to use/introduce that here, too? Also, may be worth mentioning that kernel-local dfb actually require double buffering.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,330 @@ def some_compute():
...
-b_dfb = ttl.make_dataflow_buffer_like(B, shape = (1, ))
-y_dfb = ttl.make_dataflow_buffer_like(Y, shape = (1, ))
-z_dfb = ttl.make_dataflow_buffer_like(Z, shape = (1, ))
+# z[m] = ∑(√(a[n, m]² - b[n]² - c[m]² - d²))
+#        i
+#
+# Tensor   Torch shape   Note
+# a        N, M          N >> M
+# b        N, 1          One-dimensional vector — broadcast to match a along M
+# c        M             One-dimensional vector — broadcast to match a along N
```
brnorris03: nit: vector is one-dimensional by definition, so just "vector"

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,330 @@ def some_compute():
...
 def elwise_read():
-    for nt in range(NT):
 
-        # acquire a_blk and b_blk from a_dfb and b_dfb:
+    # Reserve c_blk and d_blk blocks
+    with (
+        c_dfb.reserve() as c_blk,
+        d_dfb.reserve() as d_blk,
+    ):
+        # Load entire (1×M_TILES) c
```
brnorris03: Nit: " c" -> " of c"

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,330 @@ def some_compute():
...
+                z_partial = ttl.math.reduce_sum(expanded_z, dims=[0], shape=(1, M_TILES))
 
-            y_blk.store(y)
-            z_blk.store(z)
+                # Store y
+                y_blk.store(y)
 
-            # release a_blk, b_blk and y_blk
+                # Accumulate-add partial z
+                z += z_partial
```
brnorris03: technically the compiler should support correct += accumulation without the initial explicit fill with 0s; but probably not worth the minimal performance difference (also not sure if the sim supports that)

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,330 @@ def some_compute():
...
 
+                # Pop a_blk and b_dfb to make them available for elwise_read to load and push next blocks;
+                # Push y_blk to make it ready for elwise_write
```
brnorris03, line 410: Are these comments referring to the implicit pop and push? Seems a bit confusing seeing them without any actual pops and push calls

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,330 @@ def some_compute():
...
 
-        # acquire y_blk and z_blk from y_dfb and z_dfb:
+    # Wait for elwise_compute to store and push z_blk
+    with z_dfb.wait() as z_blk:
 
-        with (
-            y_dfb.wait() as y_blk,
-            z_dfb.wait() as z_blk,
-        ):
+        # Store entire (1xM_TILES) z
```
brnorris03: nit: "of z"

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1075,10 +1222,10 @@ def matmul_read():
...
...but only if reduction specifies one or both of last (innermost) dimensions; <br><br>For row-major blocks reduction is not supported.<br><br>Example for reduction over dimension -1 (innermost): `ttl.math.reduce_sum(a, dims=[-1], shape=(N, 1))`. Here if `a` has shape of `(N, M)` then the result will have the shape of `(N, 1)`. In step (1) M rows of tiles in `a` are elementwise reduced to a single row of tiles. In step (2) each tile has all of its rows reduced to row 0.<br><br>Example for reduction over dimension 0 (outermost): `ttl.math.reduce_max(a, dims=[0], shape=(1, M))`. Here if `a` has shape of `(N, M)` then the result will have the shape of `(N, 1)`. In step (1) N columns of tiles in `a` are elementwise reduced to a single column of tiles. In step (2) each tile has all of its columns reduced to column 0.<br><br>Example for reduction over two innermost dimensions: `ttl.math.reduce_sum(a, dims=[-1, -2], shape=(1, 1))`. Here if `a` has shape of `(N, M)` then the result will have the shape of `(1, 1)`.  In step (1) M rows and N columns of tiles in `a` are elementwise reduced to a single tiles. In step (2) a tile has all of its scalar values reduced to a value in position (0, 0). |
```
brnorris03, line 1272: nit: there are `<br><br>` in this block of text, but would be better to have actual new lines, too, for easier change viewing

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1098,11 +1245,17 @@ def matmul_read():
...
+| :---- | :---- |
+| `ttl.math.squeeze(expr: ttl.BlockExpr, dim: int) -> ttl.BlockExpr` | Remove shape dimension at position specified by `dim`. Removed shape dimension must be 1.<br><br>Example for squeeze over dimension 0 (outermost): `ttl.math.squeeze(a, dim=0)`. Here if the shape of `a` is `(1, N, M)` the shape of the result will be `(N, M)`.<br><br>Example for squeeze over dimension -1 (innermost): `ttl.math.squeeze(a, dim=-1)`. Here if the shape of `a` is `(N, M, 1)` the shape of the result will be `(N, M)`. |
+| `ttl.math.unsqueeze(expr: ttl.BlockExpr, dim: int) -> ttl.BlockExpr` | Add shape dimension of 1 at position specified by `dim`. For positive `dim` the dimension is inserted before the position specified in `dim`. For negative `dim` the dimension is inserted after the position specified in `dim`. <br><br>Example for unsqueeze over dimension 0 (outermost): `ttl.math.unsqueeze(a, dim=0)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(1, N, M)`.<br><br>Example for unsqueeze over dimension -1 (innermost): `ttl.math.unsqueeze(a, dim=-1)`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(N, M, 1)`. |
```
brnorris03: what about multiple dims (e.g., specifying a list of consecutive dims)?

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: Unfortunately `<br><br>` with actual new lines ends up breaking markdown tables. So I went be to just `<br><br>`.

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: Good point.

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: They point to the end of `with` scope. Added clarification.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1254,8 +1254,8 @@ def matmul_read():
...
+| `ttl.math.squeeze(expr: ttl.BlockExpr, dims: List[int]) -> ttl.BlockExpr` | Remove shape dimension at positions specified by `dims`. Removed shape dimension must be 1.<br><br>Example for squeeze over dimensions 0 (outermost) and 2: `ttl.math.squeeze(a, dims=[0, 2])`. Here if the shape of `a` is `(1, N, 1, M)` the shape of the result will be `(N, M)`.<br><br>Example for squeeze over dimensions -1 (innermost) and -3: `ttl.math.squeeze(a, dim=[-1, -3])`. Here if the shape of `a` is `(N, 1, M, 1)` the shape of the result will be `(N, M)`. |
+| `ttl.math.unsqueeze(expr: ttl.BlockExpr, dims: List[int]) -> ttl.BlockExpr` | Add shape dimension of 1 at positions specified by `dims`. Position values in `dims` refer to a position in the resulting shape. <br><br>Example for unsqueeze over dimensions 0 (outermost) and 2: `ttl.math.unsqueeze(a, dims=[0, 2])`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(1, N, 1, M)`.<br><br>Example for unsqueeze over dimensions -1 (innermost) and -3: `ttl.math.unsqueeze(a, dims=[-1, -3])`. Here if the shape of `a` is `(N, M)` the shape of the result will be `(N, 1, M, 1)`. |
```
FConstantinos, line 1305: I think here, it's worth mentioning that positive index values in dims means the position before the position indicated by the index on the original block and negative index values in dims refer to a position after the position indicated by the index on the original block.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,342 @@ def some_compute():
...
 
+                        # Wait for a_blk and b_blk to be loaded and pushed by matmul_read
                         with (
                             a_dfb.wait() as a_blk,
                             b_dfb.wait() as b_blk,
                         ):
+                            # b_blk has shape K_BLOCK_SIZE×N_BLOCK_SIZE;
+                            # Unsqueeze it to 1×K_BLOCK_SIZE×N_BLOCK_SIZE and then
+                            # broadcast it over dim 0 to I_BLOCK_SIZE×K_BLOCK_SIZE×N_BLOCK_SIZE
+                            b = ttl.math.broadcast(ttl.math.unsqueeze(b_blk, dims=[0]), dims=[0], shape=(I_BLOCK_SIZE, K_BLOCK_SIZE, N_BLOCK_SIZE))
```
zoecarver: nit: shadows input variable name

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -200,213 +228,342 @@ def some_compute():
...
-    for nt in range(NT):
 
-        # acquire a_blk and b_blk from a_dfb and b_dfb:
+    # Reserve c_blk and d_blk blocks
+    with (
+        c_dfb.reserve() as c_blk,
+        d_dfb.reserve() as d_blk,
+    ):
+        # Load entire (1×M_TILES) of c
+        c_xf = ttl.copy(c[0, :], c_blk)
```
zoecarver, line 315: I think you call this out elsewhere but maybe good to expand the comment to explain that torch 1D is mapped to padded 2D in ttnn.

zoecarver: As a concrete example of 2d indexing in a 1d torch tensor

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: > "With double buffered dataflow buffer" -- we use different terminology in the actual methods (block count), so may be good to use/introduce that here, too?

We have this sentence preceding: "For the most common case block count defaults to 2 to support double buffering."

Is this sufficient to relate double buffering and block count?

zoecarver:
Do we need both `dims` and `output` shape parameters to be able to express both block-based broadcast/reduce and elementwise (inside tile) broadcast/reduce? 

The unsqueeze-broadcast-matmul example is making me think we should maybe spend more timing thinking about if the compiler can infer some of this (though I think we landed on "no") last time. 

Anyway, I like the added clarity, I think it makes what was implicit now explicit and easier to understand what exactly is happening, which is priority 1. LGTM.

zoecarver (approved):

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: Will provide another code example for row-major.

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: I replaced it with:

"Position values in `dims` refer to a position in the resulting shape."

Which I think clarifies this more succinctly and sufficiently. E.g. if resulting shape has number original dims + length of `dims`, positions in `dims` unambiguously apply to the resulting shape.

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: Please take a look.

docs/sphinx/specs/TTLangSpecification.md
phizalev-TT: Added clarification.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -179,13 +179,38 @@ shape_in_tiles(from_torch(torch.randn((128, 1)))) #        prints [4, 1]
...
+row_major_shape(from_torch(torch.randn((2, 128, 32)))) #    prints [2, 128, 32]
+row_major_shape(from_torch(torch.randn((2, 2, 128, 32)))) # prints [2, 2, 128, 32]
+row_major_shape(from_torch(torch.randn((2, 2, 120, 30)))) # prints [2, 2, 120, 30]
+```
+
+Shape determines the shape of a *block* returned by one of the *acquisition functions*: `wait` and `reserve`. The size of a block in L1 memory is determined by shape, shape unit and data type. For example, for a block with shape `(2, 2, 4, 1)`, shape unit of a tile (32 by 32 scalar elements) and BF16 data type (2 bytes), its size in L1 will be `2 * 2 * (4 * 32) * (1 * 32) * 2 = 32768` bytes. The block count determines the total size of L1 memory allocated for a dataflow buffer. This size as a product of a block size and block count. For the most common case block count defaults to 2 to support double buffering. With double buffered dataflow buffer one kernel can write to a block while another is reading from a block thus enabling the pipelining. For the example above, this means there will be a total of 32768 bytes of L1 memory allocated for the dataflow buffer.
```
brnorris03: Nit: "This size as a product of a block size and block count." -> "This size is a product of a block size and block count."

brnorris03 (approved):
lgtm, thank you!

PR #440, opened 2026-03-31
Remove acc argument from store

FConstantinos:
Remove acc argument from store and add '=' and '+=' operations for blocks


examples/matmul.py
```diff
@@ -106,13 +102,18 @@ def matmul_compute():
...
                             ):
-                                # then compute y += a @ b:
 
-                                y_blk.store(a_blk @ b_blk, acc=True)
+                                y += a_blk @ b_blk
 
                                 # release a_blk and b_blk
 
-                        # release y_blk
+                        with c_dfb.wait() as c_blk:
```
arichinsTT: I think this needs to be before the K loop or change the reader kernel for reading the bias after the K loop to match.  I don't think there is a deadlock, but it is confusing/inviting more errors

PR #446, opened 2026-04-01
[ttl] Add block expressions (lazy IR creation)

brnorris03:


brnorris03:
lower priority, closing for now -- will reopen when ready to resume

PR #450, opened 2026-04-01
[ttl] Add eq, ne, gt and lt ops

vtsilytskyiTT:
### How?
By using corresponding ttnn ops.
Test method changed. ULP and PCC are not applicable to bool output. 1 ≠ 0.
So a new boolean comparison function introduced.
FP32 support is done via proper SFPU comparison.
BF16 in the other hand utilizes sub+gtz approach. Pros: faster, cons: can't deal with NaN and other special values properly.

### Dependency

1. https://github.com/tenstorrent/tt-metal/pull/41304
2. https://github.com/tenstorrent/tt-mlir/pull/8174

### How to Test?

```bash
pytest test/python/test_elementwise_ops.py test/python/test_elementwise_shapes.py test/python/test_nd_shapes.py
pytest test/me2e/test_compare_ops.py test/me2e/ops/test_binary.py
```

### Checklist:
*   [X] Self-reviewed (style, logic)
*   [X] Added tests (or justified none needed)
*   [X] PR is small and focused (one task)


zoecarver (approved):
LGTM!

vtsilytskyiTT:
PR https://github.com/tenstorrent/tt-metal/pull/41304 merged, now tt-metal has a great support for comparison OPs.

vtsilytskyiTT:
PR https://github.com/tenstorrent/tt-mlir/pull/8174/changes also merged, all pieces of the puzzle are in place. After tt-mlir uplift I'll continue this PR.

test/me2e/test_compare_ops.py
```diff
@@ -0,0 +1,97 @@
...
+        skip_f32_1x1_compare = pytest.mark.skipif(
+            # TODO: investigate and fix this, float32 should work
+            dtype == torch.float32
+            and cfg.block_h == 1
+            and cfg.block_w == 1
+            and op.name in ("lt", "eq", "ne"),
+            reason=(
+                "Compare ME2E: skip float32 1x1 lt/eq/ne "
+                "(exact mask vs torch golden)."
+            ),
```
brnorris03: I would prefer these cases to xfail (unless xfail can't work because of parser issues, but if it's just correctness, then it should be fine), not be skipped, and include an issue number in the xfail message (for tracking, and eventual fixing) -- similar to other xfails we have now.

test/me2e/test_compare_ops.py
```diff
@@ -0,0 +1,97 @@
...
+the matching torch comparison ops; bool tensors are converted to float 0/1 for
+device I/O.
+Validation requires binary-identical masks (torch.equal), not PCC/ULP.
+
+Parametrization applies the test dtype to TestConfig before stringifying, so ids
+look like ``float32-2x2_float32_buf2_interleaved-gt`` (tensor dtype is uniform;
+there is no separate lhs/rhs dtype). Older patterns such as
+``float32-2x2_bfloat16_...`` came from ``str(config)`` using the template default
+dtype while the first segment was the parametrized dtype — misleading, not mixed
+precision.
```
brnorris03, line 18: Maybe should have a different naming implemented for this particular test? I find this comment a bit confusing, usually I would just include a brief summary of what is tested in the top comment.

test/me2e/ops/__init__.py
brnorris03: For the bool ops, perhaps should consider a set constant inputs for which we know the result to ensure we don't always get false or always true with the random values (missing some code paths)

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -955,6 +955,12 @@ defm TTL_Max : TTL_BinarySfpuElementwisePair<"max", "max_tiles",
 defm TTL_Min : TTL_BinarySfpuElementwisePair<"min", "binary_min_tile",
                                              [TTL_InPlaceOpTrait]>;
 
+// Elementwise compare: maps to TTKernel's native SFPU comparison ops.
+defm TTL_Eq : TTL_BinarySfpuElementwisePair<"eq", "eq_binary_tile">;
+defm TTL_Ne : TTL_BinarySfpuElementwisePair<"ne", "ne_binary_tile">;
+defm TTL_Gt : TTL_BinarySfpuElementwisePair<"gt", "gt_binary_tile">;
+defm TTL_Lt : TTL_BinarySfpuElementwisePair<"lt", "lt_binary_tile">;
```
brnorris03, line 967: If you want, consider adding to the comment a bit more on the result semantics here -- same element type as input, numeric 0/1 mask, not a bool tensor.

brnorris03 (commented):
A few inlined suggestions. The biggest issue I see is that the various tests have inputs that don't fully exercise the compare semantics. `test_binary_op` uses constant `lhs=2.0`, `rhs=3.0` (BF16). So gt/eq produce an all-zeros mask and lt/ne an all-ones mask — each test checks a single uniform value, never a 0/1 mix within a tile, and the equality boundary (lhs == rhs) -- having a mix would be better.

Also BF16 compare runs on-device in e2e but is explicitly skipped as unreliable in me2e -- does it pass on e2e because it has the const inputs? I think that it should fail similarly in both cases if the same types of inputs are used; and this definitely should be tracked in a bug issue.

I would add an e2e pytest with a kernel computing a == b where a and b are tile-filled with the same value should yield an all-ones mask. Given gt passes but eq/ne/lt are skipped on 1×1 f32, confirm:
```
  lhs = torch.full((32,32), 1.5, dtype=torch.float32)
  rhs = torch.full((32,32), 1.5, dtype=torch.float32)
  # eq -> expect all 1.0 ; ne -> all 0.0 ; lt -> all 0.0 ; gt -> all 0.0
```
If you can't identify/fix the problem here, I am fine with xfailing this with an issue pointer. Do you have an idea about what may be causing this?

test/me2e/test_compare_ops.py
vtsilytskyiTT: These tests are working already. Forgot to remove this code snippet.

test/me2e/ops/__init__.py
vtsilytskyiTT: I'll make a function `make_compare_inputs` for reproducible input generation.

zoecarver (approved):

brnorris03 (approved):
Looks great! Glad it was just forgetting to remove the skipped tests and not a real problem, phew. :-)

phizalev-TT (approved):

PR #453, opened 2026-04-02
Add `ttl.math.fill` op (with output cb)

zoecarver:


brnorris03 (commented):
Looks good in general, only structural issue I have is with emitting the init outside the dedicated  `TTKernelInsertInits` pass. 

Can NaN or Inf, etc be used with fill? Not sure what happens to those when we do `convertToFloat()` in the TTKernel lowering.

Optional, but can possibly add a couple more tests:

- no input cb
```python
  @ttl.operation(grid=(1, 1))
  def fill_only_kernel(out):
      out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), buffer_factor=2)
      @ttl.compute()
      def compute_fn():
          with out_dfb.reserve() as o:
              o.store(ttl.math.fill(o, 42.0))
      @ttl.datamovement()
      def dm_write():
          with out_dfb.wait() as blk:
              tx = ttl.copy(blk, out[0, 0])
              tx.wait()
```

- fused with elementwise
```python
  @ttl.compute()
  def compute_fn():
      with inp_dfb.wait() as inp_blk, out_dfb.reserve() as o:
          filled = ttl.math.fill(o, 1.0)
          result = ttl.math.add(inp_blk, filled)
          o.store(result)
```

- maybe negative test for fill without a store? Not sure it's possible to construct, though

PR #457, opened 2026-04-02
Simulator: Add support for sharding in ttnnsim

FConstantinos:
This just adds the necessary front end for defining Tensors that are sharding. It has no functional effect


zoecarver (approved):
LGTM. Do we need a stub for the ttnn apis too?

FConstantinos:
> LGTM. Do we need a stub for the ttnn apis too?

Added some of the ttnn APIs around sharding.

PR #461, opened 2026-04-03
[fix][ttl] Fix multi-tile matmul+add fusion in ConvertTTLToCompute

### LINKED ISSUE #460: [bug] ConvertTTLToCompute fails to fuse multi-tile matmul+add (prev + a @ b)
### Problem description

`ConvertTTLToCompute` fails to fuse `ttl.add(prev, ttl.matmul(a, b))` into a 3-operand `tile_matmul_block` when the matmul operands have multi-tile shapes (e.g., `[4,2] @ [2,4] -> [4,4]`). The matmul and add survive the pass unconverted, and the pipeline eventually fails at `ConvertTTLToTTKernel` with:

```
error: unknown: failed to legalize operation 'ttl.matmul' that was explicitly marked illegal
```

The same pattern works correctly with single-tile shapes (`[1,1] @ [1,1] -> [1,1]`).

### Minimal reproducer

```mlir
// RUN: ttlang-opt --convert-ttl-to-compute %s
// Multi-tile fused matmul+add: ttl.matmul and ttl.add survive unconverted.
module {
  func.func @compute() attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
    %a_cb = ttl.bind_cb{cb_index = 0, buffer_factor = 2} : <[4, 2], !ttcore.tile<32x32, bf16>, 2>
    %b_cb = ttl.bind_cb{cb_index = 1, buffer_factor = 2} : <[2, 4], !ttcore.tile<32x32, bf16>, 2>
    %acc_cb = ttl.bind_cb{cb_index = 2, buffer_factor = 2} : <[4, 4], !ttcore.tile<32x32, bf16>, 2>
    %c3 = arith.constant 3 : index
    %c1 = arith.constant 1 : index
    scf.for %i = %c0 to %c3 step %c1 {
      %a_t = ttl.cb_wait %a_cb : <[4, 2], !ttcore.tile<32x32, bf16>, 2> -> tensor<4x2x!ttcore.tile<32x32, bf16>>
      %a = ttl.attach_cb %a_t, %a_cb : (tensor<4x2x!ttcore.tile<32x32, bf16>>, !ttl.cb<[4, 2], !ttcore.tile<32x32, bf16>, 2>) -> tensor<4x2x!ttcore.tile<32x32, bf16>>
      %b_t = ttl.cb_wait %b_cb : <[2, 4], !ttcore.tile<32x32, bf16>, 2> -> tensor<2x4x!ttcore.tile<32x32, bf16>>
      %b = ttl.attach_cb %b_t, %b_cb : (tensor<2x4x!ttcore.tile<32x32, bf16>>, !ttl.cb<[2, 4], !ttcore.tile<32x32, bf16>, 2>) -> tensor<2x4x!ttcore.tile<32x32, bf16>>
      %prev_t = ttl.cb_wait %acc_cb : <[4, 4], !ttcore.tile<32x32, bf16>, 2> -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      %prev = ttl.attach_cb %prev_t, %acc_cb : (tensor<4x4x!ttcore.tile<32x32, bf16>>, !ttl.cb<[4, 4], !ttcore.tile<32x32, bf16>, 2>) -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      %reserve = ttl.cb_reserve %acc_cb : <[4, 4], !ttcore.tile<32x32, bf16>, 2> -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      %out = ttl.attach_cb %reserve, %acc_cb : (tensor<4x4x!ttcore.tile<32x32, bf16>>, !ttl.cb<[4, 4], !ttcore.tile<32x32, bf16>, 2>) -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      %mm = ttl.matmul %a, %b : tensor<4x2x!ttcore.tile<32x32, bf16>>, tensor<2x4x!ttcore.tile<32x32, bf16>> -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      %sum = ttl.add %prev, %mm : tensor<4x4x!ttcore.tile<32x32, bf16>>, tensor<4x4x!ttcore.tile<32x32, bf16>> -> tensor<4x4x!ttcore.tile<32x32, bf16>>
      ttl.store %sum, %reserve : tensor<4x4x!ttcore.tile<32x32, bf16>>, tensor<4x4x!ttcore.tile<32x32, bf16>>
      ttl.cb_push %acc_cb : <[4, 4], !ttcore.tile<32x32, bf16>, 2>
      ttl.cb_pop %acc_cb : <[4, 4], !ttcore.tile<32x32, bf16>, 2>
      ttl.cb_pop %b_cb : <[2, 4], !ttcore.tile<32x32, bf16>, 2>
      ttl.cb_pop %a_cb : <[4, 2], !ttcore.tile<32x32, bf16>, 2>
    }
    return
  }
}
```

### Root cause

Three points in `ConvertTTLToCompute.cpp` reject multi-tile matmul+add fusion:

1. **Matmul defer-to-fusion check** (`LowerMatmulToCompute`, line 1068-1077): When a matmul's sole user is an elementwise op (the add), the pattern attempts to defer lowering so that the elementwise pattern can fuse both ops. The defer is conditional on the matmul inputs being broadcast-compatible with the user's output type. For `[4,2]` vs `[4,4]`, the K dimension (2 vs 4) is neither equal nor 1, so the check fails and the matmul is lowered standalone instead of deferring.

2. **Fusion broadcast validation** (`buildFusedCompute`, line 309-319): Even if the matmul deferred, `traceFusionToRoots` collects the matmul's `[4,2]` and `[2,4]` inputs as root inputs. The per-input broadcast compatibility check against the `[4,4]` output rejects them for the same reason.

3. **Iteration space and indexing maps** (`buildFusedCompute`, line 326-358): Even if both checks were bypassed, `buildFusedCompute` constructs a 2D all-parallel iteration space with identity/broadcast maps. A fused matmul+add requires a 3D iteration space with the K dimension marked as `reduction` and matmul-specific affine maps: `lhs->(d0,d2)`, `rhs->(d2,d1)`, `acc->(d0,d1)`, `out->(d0,d1)`.

For single-tile blocks (`[1,1]`), all three points pass trivially: `isBroadcastCompatible([1,1], [1,1])` is always true, and the 2D compute with identity maps works because K=1 makes the reduction degenerate.

### Expected behavior

`ConvertTTLToCompute` should fuse `prev + a @ b` with multi-tile blocks into a `ttl.compute` with a 3-operand `tile_matmul_block`:

```
ttl.compute ins(%prev, %a, %b) outs(%out)
  {maps = [acc->(d0,d1), lhs->(d0,d2), rhs->(d2,d1), out->(d0,d1)],
   iters = [parallel, parallel, reduction]} {
  tile_matmul_block(%lhs, %rhs, %acc)
}
```

This is the same structure `LowerMatmulToCompute` builds for standalone matmul, extended with an accumulator input.

brnorris03:
### Problem description
Fixes #460. `prev + a @ b` and longer chains like `relu(prev + a @ b)` only fused into `tile_matmul_block` for single-tile blocks. Multi-tile shapes (e.g., `[4,2] @ [2,4] -> [4,4]`) failed because `buildFusedCompute` assumed all inputs are broadcast-compatible with the output. For matmul, the K (contraction) dimension differs from the output shape, so the broadcast check rejected it and the 2D iteration space could not express the K reduction.

### What's changed
~61 lines of new implementation (most of it in `buildFusedCompute`)

`buildFusedCompute` now analyzes the fusion chain before building the `ttl.compute` to determine the correct iteration space. When the chain contains a matmul with a non-trivial K dimension (multi-tile), the iteration space is 3D `[M, N, K]` with K marked as `reduction` instead of the 2D `[M, N]` all-parallel space used for pure elementwise chains. This is necessary because:
- The matmul's LHS needs indexing `(d0, d2)` (M, K) and RHS needs `(d2, d1)` (K, N), which cannot be expressed in a 2D iteration space.
- Non-matmul inputs in the chain (accumulator, broadcast bias) use `(d0, d1)` maps that project only M and N from the 3D space.
- The deferred-matmul fold in the compute body emits `tile_matmul_block` which handles the per-tile K reduction via hardware accumulation in DST.

For single-tile blocks (`[1,1]`), K=1 makes the 2D identity maps trivially correct, so no promotion is needed -- the existing mechanism continues to work unchanged.

`LowerMatmulToCompute` unconditionally defers to `buildFusedCompute` when the matmul feeds an elementwise op. Standalone matmul (result stored directly) is still lowered by the matmul pattern with its own 3D iteration space.

Multiple multi-tile matmuls in a single fusion chain are rejected with a clear diagnostic (not supported at the moment).

### Testing
- MLIR positive: `convert_ttl_matmul_add_multitile.mlir` (fused multi-tile, commuted operands, standalone matmul)
- MLIR negative: `convert_ttl_matmul_add_multitile_invalid.mlir` (multiple multi-tile matmuls rejected with diagnostic)
- MLIR updated: `matmul_fusion.mlir` CHECKs updated for single-tile cases
- E2e pytest: `test_matmul_multitile_acc.py`:
  - matmul+bias with K-loop accumulation (block sizes 1x1x1, 2x1x2, 2x2x2, 4x2x4)
  - matmul+bias with full K in one DFB block, no K-loop (block sizes 1x1, 2x2, 4x4)
  - multi-tile relu after matmul+add (block sizes 1x1x1, 2x2x2, 4x2x4)

### DIFF
--- a/python/sim/__init__.py
+++ b/python/sim/__init__.py
@@ -124,7 +124,7 @@ __all__ = [
     "node",
     "compute",
     "datamovement",
-    "kernel",
+    "operation",
     "ttl",
     "ttnn",
     "TTNN_AVAILABLE",
--- a/python/sim/dfb.py
+++ b/python/sim/dfb.py
@@ -40,7 +40,13 @@ from .dfbstate import DFBState
 from .constants import TILE_SHAPE
 from .errors import DFBContractError
 from .stats import record_dfb_reserve, record_dfb_wait
-from .ttnnsim import ROW_MAJOR_LAYOUT, TILE_LAYOUT, Tensor, tile_count_from_tensor
+from .ttnnsim import (
+    ROW_MAJOR_LAYOUT,
+    TILE_LAYOUT,
+    Tensor,
+    tile_count_from_tensor,
+    tile_shape_from_tensor,
+)
 from .typedefs import Index, IndexType, PositiveInt, Shape, Size
 
 
@@ -498,8 +504,6 @@ class Block:
                     f"1-D tensor dimension ({w},) must be a multiple of "
                     f"TILE_SHAPE[0]={TILE_SHAPE[0]}, or exactly 1"
                 )
-            tk = 1 if w == 1 else w // TILE_SHAPE[0]
-            tile_shape: Shape = (tk,)
         else:
             h, w = elem_shape[-2], elem_shape[-1]
             if (h != 1 and h % TILE_SHAPE[0] != 0) or (
@@ -509,10 +513,7 @@ class Block:
                     f"Last two tensor dimensions ({h}, {w}) must be multiples of "
                     f"TILE_SHAPE {TILE_SHAPE}, or exactly 1"
                 )
-            batch_shape = elem_shape[:-2]
-            tm = 1 if h == 1 else h // TILE_SHAPE[0]
-            tk = 1 if w == 1 else w // TILE_SHAPE[1]
-            tile_shape = (*batch_shape, tm, tk)
+        tile_shape: Shape = tile_shape_from_tensor(t)
 
         return cls(
             tensor=t,
--- a/python/sim/ttnnsim.py
+++ b/python/sim/ttnnsim.py
@@ -18,6 +18,9 @@ Scope:
 
 from __future__ import annotations
 
+import math
+from dataclasses import dataclass
+from enum import Enum, auto
 from typing import Any, Callable, List, Optional, Tuple, Union, cast
 
 import torch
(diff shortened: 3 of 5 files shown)


zoecarver (approved):
I don't love this approach, and I think we could flip fusion to make it simpler. But I'm OK with this as long as we revisit after the release. 

This fusion logic started as:
1. only eltwise ops that do no operate on cbs can be fused together (trivial) 
2. the start of a chain can operate on CBs <-- where we are now
3. the start of the chain can be an add + matmul 

But we actually look at it in the opposite way, I think. Because we start with the last store, and sink everything into that. So we kind of go backwards, which is why I think there's a lot of complexity here. Also the 3D iteration space for matmul is adding complexity. 

If you wanted, I think we could start from the first op, and apply those rules sinking from top to bottom, pulling ops into the compute region above (instead of below). Up to you.

Issue #420, opened 2026-03-24
[ttkernel] Add non-experimental matmul_block op to tt-mlir


brnorris03:
## Problem

The current matmul lowering depends on `ttkernel.experimental_matmul_block`, which is marked experimental in the TTKernel dialect. This creates a dependency on an unstable API.

## Proposed solution

Add a stable `ttkernel.matmul_block` op to tt-mlir's TTKernel dialect that:
- Wraps the tt-metal `matmul_block` API (non-experimental)
- Supports the 3-operand accumulator form (lhs, rhs, optional accumulator)
- Has proper ODS definition with verifier

This would replace the current `experimental::matmul_block` emission in `TTLLowerMatmulBlock.cpp` and `ConvertTTLTileOpsToTTKernel.cpp`.

## Context

Related to #123 (ttl.tile_matmul).

brnorris03:
[...]

PR #464, opened 2026-04-03
[fix][ttl] Fix incorrect indexing maps bug after fusing matmul with add and add more robust fusion compatibility checks

brnorris03:
### Bug

When `acc + a @ b` is fused into a single `ttl.compute`, the elementwise fusion in `buildFusedCompute` (incorrectly) assigned 2D identity indexing maps `(d0,d1)->(d0,d1)` to all inputs, including the matmul's LHS and RHS. The K reduction dimension was absent from the iteration space. When DST subblocking partitioned the compute along M, B (which is `[K,N]`) was sliced identically to A (along M), producing wrong tile indices in the generated `matmul_block` calls. Result: garbage output (PCC ~0.25) for any configuration where M subblocking occurs (e.g., 4x4x4 blocks with f32 DST accumulation).

### Changes

Generalize iteration dimension consistency check to cover both matmul and non-matmul inputs uniformly. In `ConvertTTLToCompute.cpp`, we pre-scan the fusion chain for `MatmulOp`. When present, build a 3D `[M, N, K]` iteration space with correct matmul indexing maps: LHS `(d0,d1,d2)->(d0,d2)`, RHS `(d0,d1,d2)->(d2,d1)`, all other inputs/outputs `(d0,d1,d2)->(d0,d1)`, K marked as `reduction`. Input shapes are validated against their assigned indexing maps: each iteration dimension must have a consistent size across all inputs that reference it. Structured for future transpose support (TODO #420).

Other cleanup: 
- In `ConvertTTLTileOpsToTTKernel.cpp`, we remove the `kt == 1` special case. The K loop is now always emitted as `scf.for`. Replaced `arith.addi`/`arith.muli` index arithmetic with `affine.apply` maps for A and B tile index computation.

### Testing

- `matmul_fusion.mlir` — updated CHECK patterns for 3D maps; added non-square regression test (A=[2,4], B=[4,3], acc=[2,3]) that would fail with the old 2D identity maps
- e2e multinode fused matmul tests: `test/python/test_matmul_multinode_fused.py`

### DIFF
--- a/lib/Dialect/TTL/IR/TTLOps.cpp
+++ b/lib/Dialect/TTL/IR/TTLOps.cpp
@@ -105,10 +105,10 @@ mlir::LogicalResult mlir::tt::ttl::BindCBOp::verify() {
   // Validate block count against type for consistency.
   int64_t blockCount = getBlockCount();
   if (blockCount <= 0) {
-    return emitOpError() << "buffer_factor must be > 0";
+    return emitOpError() << "block_count must be > 0";
   }
   if (blockCount != cbTy.getBlockCount()) {
-    return emitOpError() << "buffer_factor must match result type block count ("
+    return emitOpError() << "block_count must match result type block count ("
                          << cbTy.getBlockCount() << ")";
   }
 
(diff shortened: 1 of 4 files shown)


arichinsTT (approved):
LGTM

PR #488, opened 2026-04-09
Simulator: Introduce bfloat8_b for ttnnsim

FConstantinos:


python/sim/ttnnsim.py
```diff
@@ -295,6 +295,51 @@ def __repr__(self) -> str:
...
+        """Mantissa storage width in bytes (excludes shared exponent overhead)."""
+        return 1
+
+    def size_in_bytes(self, n_elements: int) -> int:
+        """Total bytes required to store n_elements in BFP8B encoding.
+
+        Accounts for both the per-element mantissa byte and the shared
+        exponent byte for every group of _EXPONENT_GROUP_SIZE elements.
+        """
+        return n_elements + n_elements // self._EXPONENT_GROUP_SIZE
```
zoecarver: Probably not a big issue but we'll undercount bytes that are not a multiple of 16.

zoecarver (approved):

phizalev-TT (approved):

python/sim/ttnnsim.py

PR #493, opened 2026-04-11
Simulator: Add sharding computations

FConstantinos:
This adds the necessary computations for deriving where each shard is located.

It also adds to the sharding interface. The examples described here for ttnn should work:

https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/tensor_sharding


python/sim/ttnnsim.py
```diff
@@ -931,27 +1515,21 @@ def create_sharded_memory_config(
         orientation if orientation is not None else ShardOrientation.ROW_MAJOR
     )
 
-    def _to_tile(n: int, tile_dim: int) -> int:
```
zoecarver, line 934: Is this elementwise now? Is that change intentional?

python/sim/ttnnsim.py
```diff
@@ -89,56 +102,284 @@ class ShardDistributionStrategy(Enum):
...
+    set ``distribution`` to :data:`ShardDistributionStrategy.ROUND_ROBIN_1D`.
+    ``num_cores`` applies only to ROUND_ROBIN (modulus for shard assignment).
     """
 
-    shard_grid: Shape
     shard_shape: Shape
-    distribution: ShardDistributionStrategy = ShardDistributionStrategy.ROUND_ROBIN_1D
+    core_ranges: Optional["CoreRangeSet"] = None
+    shard_grid: Optional[Shape] = None
+    distribution: ShardDistributionStrategy = ShardDistributionStrategy.GRID_2D
```
zoecarver: Any reason for the change from ROUND_ROBIN_1D?

zoecarver (approved):

python/sim/ttnnsim.py
FConstantinos: yes, it is intentional. Going through this document:

https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/tensor_sharding/tensor_sharding.md

and searching for ShardSpec, it appears to me that the shape of the shard is always given in elements, not tiles. So I'd prefer if we are closer to that definition.

python/sim/ttnnsim.py
FConstantinos: Thanks, good catch! No good reason really. I think you are right that the default is ROUND_ROBIN_1D, according to this:

https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn-nanobind/tensor.cpp#L583C1-L607C1

Changed it back

PR #494, opened 2026-04-13
Simulator: Add all_reduce to ttnnsim

FConstantinos:
Add all_reduce to ttnnsim and tests all matmul-tutorial examples at a smaller scale

The matmul-tutorial examples are always executed in CI but are otherwise off by default


test/sim/test_examples.py
```diff
@@ -376,3 +378,55 @@ def test_eltwise_1d_broadcast_warning(scheduler: str) -> None:
     assert (
         "examples/eltwise_1d_broadcast.py:" in out
     ), f"Expected source location not found in output:\n{out}"
+
+
+# ---- Matmul tutorial -------------------------------------------------------
+
+_MATMUL_TUTORIAL_LARGE_DIM = "M, K, N = 8192, 8192, 8192"
+_MATMUL_TUTORIAL_TEST_DIM = "M, K, N = 1024, 1024, 1024"
```
phizalev-TT: With 4x4 blocks in matmul and test on 8x8 grid (auto) we will have only 1x1 blocks per node (core) and even less if we simulate multiple devices.

phizalev-TT (changes requested):

python/sim/ttnnsim.py
```diff
@@ -1772,6 +1838,81 @@ def split_work_to_cores(
...
+        mc.strategy
+        in (
+            ShardingStrategy.HEIGHT_SHARDED,
+            ShardingStrategy.WIDTH_SHARDED,
+        )
+        and mc.shard_spec is not None
+    ):
+        n = mc.shard_spec.shard_grid[0]
+        if mc.strategy == ShardingStrategy.HEIGHT_SHARDED:
+            shard = mc.shard_spec.shard_shape[-2]
```
zoecarver: I think this will break for >2D. Do we want to support higher rank tensors or does the ttnn api only support 2d as well?

FConstantinos: Thanks, fixed!

test/sim/test_examples.py
FConstantinos: In this case, I think 8192 is the only viable dimension.

To address that, I resorted to running all the tutorial tests with the original dimension (8192) on the Hardware CI only *except* step_1!

Here is the reasoning:

Step 1 takes much longer than the others. The reason is because the dfbs are a single tile for each element and therefore we get greenlet switching very often. Indeed the fair scheduler is much slower than the greedy one, which is much slower than the other steps.

By removing that test, it takes an order of magnitude less time to run the tutorial tests. The other tutorial tests except step_1 take about 1.5 minutes to run on our qb as they are. So it's not all that bad to run them with the original size.

Let's see how long they take on the Hardware CI machine though!

phizalev-TT (approved):

PR #495, opened 2026-04-13
[nfc] Make tt-lang installable with "pip install"

brnorris03:
Add pip install support and PyPI wheel publishing.

Reorganize top-level Python sources under the `ttl` package and add packaging configuration so tt-lang can be installed via pip.

- Move top-level Python modules into `python/ttl/` and update all imports across examples and tests
- Update `pyproject.toml`, `python/setup.py`, and `python/CMakeLists.txt` to build distributable wheels (forwards compiler settings from CMake to `setup.py`)
- Add `call-build-wheels.yml` reusable workflow that builds and audits wheels inside the `tt-lang-ird-ubuntu-22-04` container, plus `publish-pypi.yml` for manual-dispatch PyPI publishing
- Misc CI cleanup: silence deprecated-action warnings, revert ccache-action version


zoecarver (approved):
lgtm! I know folks are excited for this.

test/sim/test_math.py
```diff
@@ -10,7 +10,7 @@
 import pytest
 import torch
 
-from sim import ttl
+from sim import ttl as sim_ttl
```
FConstantinos: Why is that change necessary here?

brnorris03: To avoid conflicting with the top-level `ttl`.

pyproject.toml
```diff
@@ -3,12 +3,12 @@ requires = ["setuptools>=61.0", "cmake", "nanobind", "wheel", "pip", "ninja"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "ttl"
+name = "tt-lang"
 description = "Python Bindings and Package for TT-Lang Compiler Project"
 requires-python = ">=3.11"
```
FConstantinos, line 8: should this be >=3.12?

pyproject.toml
```diff
@@ -31,16 +30,24 @@ classifiers = [
...
 
 [project.urls]
 Homepage = "https://github.com/tenstorrent/tt-lang"
 Repository = "https://github.com/tenstorrent/tt-lang"
 Issues = "https://github.com/tenstorrent/tt-lang/issues"
 
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements-runtime.txt"]}
+
 [tool.cibuildwheel]
```
FConstantinos: Is this used at all? A git grep gives me this as the only reference:

```
(.venv) kfragkiadakis@Konstantinos-Fr's-Mac tt-lang8 % git grep -rn cibuildwheel
pyproject.toml:49:[tool.cibuildwheel]
```

test/sim/test_math.py
FConstantinos: OK, but other tests bellow it (e.g test_operation.py) don't seem to have this problem

FConstantinos:
Looks good! A couple of questions:

- Is it possible to only install the simulator?
- does this installation work on environments that don't have ttnn or tt hardware?

pyproject.toml
brnorris03: No, we don't require 3.12 specifically -- I just want to build few wheels to start with, not for all possible versions. Can add more wheels once the 3.12 one works.

pyproject.toml
brnorris03: yes, not used right now, removing!

test/sim/test_math.py
brnorris03: hm, I think maybe it was at one point but then I changed other things to not get that conflict. So yes, no need to change here, reverting.

brnorris03:
> Looks good! A couple of questions:
> 
> * Is it possible to only install the simulator?
> * does this installation work on environments that don't have ttnn or tt hardware?

You could probably create a simulator-only wheel in addition to the full one. Can add that in a subsequent PR. Things that don't require ttnn or hardware would work the same way as any local build without ttnn or hardware.

FConstantinos (approved):
LGTM, thank you!

PR #496, opened 2026-04-13
Initial pipes implementation

zoecarver:
This is still a draft, but the lowering logic should be sound. Main fixes to make before ready for review:
* clean up the tests, add tests, move them into a sub directory 
* clean up the API to make sure it matches the spec and clean up ttl api in general


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03, line 1: I think the pipe-specific structs and class below should be in a separate header and not directly inlined here.

Also up to you, but could also have the actual lowering except for modification of copy lowering and such be in its own cpp file (this one is getting huge, should probably split out more things at some point, but at least new sizable additions can be separate files going forward).

python/ttl/_src/ttl_ast.py
```diff
@@ -314,6 +319,107 @@ def visit_AugAssign(self, node):
...
+
+            # Create body block and compile callback inside
+            block = Block.create_at_start(op.body)
+            with InsertionPoint(block):
+                # Bind the pipe parameter to the pipe value
+                self.symbol_tables.append({})
+                self.symbol_tables[-1][pipe_param_name] = pipe_val
+                # Also store the identity object for property access
+                self.symbol_tables[-1][f"__{pipe_param_name}_identity"] = pipe_identity
+
```
brnorris03, line 417: The spec describes the callback argument as a pipe identity with read-only src and dst. Here the formal pipe param is bound to the MLIR pipe value (opaque to python), not to something like SrcPipeIdentity or DstPipeIdentity, so using `.src` or `.dst` on that parameter would not work. Maybe a TODO?

python/ttl/_src/ttl_ast.py
```diff
@@ -314,6 +319,107 @@ def visit_AugAssign(self, node):
...
+            return False
+        # Check if the variable is a PipeNet
+        var_name = node.func.value.id
+        tbl = self._var_exists(var_name)
+        if not tbl:
+            return False
+        val = tbl[var_name]
+        from ..pipe import PipeNet
+
+        return isinstance(val, PipeNet)
```
brnorris03, line 342: So this only works if the lhs of `.if_src` or `.if_dst` is a plain name (not expression), so if there is a pipenet callback call that has an expression as the lhs, this would return false which would lead to a possibly confusing other error later on.

python/ttl/pipe.py
```diff
@@ -0,0 +1,240 @@
...
+            elif isinstance(x, slice) and isinstance(y, slice):
+                # Multicast over x and y
+                self.dst_start = (x.start, y.start)
+                self.dst_end = (x.stop - 1, y.stop - 1)
+                self._is_multicast = True
+            else:
+                raise ValueError(f"Invalid dst format: {dst}")
+        else:
+            raise ValueError(f"dst must be a 2-tuple, got {dst}")
+
```
brnorris03, line 138: Should there be some validation here that self.dst_start self.dst_end is a valid range (and slices are valid, e.g.,, `slice(None,4)` is a valid python slice, but not for the dst context).

Also `dst` is kind of overloaded since it's the common abbreviation for DST register, so this can be slightly confusing.

python/ttl/pipe.py
```diff
@@ -0,0 +1,240 @@
...
+        All pipes in a PipeNet share a single semaphore pair. For multicast
+        pipes, the handshake protocol cannot handle a core receiving from
+        multiple sources. Use separate PipeNets for patterns where multicast
+        destinations overlap (e.g., scatter-gather/all-to-all).
+
+        Unicast gather (multiple unicast pipes to one destination) is allowed
+        because the receiver uses cumulative semaphore waits.
+        """
+        mcast_pipes = [(i, p) for i, p in enumerate(pipes) if p.is_multicast]
```
brnorris03, line 201: Given this current constraint, maybe document higher up in the user docs (PipeNet docstring?) that overlapping multicast destination cores in a single pipenet are not supported yet.

python/ttl/ttl_api.py
```diff
@@ -1313,8 +1330,20 @@ def _compile_kernel(
...
+        # Set up pipe graph JSON path for compiler to write and runtime to read.
+        # This enables pipe synchronization for gather patterns.
+        pipe_graph_fd = None
+        pipe_graph_path = None
+        original_env = os.environ.get("TTLANG_PIPE_GRAPH_JSON")
         try:
+            pipe_graph_fd = tempfile.NamedTemporaryFile(
+                mode="w", suffix=".json", delete=False
+            )
+            pipe_graph_path = pipe_graph_fd.name
```
brnorris03: never cleaned up? So could produce wrong results if loaded later for a different pipe grpah potentially

python/ttl/ttl_api.py
```diff
@@ -1330,6 +1359,10 @@ def _compile_kernel(
 
             formatted = f"ttlang {__version__}\n{format_mlir_error(error_msg, source_lines, source_file)}"
             raise RuntimeError(formatted) from None
+        finally:
+            # Restore original environment variable if it was set
+            if original_env is not None:
+                os.environ["TTLANG_PIPE_GRAPH_JSON"] = original_env
```
brnorris03: suggest using `os.environ.pop("TTLANG_PIPE_GRAPH_JSON", None)`

test/ttlang/Dialect/TTL/Transforms/convert_pipe_ops.mlir
```diff
@@ -0,0 +1,74 @@
...
+// CHECK-LABEL: func.func @copy_pipe_to_cb
+// CHECK: ttkernel.get_semaphore
+// CHECK: ttkernel.experimental::semaphore_wait_min
+// CHECK: ttkernel.noc_semaphore_set
+func.func @copy_pipe_to_cb() attributes { "ttl.kernel_thread" = #ttkernel.thread<noc> } {
+  %cb = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
+  %p = ttl.create_pipe src(0, 0) dst(1, 0) to(1, 0) net 0 : !ttl.pipe<src(0, 0) dst(1, 0) to(1, 0) net 0>
+  %xf = ttl.copy %p, %cb : (!ttl.pipe<src(0, 0) dst(1, 0) to(1, 0) net 0>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> !ttl.transfer_handle<read>
+  ttl.wait %xf : !ttl.transfer_handle<read>
+  func.return
```
brnorris03: Claude complaint:
`convert_pipe_ops.mlir`: The `copy_pipe_to_cb` test uses `!ttl.transfer_handle<read>` for the pipe-to-CB copy, but the Python API emits untyped `!ttl.transfer_handle` (no direction). With `<read>`, `WaitLowering` emits `noc_async_read_barrier` instead of erasing the wait op (the intended no-op behavior for pipe receive). Use untyped `!ttl.transfer_handle` to match actual generated IR.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -53,6 +57,267 @@ constexpr llvm::StringLiteral kBaseCTAIndexAttr = "ttl.base_cta_index";
...
+    auto it = receiverCBs.find(key);
+    if (it == receiverCBs.end() || it->second.empty()) {
+      return nullptr;
+    }
+    auto &counter = lookupCounters[key];
+    if (counter >= it->second.size()) {
+      return nullptr;
+    }
+    return &it->second[counter++];
+  }
```
brnorris03: `getReceiverInfo()` and `getGatherRecvProgress()` advance mutable counters on const methods, making the result depend on call order. If the greedy rewriter visits CopyOps in a different order than program order (which it can), the wrong ReceiverCBInfo gets returned for gather patterns. Consider keying lookups on the CopyOp itself or its program-order index rather than relying on sequential (global) counter advancement.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -731,12 +1410,44 @@ struct CopyLowering : OpConversionPattern<CopyOp> {
...
+            pipeType.getDstEndY(), pipeType.getPipeNetId());
+      }
+      // Determine CB access context: consumer (cb_wait/cb_pop) vs producer
+      // (cb_reserve/cb_push). This controls whether we read from the CB's
+      // read pointer or write pointer for the pipe source address.
+      bool isConsumerCB = llvm::any_of(
+          src.getUsers(), [](Operation *user) { return isa<CBWaitOp>(user); });
+      return lowerCBToPipe(op, adaptor.getSrc(), adaptor.getDst(), receiverInfo,
+                           isConsumerCB, rewriter);
+    }
```
brnorris03: This checks for any CBWaitOp user, not whether this copy is a consumer (i.e., dominated by a cb_wait). What if a CB is used in both producer and consumer roles in the same function? I'd use dominance info here instead of the user-based check.

python/ttl/kernel_runner.py
```diff
@@ -36,6 +38,69 @@ def _ensure_ttnn():
...
+                    dstEndY=p["dstEndY"],
+                    receiverCBIndex=p["receiverCBIndex"],
+                    runtimeArgSlot=p["runtimeArgSlot"],
+                    pipeNetId=p.get("pipeNetId", 0),
+                )
+            )
+        return pipes
+    except (json.JSONDecodeError, KeyError, TypeError):
+        return None
+
```
brnorris03: Silently failing if the JSON is incorrect may not be great (e.g., if there is a typo, the whole pipe graph will be dropped), maybe better to not catch and silence `json.JSONDecodeError`s?

python/ttl/_src/ttl_ast.py
```diff
@@ -296,6 +301,125 @@ def visit_Call(self, node):
...
+        else:
+            self._raise_error(
+                callback_node,
+                f"PipeNet.{method_name}() requires a lambda or function reference",
+            )
+
+        # Iterate over all pipes and emit if_src/if_dst for each
+        for pipe in pipenet.pipes:
+            # Emit the pipe MLIR value
+            pipe_val = self._emit_pipe_from_capture(pipe)
```
brnorris03, line 397: Optional: This creates a new `ttl.create_pipe` op every time (so a pipe used in both `if_src` and `if_dst` gets 2+ redundant ops). The code sets `pipe._mlir_value` but never checks it on subsequent calls. This is up to you, but could cache and reuse the value, e.g., in a compiler-local `Dict[int, MlirValue]` keyed by `id(pipe)`).

python/ttl/operators.py
```diff
@@ -294,21 +315,83 @@ def _process_tensor_subscript(subscript_tuple, cb_shape):
...
+def _get_pipe_mlir_value(pipe):
+    """Get the MLIR value for a pipe (either MLIR value or Python Pipe object)."""
+    # If it's already an MLIR value, return it
+    if hasattr(pipe, "type"):
+        type_str = str(pipe.type)
+        if "ttl.pipe" in type_str:
+            return pipe
+    # If it's a Python Pipe object, get its MLIR value
+    return pipe._mlir_value
+
```
brnorris03, line 331: Consider using `PipeType.maybe_downcast(val.type)` instead of using "ttl.pipe" in str(val.type) for type checking.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -53,6 +57,267 @@ constexpr llvm::StringLiteral kBaseCTAIndexAttr = "ttl.base_cta_index";
...
+      }
+    };
+    struct DstCBKeyHash {
+      std::size_t operator()(const DstCBKey &k) const {
+        return llvm::hash_combine(k.dstStartX, k.dstStartY, k.dstEndX,
+                                  k.dstEndY, k.cbIndex);
+      }
+    };
+    using Entry = std::pair<PipeKey, size_t>;
+    std::unordered_map<DstCBKey, SmallVector<Entry>, DstCBKeyHash> groups;
```
brnorris03: This and a few other places -- why not use llvm::DenseMap and DenseMapInfo?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -410,12 +675,61 @@ struct TileStoreLowering : OpConversionPattern<TileStoreOp> {
...
+      return;
+    }
+
+    // Trace to the BindCBOp to get the CB index
+    Value cbVal = traceUnrealizedCasts(dstCB);
+    auto bindOp = cbVal.getDefiningOp<BindCBOp>();
+    if (!bindOp) {
+      return;
+    }
+
```
brnorris03, line 457: Silent early return here -- shouldn't this be an error or warning?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -800,6 +1516,130 @@ struct WaitLowering : OpConversionPattern<WaitOp> {
...
 };
 
+//===----------------------------------------------------------------------===//
+// Pipe conditional operation lowering patterns
+//===----------------------------------------------------------------------===//
+
+struct IfSrcLowering : OpConversionPattern<IfSrcOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
```
brnorris03: Optional since I am not sure it would actually be an improvement, but if trying to reduce similar code, maybe common parts of `IfSrcLowering` and `IfDstLowering` (get core coordinates, build condition, create `scf::IfOp`, inline body) can be shared?

python/ttl/__init__.py
```diff
@@ -41,6 +42,10 @@
     "CompilerOptions",
     "TensorBlock",
     "CopyTransferHandler",
+    "Pipe",
+    "PipeNet",
+    "SrcPipeIdentity",
+    "DstPipeIdentity",
```
brnorris03: are SrcPipeIdentity and DstPipeIdentity actually used externally (or just in the tracer)?

python/ttl/kernel_runner.py
```diff
@@ -112,7 +179,34 @@ def build_kernel_descriptors(
...
+            )
+            print(
+                f"    receiverCBIndex={p.receiverCBIndex}, runtimeArgSlot={p.runtimeArgSlot}"
+            )
+
     for spec in kernel_specs:
+        # runtime_args structure: [x][y][args_per_core].
+        # Each core gets an empty arg list (we use my_x/my_y for indexing).
+        runtime_args = [[[] for _ in range(grid_rows)] for _ in range(grid_cols)]
+
```
brnorris03: How are these supposed to be used? I don't see them passed below to the KernelDescriptor.

test/ttlang/Dialect/TTL/IR/pipes.mlir
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// -----
+
+// CHECK-LABEL: func.func @copy_cb_to_pipe
+// CHECK: %[[CB:.*]] = ttl.bind_cb
+// CHECK: %[[P:.*]] = ttl.create_pipe
+// CHECK: ttl.copy %[[CB]], %[[P]]
+// CHECK: ttl.wait
+func.func @copy_cb_to_pipe() {
```
brnorris03, line 78: nit: checks are a bit loose, e.g., not checking if wait consumes the correct handle

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -1211,3 +1230,36 @@ mlir::LogicalResult mlir::tt::ttl::TransposeOp::verify() {
...
+  }
+
+  // Validate coordinates are non-negative.
+  if (srcX < 0 || srcY < 0) {
+    return emitOpError() << "source coordinates must be non-negative";
+  }
+  if (dstStartX < 0 || dstStartY < 0 || dstEndX < 0 || dstEndY < 0) {
+    return emitOpError() << "destination coordinates must be non-negative";
+  }
+
```
brnorris03, line 1263: Just confirming that self-loop pipes are ok? This allows them, i.e., both src and dest are the same core.

brnorris03 (commented):
Thank you for tackling this beast! Made a first pass with comments.

Would be good to add negative tests (with `--verify-diagnostics`) pipes_invalid.mlir, e.g.
  - Pipe-to-pipe copy: "cannot copy directly between pipes"                        
  - Pipe copy without CB: "pipe transfers require one operand to be !ttl.cb"
  - Negative coordinates: "source/destination coordinates must be non-negative"
  - Attribute/type mismatch: "attributes must match result pipe type"

Could extend `convert_pipe_ops.mlir`, which only tests unicast lowering, with multicast tests, e.g.
- Multicast lowering (`noc_async_write_multicast` + semaphore handshake protocol)
- Loopback multicast (`noc_async_write_multicast_loopback_src` + sender-core skip logic)
- Gather pattern

brnorris03:
I am not a huge fan of using the env var and a JSON file for the gather implementation, but don't really know what better thing could be done given the inability to get the required L1 addresses at compile time; maybe that can be another entry in the tt-metal pain points document.

Actually, I am still kind of confused about when/how this JSON is supposed to be used... Right now it seems like it's never really passed into the KernelDescriptor (as I commented earlier), so this is dead code?

python/ttl/kernel_runner.py
zoecarver: I think this broke in a rebase. Removed

python/ttl/_src/ttl_ast.py
zoecarver: Added a TODO, we can revisit this as an API question

python/ttl/pipe.py
zoecarver: Added validation. I think dst will be clear, its the api name.

python/ttl/ttl_api.py
zoecarver: used a temp file instead.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: I think we'd only share about 5 lines so I'm going to keep it separate.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: I think this cannot be hit today given the API, but fixed anyway.

python/ttl/_src/ttl_ast.py
zoecarver: I am not too concerned about a lot of ops here... unfortunately pipes are somewhat codegen verbose. I think I'll leave it like this instead of the cache but happy to revisit if you feel strongly.

brnorris03: I don't feel strongly, fine to leave as is.

zoecarver:
> Actually, I am still kind of confused about when/how this JSON is supposed to be used... Right now it seems like it's never really passed into the KernelDescriptor (as I commented earlier), so this is dead code?

We use it to figure out what semaphores to allocate when building the kernel descriptor. Is there a better way to send a graph back to python after compilation?

python/ttl/_src/ttl_ast.py
zoecarver: The expression evaluation to really fix this would be annoying and should be its own commit I think. For now I added a better error.

brnorris03:
> > Actually, I am still kind of confused about when/how this JSON is supposed to be used... Right now it seems like it's never really passed into the KernelDescriptor (as I commented earlier), so this is dead code?
> 
> We use it to figure out what semaphores to allocate when building the kernel descriptor. Is there a better way to send a graph back to python after compilation?

I think there is, but I would need to prototype to make sure it works as I suspect it should based on tt-metal browsing. We'll need this for other things most likely, so worth the exploratory effort. I am not going to hold this PR up for this, but maybe have an issue to get back to this. I think the env var + file approach will not be robust enough long term.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: Did a big refactor to use two densemaps.

lib/Dialect/TTL/IR/TTLOps.cpp
zoecarver: Yes, they work, but are kind of pointless. The more interesting case is A -> B then B -> A which also works or the gather pattern of many -> A. In both cases order is not guaranteed, but it does work because each copy will get it's own reserved block to put the dest payload into and the semaphores use inc not set. 

If you wanted to synchronize A -> B then B -> A that would be easy enough to do, you would just wait for the first pipe to finish before issuing the second pipe copy.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: I had to add a third datatype to maintain order. Up to you if you want me to keep whats here now or revert to what I had before which I think is a bit simpler, but either way.

zoecarver:
> Actually, I am still kind of confused about when/how this JSON is supposed to be used... Right now it seems like it's never really passed into the KernelDescriptor (as I commented earlier), so this is dead code?
We use it to figure out what semaphores to allocate when building the kernel descriptor. Is there a better way to send a graph back to python after compilation?
I think there is, but I would need to prototype to make sure it works as I suspect it should based on tt-metal browsing. We'll need this for other things most likely, so worth the exploratory effort. I am not going to hold this PR up for this, but maybe have an issue to get back to this. I think the env var + file approach will not be robust enough long term.

Updated to just get the info from the AST and remove the pipe graph

zoecarver:
> Would be good to add negative tests (with --verify-diagnostics) pipes_invalid.mlir, e.g.
> Pipe-to-pipe copy: "cannot copy directly between pipes"
Pipe copy without CB: "pipe transfers require one operand to be !ttl.cb"
Negative coordinates: "source/destination coordinates must be non-negative"
Attribute/type mismatch: "attributes must match result pipe type"

Added

> Could extend convert_pipe_ops.mlir, which only tests unicast lowering, with multicast tests, e.g.
Multicast lowering (noc_async_write_multicast + semaphore handshake protocol)
Loopback multicast (noc_async_write_multicast_loopback_src + sender-core skip logic)
Gather pattern

`convert_pipe_ops.mlir` covers mcast and `scatter.py` should cover the gather ops gather.

lib/Dialect/TTL/Transforms/PipeGraph.h
```diff
@@ -0,0 +1,263 @@
...
+  bool hasPipes() const { return !receiverCBs.empty(); }
+
+  /// Add a receiver CB mapping for a pipe.
+  void addReceiverCB(int64_t srcX, int64_t srcY, int64_t dstStartX,
+                     int64_t dstStartY, int64_t dstEndX, int64_t dstEndY,
+                     int64_t pipeNetId, int64_t cbIndex, int64_t blockCount,
+                     Location loc, Operation *receiverCopyOp) {
+    PipeKey key{srcX, srcY, dstStartX, dstStartY, dstEndX, dstEndY, pipeNetId};
+    assert(receiverCBs.count(key) == 0 &&
+           "duplicate receiver CB for the same pipe");
```
brnorris03: assert on duplicate receiver CB is user-triggerable (two `ttl.copy %pipe, %cb` for the same pipe), so emitError with `return failure()` seems more appropriate here (and need negative test if added)

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -731,12 +795,44 @@ struct CopyLowering : OpConversionPattern<CopyOp> {
...
+        auto pipeType = llvm::cast<PipeType>(adaptor.getDst().getType());
+        receiverInfo = pipeGraph->getReceiverInfo(
+            pipeType.getSrcX(), pipeType.getSrcY(), pipeType.getDstStartX(),
+            pipeType.getDstStartY(), pipeType.getDstEndX(),
+            pipeType.getDstEndY(), pipeType.getPipeNetId());
+      }
+      // Determine CB access context: consumer (cb_wait/cb_pop) vs producer
+      // (cb_reserve/cb_push). This controls whether we read from the CB's
+      // read pointer or write pointer for the pipe source address.
+      bool isConsumerCB = llvm::any_of(
```
brnorris03: `isConsumerCB` check here uses `any_of` over all users of the CB, not whether this specific `CopyOp` is after a `cb_wait`. A CB used in both producer and consumer roles would always return true. Could use `mlir::DominanceInfo` to check whether a `CBWaitOp` on the same CB dominates this `CopyOp`.

lib/Dialect/TTL/Transforms/PipeGraph.h
```diff
@@ -0,0 +1,263 @@
...
+        continue;
+      }
+      gatherRecvProgress[copyOp] = ++dstCounters[dk];
+    }
+  }
+
+  /// Verify that gather receiver CBs have enough blocks for all senders.
+  /// Each sender writes to a different slot, so block_count must be >= the
+  /// number of senders targeting that CB.
+  LogicalResult verifyGatherBlockCounts() const {
```
brnorris03, line 197: `verifyGatherBlockCounts` only checks the first matching receiver entry per destination group (`break` after first match). If multiple entries share the same destination but have different `block_count` values, only the first is validated.

lib/Dialect/TTL/Transforms/PipeGraph.h
```diff
@@ -0,0 +1,263 @@
...
+               cbIndex == o.cbIndex;
+      }
+    };
+    struct DstCBKeyHash {
+      std::size_t operator()(const DstCBKey &k) const {
+        return llvm::hash_combine(k.dstStartX, k.dstStartY, k.dstEndX,
+                                  k.dstEndY, k.cbIndex);
+      }
+    };
+    std::unordered_map<DstCBKey, SmallVector<PipeKey>, DstCBKeyHash> groups;
```
brnorris03: Another candidate for using DenseMap (and `dstCounters` below)?

test/ttlang/Dialect/TTL/Transforms/convert_pipe_ops.mlir
```diff
@@ -0,0 +1,137 @@
+// RUN: ttlang-opt %s --split-input-file -convert-ttl-to-ttkernel | FileCheck %s
```
brnorris03, line 1: nit: only CHECK: used throughout, should be able to have at least some CHECK-NEXT

lib/Dialect/TTL/IR/TTLOps.cpp
brnorris03: That makes sense, thanks for explaining!

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -1211,3 +1230,36 @@ mlir::LogicalResult mlir::tt::ttl::TransposeOp::verify() {
...
+    return emitOpError() << "attributes must match result pipe type";
+  }
+
+  // Validate coordinates are non-negative.
+  if (srcX < 0 || srcY < 0) {
+    return emitOpError() << "source coordinates must be non-negative";
+  }
+  if (dstStartX < 0 || dstStartY < 0 || dstEndX < 0 || dstEndY < 0) {
+    return emitOpError() << "destination coordinates must be non-negative";
+  }
```
brnorris03, line 1262: So just noticed that this doesn't verify start <= end for either dim, is that intentional?

python/ttlang/TTLModule.cpp
```diff
@@ -103,4 +103,29 @@ void populateTTLModule(nb::module_ &m) {
...
+          nb::arg("dst_start_x"), nb::arg("dst_start_y"), nb::arg("dst_end_x"),
+          nb::arg("dst_end_y"), nb::arg("pipe_net_id"))
+      .def_prop_ro("src_x", &PipeType::getSrcX)
+      .def_prop_ro("src_y", &PipeType::getSrcY)
+      .def_prop_ro("dst_start_x", &PipeType::getDstStartX)
+      .def_prop_ro("dst_start_y", &PipeType::getDstStartY)
+      .def_prop_ro("dst_end_x", &PipeType::getDstEndX)
+      .def_prop_ro("dst_end_y", &PipeType::getDstEndY)
+      .def("is_unicast", &PipeType::isUnicast)
+      .def("is_multicast", &PipeType::isMulticast);
```
brnorris03, line 131: Missing `pipe_net_id` read accessor -- `get()` accepts it but can't read it back from Python.

brnorris03 (approved):
Thank you for all the updates!

lib/Dialect/TTL/IR/TTLOps.cpp
zoecarver: Yes, that is intentional. We already normalize with min/max to get the actual range bound regardless of ordering.

PR #506, opened 2026-04-14
Automatic push/pop insertion for dfbs

zoecarver:
Adds an early MLIR pass (ttl-insert-cb-sync) that finds `cb_reserve`/`cb_wait` ops without matching `cb_push`/`cb_pop` and inserts them after the last transitive use of the CB's data. This lets users omit push/pop calls entirely. Explicit push/pop and context managers (with) are preserved when present.


zoecarver:
Test cases:
  - compute_reserve_no_push: reserve + store, no push. Push auto-inserted after store.                                                                                          
  - compute_wait_no_pop: wait + add, no pop. Pop auto-inserted after add.                                             
  - dm_reserve_copy_chain: DM thread reserve + copy + wait chain. Push after ttl.wait (not after copy).                                                                         
  - explicit_push_preserved: explicit push already present. No double-insert.                                                                                                   
  - mixed_explicit_implicit: two CBs, one with explicit push, one without. Only the missing one gets inserted.                                                                  
  - multiple_waits: two waits on different CBs, both need auto-pop.                                                                                                             
  - same_cb_double_wait_mixed: same CB waited twice sequentially. First has explicit pop, second gets auto-pop.
  - wait_inside_loop: wait + use inside scf.for body. Pop inserted inside the loop body.                                                                                        
  - pop_hoisted_from_branch: pop in one scf.if branch, missing from else. Nested pop erased, single pop hoisted after the scf.if.
  - reserve_no_uses: reserve with zero uses. Push right after reserve.                                                                                                          
  - attach_cb_dead_result: attach_cb with unused result. Push after attach.                                           
  - two_copies_same_cb: two copies on the same CB in DM thread. Push after the last ttl.wait.                                                                                   
  - pops_both_branches: pops in both scf.if branches. Both erased, single pop hoisted after scf.if.                                                                             
  - reserve_outside_loop: reserve outside loop, store inside. Push after the scf.for.
  - if_inside_for: scf.if nested inside scf.for, pop in one if-branch. Pop hoisted out of if, stays in for body.                                                                
  - nested_for_loops: nested scf.for loops, wait + use in inner loop. Pop in inner loop body, nothing leaks out.

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
```diff
@@ -0,0 +1,246 @@
...
+  return hasSameLevelRelease;
+}
+
+/// Compute the last operation (in the acquire's block) in the transitive
+/// use closure of an acquire.
+///
+/// Uses in nested regions are projected up to their ancestor in the
+/// acquire's block (e.g., an add inside an scf.if projects to the scf.if).
+static Operation *findLastTransitiveUse(Value cb, Operation *acquire,
+                                        Operation *bound) {
```
brnorris03, line 95: I think it may be possible to use `getForwardSlice` + `Block::findAncestorOpInBlock` and avoid having manual worklist/visited/cycle-detection infrastructure here. What do you think?

zoecarver: Happy to use `findAncestorOpInBlock`, I think `getForwardSlice` would be hard because we'd still need a lot of logic to loop over cb uses, filter, last finding post pass, etc. so I don't really want to use getForwardSlice.

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+      CBPushOp::create(builder, reserve.getLoc(), cb,
+                       /*num_tiles=*/IntegerAttr{});
+    }
+
+    for (auto wait : waits) {
+      Value cb = wait.getCb();
+
+      Operation *nextWait = nullptr;
+      for (auto other : waits) {
+        if (other == wait || other.getCb() != cb)
```
brnorris03: This loop is very similar to the next one, consider extracting a templated helper parameterized by the acquire/release op type and taking a release-creation callback.

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+      return false;
+    return true;
+  };
+
+  for (auto &use : cb.getUses()) {
+    Operation *user = use.getOwner();
+    if (user == acquire)
+      continue;
+    if (isa<CBPushOp, CBPopOp, CBReserveOp, CBWaitOp>(user))
+      continue;
```
brnorris03: nit: getAncestorInBlock will be called twice for each user

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+template <typename ReleaseOpTy>
+static bool findReleases(Value cb, Operation *acquire, Operation *bound,
+                         SmallVectorImpl<ReleaseOpTy> &allReleases,
+                         SmallVectorImpl<ReleaseOpTy> &toHoist,
+                         const DenseSet<Operation *> &erased) {
+  Block *block = acquire->getBlock();
+  bool hasSameLevelRelease = false;
+
+  for (auto release : allReleases) {
+    if (erased.contains(release))
```
brnorris03: nit: `const SmallVectorImpl<ReleaseOpTy> &allReleases`

test/ttlang/Dialect/TTL/Transforms/insert_cb_sync.mlir
```diff
@@ -0,0 +1,432 @@
...
+
+// CHECK-LABEL: func.func @multiple_waits
+// CHECK: %[[CB0:.+]] = ttl.bind_cb{cb_index = 0
+// CHECK: %[[CB1:.+]] = ttl.bind_cb{cb_index = 1
+// CHECK: ttl.cb_wait %[[CB0]]
+// CHECK: ttl.cb_wait %[[CB1]]
+// CHECK: ttl.add
+// CHECK: ttl.cb_pop
+// CHECK: ttl.cb_pop
+// CHECK-NOT: ttl.cb_pop
```
brnorris03: check which CB was popped?

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
```diff
@@ -0,0 +1,239 @@
...
+        erased.insert(nested);
+        nested.erase();
+      }
+
+      Operation *last = findLastTransitiveUse(cb, reserve, nextReserve);
+      builder.setInsertionPointAfter(last);
+      CBPushOp::create(builder, reserve.getLoc(), cb,
+                       /*num_tiles=*/IntegerAttr{});
+    }
+
```
brnorris03: So for an example like the one below, the user push would be erased? Would be good to add a test for this.
```python
@ttl.compute()
def compute_fn():
    blk = out_dfb.reserve()
    if cond:
        blk.store(result_a)
        blk.push()
    else:
        blk.store(result_b)
```

brnorris03 (commented):
LGTM, a few minor things.

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
zoecarver: Currently the push will be moved outside of the if-stmt, there are mlir tests for this case.

lib/Dialect/TTL/Transforms/TTLInsertCBSync.cpp
zoecarver: refactored into lambda

zoecarver:
All comments addressed, thank you @brnorris03!

brnorris03 (approved):
LGTM, thank you for the updates!

PR #514, opened 2026-04-16
[ttl] Implement DFB index reuse and consolidate DFB management

brnorris03:
This patch implements compiler-allocated DFB index reuse via linear scan allocation in `ttl-finalize-dfb-indices`. Reduces L1 pressure by assigning the same index to intermediate DFBs with non-overlapping lifetimes. Moves the `kMaxCircularBuffers` limit check from insertion to finalization (after reuse). Reorders the pipeline so finalization precedes CB annotation to ensure valid indices are used.

When capacity is exceeded, `ttl-finalize-dfb-indices` produces this error:
```error: need 33 DFB indices but hardware supports at most 32 (1 compiler-allocated after reuse); reduce the number of user-declared dataflow buffers or split the computation into multiple kernels.```

Design and algorithm details in the new design doc `docs/development/DFBManagement.md`.


zoecarver (approved):
Nice!

PR #516, opened 2026-04-16
[nfc] rename --ttl-auto-sync option to --ttl-subblock-sync

brnorris03:
The old option name is misleadingly broad, especially given the recent addition of general push/pop insertion by the compiler (not guarded by an option), so renaming `--ttl-auto-sync` option to `--ttl-subblock-sync` to reflect more precisely what it actually does.


arichinsTT (approved):
LGTM

PR #517, opened 2026-04-16
Add automatic wait insertion for copy

zoecarver:
Automatically inserts wait for copy if user has not provided a specific wait op.


brnorris03 (approved):
Looks great, can you also please update `docs/sphinx/reference/compiler-options.md` with the new option. Thanks!

PR #520, opened 2026-04-16
DFB over allocation error

### LINKED ISSUE #511: [ttl] Emit compiler error when CB allocation exceeds L1 capacity
When a kernel's circular buffers exceed L1 memory, the user gets an opaque tt-metal runtime crash:

```
RuntimeError: TT_THROW @ .../program.cpp:1043: tt::exception
info:
Statically allocated circular buffers on core range [(x=0,y=0) - (x=12,y=9)] grow to 1683968 B which is beyond max L1 size of 1572864 B
```

The compiler knows CB sizes and block shapes at allocation time. It should validate total L1 usage and emit a clear diagnostic before code generation, e.g.:

```
error: total CB allocation (1683968 B) exceeds L1 capacity (1572864 B) on core range [(0,0)-(12,9)]
  CB0 (a_dfb): 8x8 tiles = 131072 B
  CB1 (b_dfb): 8x8 tiles = 131072 B
  ...
note: reduce block shapes or number of DFBs to fit in L1
```

**Reproducer**: any kernel with enough DFBs and large block shapes, e.g., 8 DFBs with `shape=(8,8)` blocks.

arichinsTT:
### Problem description                                                                                                                                                                                                                          
Oversized static DFB backing stores fail late in tt-metal with cryptic L1/CB
diagnostics. Because compiler-inserted DFBs only become visible after late
MLIR passes, Python-only pre-flight checks are insufficient. Closes #511.                                                                                                                                                                
                                                                                                                                                                                                                                                   
### Changes                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                   
For hardware runs, the budget is resolved in priority order: a user-supplied 
override via compilation flag takes precedence, then a live device query 
(accounting for reduced `worker_l1_size` and existing tensor allocations on 
the worst-case core), then `ChipDescAttr::getUsableL1Size()` from the 
MLIR system descriptor, and finally a static 1432 KiB fallback that matches 
the simulator. Note: for mesh devices, L1 usage is sampled from the parent 
mesh device rather than per-device submeshes; this is a known 
approximation for multi-device cases.  This relies on uniform tensor 
distribution across devices, which tt-lang now requires and enforces for 
tensor arguments.

                                                                                                                                                                                                                                                   
### Ticket      
Closes #511

Checklist


zoecarver:
Example of custom program size: https://github.com/zoecarver/tt-lang-models/blob/main/micelle-demo/kernels/test_bond_f32.py#L304

Example of constants in L1 for fast access: https://github.com/zoecarver/nanochat/blob/master/ttlang/inference.py#L1125

python/ttl/kernel_runner.py
```diff
@@ -35,13 +35,41 @@ def _ensure_ttnn():
...
+    """
+    Get the remaining L1 for a core.
+    handles reduced worker_l1_size devices and tensor L1 allocations
+    ex:
+    device = ttnn.open_device(device_id=0, worker_l1_size=reduced_l1_size)
+    tensor = ttnn.empty([1024, 1024], device=device, memory_config=ttnn.L1_MEMORY_CONFIG)
+
+    this helper should return 1436032 bytes for core (0, 0)
+    """
+    _ensure_ttnn()
```
brnorris03, line 62: If the ttnn import failed, `_ensure_ttnn()` leaves `ttnn` as `None` and this line raises `AttributeError` instead of the `RuntimeError("ttnn is not available")` that the rest of the file uses. Can we add the same guard here?

lib/Dialect/TTL/Transforms/TTLValidateCBBudget.cpp
```diff
@@ -0,0 +1,218 @@
...
+    return std::nullopt;
+  }
+
+  auto chipIds = deviceOp.getDeviceAttr().getChipIds();
+  if (chipIds.empty()) {
+    return std::nullopt;
+  }
+
+  mlir::tt::ttcore::ChipDescAttr chip = systemDesc.getChipDesc(chipIds[0]);
+  return static_cast<uint32_t>(chip.getUsableL1Size());
```
brnorris03: `getUsableL1Size()` returns `unsigned` (can be 32 bits), but `totalBytes` below is `uint64_t`. Better to be consistent and use unsigned 64-bit ints everywhere for the budget (return type here, `budgetBytes` ~ line 119, and `fallbackUsableL1Bytes` ~ line 54).

lib/Dialect/TTL/Transforms/TTLValidateCBBudget.cpp
```diff
@@ -0,0 +1,218 @@
...
+  if (!deviceOp) {
+    return std::nullopt;
+  }
+
+  auto chipIds = deviceOp.getDeviceAttr().getChipIds();
+  if (chipIds.empty()) {
+    return std::nullopt;
+  }
+
+  mlir::tt::ttcore::ChipDescAttr chip = systemDesc.getChipDesc(chipIds[0]);
```
brnorris03: This always selects chip 0. I don't know if such configurations exist, but if at some point there are multiple chips with different `usable_l1_size`, the budget may be incorrect. Taking a min over all `chipIds` seems safest, or add a TODO to confirm this can't happen.

python/ttl/kernel_runner.py
```diff
@@ -191,7 +228,40 @@ def build_cb_descriptors(
...
+                    page_size,
+                    total_size,
+                    f"  CB[{i}]: shape={cb.shape} block_count={cb.block_count} -> {total_size} bytes",
+                )
+            )
+
+        total_cb_bytes += total_size
 
+    remaining_bytes = DEFAULT_L1_CB_BUDGET_BYTES
+    for tensor in tensors:
```
brnorris03, line 249: Are those necessarily in L1?

lib/Dialect/TTL/Transforms/TTLValidateCBBudget.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "ttlang/Dialect/TTL/Passes.h.inc"
+
+namespace {
+
+/// Fallback when the module has no `system_desc` / device: same static CB
+/// region per core as Wormhole and Blackhole (1464 KiB L1 minus 32 KiB
+/// reserved in tt-metal dev_mem_map). Matches
+/// `ttl.constants.DEFAULT_L1_CB_BUDGET_BYTES` and
+/// `ChipDescAttr::getUsableL1Size()` for those chips when IR carries attrs.
+static constexpr uint32_t fallbackUsableL1Bytes =
```
brnorris03: nit: constant names start with k, so `kFallbackUsableL1Bytes` (naming conventions in llvm projects)

python/ttl/kernel_runner.py
```diff
@@ -191,7 +228,40 @@ def build_cb_descriptors(
...
+        if tensor is not None and hasattr(tensor, "device"):
+            device = tensor.device()
+            if device is None:
+                continue
+            remaining_bytes = get_remaining_l1_for_core(device)
```
brnorris03: This call uses `get_remaining_l1_for_core`'s default `core_x=0, core_y=0`. For a kernel running on a different core or grid, or with sharded L1 tensors that don't touch (0,0), the result can be too pessimistic or too optimistic. Could we accept the worst case across the kernel's `core_ranges` instead, or add a TODO + issue to track it?

python/ttl/kernel_runner.py
```diff
@@ -191,7 +228,40 @@ def build_cb_descriptors(
...
+            break
```
brnorris03, line 255: Two cases this doesn't handle:
1. If `tensors` contains tensors on different devices, only the first one's budget is consulted (the `break` here).
2. If `tensor.device()` returns a `MeshDevice` (mesh tensor case — `_is_mesh_tensor` exists in `ttl_api.py`), passing it into `get_remaining_l1_for_core` may not give the right answer per physical core.

Could we either restrict to single-device homogeneous setups (assert), handle the mesh case explicitly, or add a TODO + open an issue to track it?

python/ttl/ttl_api.py
```diff
@@ -1195,6 +1196,14 @@ def _compile_kernel(
             )
             print(f"[TTNN interop] Detected {memory_space} memory space")
 
+    l1_budget_override = compiler_options.l1_budget
+    if l1_budget_override == 0 and has_ttnn_tensors:
+        try:
+            device = _require_device(args)
+            l1_budget_override = get_remaining_l1_for_core(device)
+        except ValueError:
```
brnorris03, line 1285: If `_require_device(args)` raises `ValueError` (no device-bearing tensors despite `has_ttnn_tensors` being true), `l1_budget_override` stays at 0 and the pass uses the WH/BH fallback constant instead of the actual device budget. Suggest re-raising rather than suppressing.

test/ttlang/Dialect/TTL/Transforms/validate_cb_budget.mlir
```diff
@@ -0,0 +1,214 @@
+// Tests for ttl-validate-cb-budget: overflow, multi-function/index behavior, and all four
+// layout/dtype combinations for CB element types:
+//   - ttcore.tile<32x32, bf16>  -> 2048 bytes per slot (explicit tile)
+//   - ttcore.tile<32x32, f32>   -> 4096 bytes per slot (explicit tile)
+//   - bf16 (row-wise, builtin)  -> TileType::get(bf16)  -> 2048 bytes per slot
+//   - f32  (row-wise, builtin)  -> TileType::get(f32)   -> 4096 bytes per slot
```
brnorris03, line 6: No test exercises `tryBudgetFromModule` (the `ttcore.system_desc` → `ChipDescAttr::getUsableL1Size()` path) — all tests rely on the WH/BH fallback constant. Could we add a module with `ttcore.system_desc` + a `default device` so that branch has coverage? Same for the override-vs-system_desc precedence (override should win) — currently override is only tested against the fallback.

python/ttl/kernel_runner.py
arichinsTT: nope, but I need to get the device from them.  tensors in L1 will be accounted for by the allocator in get_remaining_l1_for_core

python/ttl/ttl_api.py
arichinsTT: Thats the behavior I want, if we can't get a device, then I can't calculate the remaining l1 more accurately, so the pass will use the defaults first trying from the mlir arch descriptor, then using the same default as the sim as a final fallback

python/ttl/kernel_runner.py
arichinsTT: changed function to operate on all devices within the mesh using sub meshes to expose some of the apis I need.  I think it is a safe assumption to have that if a tensor has a device, that the device should be the same for all the tensors and for what the kernel is running on, so I can grab the first one available, but still do calculations on all, accounting for non uniform shardings and giving back the minimal usable space

test/ttlang/Dialect/TTL/Transforms/validate_cb_budget.mlir
arichinsTT: test cases added!

brnorris03:
This looks like a rebase mishap — the diff accidentally reverts #576 (the `ttl._pipenets` packaging fix). Could you rebase onto current main and undo that?

test/ttlang/Dialect/TTL/Transforms/validate_cb_budget_l1_override.mlir
```diff
@@ -0,0 +1,25 @@
+// Tests for ttl-validate-cb-budget with the l1-budget-override pass option.
+//
+// A small override budget (4096 bytes) causes a normally-passing CB to fail.
+// A large override budget (2000000 bytes) causes a normally-failing CB to pass.
+//
+// RUN: ttlang-opt %s --split-input-file --verify-diagnostics -pass-pipeline='builtin.module(ttl-validate-cb-budget{l1-budget-override=4096})'
```
brnorris03, line 7: Is the override-vs-`system_desc` precedence covered anywhere? Both override files use bare `func.func`, so it looks like only override-vs-fallback is exercised. Would a split section pairing the 204800-byte `system_desc` with `l1-budget-override=4096` make sense?

python/ttl/ttl_api.py
```diff
@@ -1250,6 +1251,14 @@ def _compile_kernel(
             )
             print(f"[TTNN interop] Detected {memory_space} memory space")
 
+    l1_budget_override = compiler_options.l1_budget
+    if l1_budget_override == 0 and has_ttnn_tensors:
+        try:
+            device = _require_device(args)
+            l1_budget_override = get_min_remaining_l1_for_device(device)
+        except ValueError:
```
brnorris03, line 1285: Probably not an issue, but just checking -- on the `except ValueError: pass` — is it meant to only ignore the "no device-bearing tensor" case from `_require_device`? As written it would also not raise if `get_min_remaining_l1_for_device` ever raises `ValueError` for some other reason. Is the scope of this try: intentionally broad?

```python
try:
    device = _require_device(args)
except ValueError:
    device = None
if device is not None:
    l1_budget_override = get_min_remaining_l1_for_device(device)
```

lib/Dialect/TTL/Transforms/TTLValidateCBBudget.cpp
```diff
@@ -0,0 +1,219 @@
...
+//===----------------------------------------------------------------------===//
+//
+// TTL Validate CB Budget
+//
+// Validates that the sum of static circular-buffer backing stores (per unique
+// cb_index) does not exceed a per-core L1 budget. Per-slot sizes use
+// ttcore::TileType::getSizeBytes() when the CB already carries a tile type, and
+// ttcore::TileType::get(elemTy).getSizeBytes() for row-wise / scalar element
+// types (same rule as tt-mlir DeviceAttr::getMemrefCBPageSizeBytes). Python
+// uses ttl.dtype_utils.tile_bytes_from_dtype — if those ever diverge, align
```
brnorris03: Nit: I think the diverging Python check actually lives in `python/ttl/kernel_runner.py:build_cb_descriptors` rather than `ttl.dtype_utils` — would it be worth pointing the "if those ever diverge" note at the sibling check directly so it's easier to grep for?

brnorris03 (changes requested):
Mostly minor things, except the reverting of #576 during rebase (that functionality is required for correct docker and wheel builds).

test/python/test_gdn_kernel.py
```diff
@@ -1,243 +0,0 @@
-# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
brnorris03, line 1: Got deleted — can you restore it?

python/ttl/kernel_runner.py
```diff
@@ -191,7 +234,40 @@ def build_cb_descriptors(
...
+        total_cb_bytes += total_size
+
+    remaining_bytes = DEFAULT_L1_CB_BUDGET_BYTES
+    for tensor in tensors:
+        if tensor is not None and hasattr(tensor, "device"):
+            device = tensor.device()
+            if device is None:
+                continue
+            remaining_bytes = get_min_remaining_l1_for_device(device)
+            break
```
brnorris03, line 255: First tensor with a non-None device wins and the loop breaks. If `tensors` spans multiple devices, only the first one's budget is used. Is that invariant enforced upstream? A short assert or a comment pointing at where it's established would help.

python/sim/context_types.py
```diff
@@ -17,9 +17,8 @@
 from .typedefs import Count, Shape, BindableTemplate
 from .blockstate import ThreadType
 
-# Default L1 memory limit per core: Blackhole/Wormhole L1 (1464K) minus
-# large-program reserved space (128K).
-DEFAULT_MAX_L1_BYTES: int = (1464 - 128) * 1024  # 1336 KiB = 1_368_064 bytes
+# Default L1 memory limit per core (simulator): same as runtime CB budget helper.
+DEFAULT_MAX_L1_BYTES: int = 1432 * 1024
```
brnorris03, line 22: Can we use `ttl.constants.DEFAULT_L1_CB_BUDGET_BYTES` here directly? The two will drift if either side changes. If `sim` can't import from `ttl`, a `# keep in sync with ttl.constants.DEFAULT_L1_CB_BUDGET_BYTES` line at least makes it grep-findable.

python/ttl/kernel_runner.py
```diff
@@ -35,13 +35,47 @@ def _ensure_ttnn():
...
+    per-core L1 buffer usage reported by the device.
+
+    ``get_buffer_pages`` is called on the original device rather than on
+    per-coordinate submeshes because ``create_submesh`` produces a new
+    device view that does not inherit buffer tracking from the parent.
+    For mesh devices this reports allocations for the first physical
+    device, which is representative because tt-lang distributes tensors
+    uniformly across the mesh. if it becomes needed to track individual
+    physical devices in the submesh, ttnn.reports.get_buffer_pages will
+    need to be changed to reflect allocations on the parent mesh.
```
brnorris03: Docstring nit: "if it becomes needed to track individual physical devices..." reads oddly. Maybe: "If individual physical devices need tracking, `ttnn.reports.get_buffer_pages` would have to report allocations on the parent mesh."

test/python/test_gdn_kernel.py
arichinsTT: ahhh I thought this was one of my variations that slipped through, will add back

python/ttl/kernel_runner.py
arichinsTT: I think its somewhat unreasonable to have tensors span multiple different mesh devices, as we run the kernel on the same device with the same pickup pattern.  Will add a check in _require_device to have all ttnn tensors using the same mesh device

python/ttl/ttl_api.py
arichinsTT: if get_min_remaining_l1_for_device fails in some way, I thought it better that the compilation moves on and uses progressive fallbacks.

python/ttl/kernel_runner.py
brnorris03: I am not really familiar with what's allowed (in tt-metal land), so whatever you think makes sense.

brnorris03 (approved):
lgtm, thank you for all the updates! nice work!

test/python/test_missing_device.py
```diff
@@ -138,3 +141,139 @@ def test_auto_grid_mixed_host_and_device(device):
...
+
+
+# =========================================================================
+# Real TTNN tensors on a single device (requires hardware)
+# =========================================================================
+
+
+def test_require_device_two_tensors_same_device(device):
+    """Two real TTNN tensors on the same device should pass."""
+    from ttl.ttl_api import _require_device
```
zoecarver: nit: import at the top of the file rather than several times throughout

lib/Dialect/TTL/Transforms/TTLValidateCBBudget.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    bool hadFailure = false;
+    moduleOp.walk([&](BindCBOp bindOp) {
+      if (hadFailure) {
+        return;
+      }
+      FailureOr<uint64_t> bytes = cbBytesForBind(bindOp);
+      if (failed(bytes)) {
+        hadFailure = true;
+        return;
```
zoecarver: Optional nit: use `WalkResult::interrupt()`

zoecarver (approved):

test/python/test_missing_device.py
```diff
@@ -138,3 +141,134 @@ def test_auto_grid_mixed_host_and_device(device):
...
+
+    def id(self):
+        return self._id
+
+    def __repr__(self):
+        return f"FakeDevice({self._id})"
+
+
+def test_same_device_identity():
+    from ttl.ttl_api import _same_device
```
zoecarver: still a few of these, optional

arichinsTT: oops good catch!

PR #507, opened 2026-04-15
[ttl] Thread local DFBs

brnorris03:
Automatically insert intermediate dataflow buffers when a fused expression chain feeds into an op whose tile-level lowering reads from the unpacker (reduce, bcast, matmul, transpose). Previously the user had to manually declare and wire these buffers; now the compiler materializes them.

A new `DFBInputOpInterface` declares which operands of a tensor-level op require DFB-attached values. The `ttl-insert-intermediate-dfbs` pass walks all ops implementing this interface and, for each non-DFB-attached operand, inserts a scratch DFB with `bind_cb`, `cb_reserve`, `store`, `cb_wait`, and `attach_cb`. The `ttl-insert-cb-sync` pass then fills in the matching `cb_push` and `cb_pop`.

Scratch DFBs use `block_count=2` because the hardware requires double-buffering when the packer and unpacker operate on the same CB within one thread. They are tagged with `ttl.compiler_allocated` so that `ttl-finalize-dfb-indices` can collect them into a `ttl.scratch_dfbs` module attribute. The Python runtime reads this attribute after compilation and builds the corresponding `CBDescriptor` entries for `ttnn.generic_op`.

The `ttl-insert-cb-sync` push-placement logic is extended to bound the search by the next `cb_wait` on the same CB, not just the next `cb_reserve`. Without this, `cb_push` for intra-thread DFBs was placed after `cb_pop`, causing a deadlock.

Pipeline order: `ttl-insert-intermediate-dfbs` runs first (creates scratch DFBs), then `ttl-insert-cb-sync` (inserts push/pop for all DFBs including scratch), then `convert-ttl-to-compute` (sees all operands as DFB-attached).

Tested with add->reduce, matmul->reduce, and a multinode per-tile softmax kernel where the user declares only I/O and scaler DFBs and the compiler creates three intermediates for `reduce_max`, `exp`, and `reduce_sum` results.

Not in this PR: 
- DFB index reuse. Currently each scratch DFB gets a fresh index, which can lead to unnecessary L1 pressure when many intermediates are created. A follow-up PR will implement linear scan allocation to reuse indices when lifetimes do not overlap.
- Result materialization (e.g., DST result to DFB-input op), #508

Closes #474

PR #523, opened 2026-04-17
Fix doulbe use dfb push being dropped (519)

### LINKED ISSUE #519: [external bug report][from claude] Missing cb_push_back / double cb_push_back when two passes share a dataflow buffer
## Symptom

Runtime hang on hardware. Compilation succeeds, generated compute kernel is missing a `cb_push_back` in the first "pass" and emits a compensating double `cb_push_back` in the second "pass". Total push/reserve counts balance, so the imbalance isn't obvious from counts alone.

ttl version: `ttlang 0.1.8.dev38`

## Minimal reproducer

Two back-to-back `with x.wait() as x0: with red.reserve() as r: r.store(reduce_sum(x0, sc, dims=[1]))` statements, sharing `red_dfb` (and `acc_dfb`). No outer for-loop, no if-guard, no broadcast — just two passes of the same pattern.

```python
"""Minimal reproducer: two passes sharing red/acc DFBs cause the compiler to
relocate cb_push_back from pass 1 into pass 2 (pass 2 emits the push twice)."""
import torch, ttnn, ttl

TILE = 32


@ttl.operation(grid=(1, 1))
def repro(x, scaler, out):
    x_dfb = ttl.make_dataflow_buffer_like(x, shape=(1, 1), block_count=2)
    sc_dfb = ttl.make_dataflow_buffer_like(scaler, shape=(1, 1), block_count=1)
    red_dfb = ttl.make_dataflow_buffer_like(scaler, shape=(1, 1), block_count=2)
    acc_dfb = ttl.make_dataflow_buffer_like(scaler, shape=(1, 1), block_count=2)
    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)

    @ttl.compute()
    def compute():
        with sc_dfb.wait() as sc:
            # Pass 1
            with x_dfb.wait() as x0:
                with red_dfb.reserve() as r:
                    r.store(ttl.math.reduce_sum(x0, sc, dims=[1]))
            with red_dfb.wait() as rv, acc_dfb.reserve() as acc:
                acc.store(rv)
            # Pass 2 (identical reduce pattern; sinks accumulator)
            with x_dfb.wait() as x0:
                with red_dfb.reserve() as r:
                    r.store(ttl.math.reduce_sum(x0, sc, dims=[1]))
            with red_dfb.wait() as rv, acc_dfb.wait() as av, acc_dfb.reserve() as acc:
                acc.store(av + rv)
            with acc_dfb.wait() as a, out_dfb.reserve() as o:
                o.store(a)

    @ttl.datamovement()
    def dm_read():
        ttl.copy(scaler[0, 0], sc_dfb.reserve()).wait()
        ttl.copy(x[0, 0], x_dfb.reserve()).wait()
        ttl.copy(x[0, 0], x_dfb.reserve()).wait()  # re-read for pass 2

    @ttl.datamovement()
    def dm_write():
        ttl.copy(out_dfb.wait(), out[0, 0]).wait()


if __name__ == "__main__":
    x = torch.randn(TILE, TILE, dtype=torch.bfloat16)
    scaler = torch.ones(TILE, TILE, dtype=torch.bfloat16)
    xt = ttnn.from_torch(x, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
    st = ttnn.from_torch(scaler, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
                         device=device, memory_config=ttnn.L1_MEMORY_CONFIG)
    ot = ttnn.from_torch(torch.zeros(TILE, TILE, dtype=torch.bfloat16),
    repro(xt, st, ot)
    print("ran (this line should not print if bug reproduces)")
    ttnn.close_device(device)
```

## Generated compute kernel (relevant excerpt)

```cpp
// Pass 1 prologue (lines 54-67):
cb_reserve_back(arg2)   // red.reserve
...reduce_sum, pack(arg2)...
cb_pop_front(arg0)      // x.pop
// MISSING: cb_push_back(arg2)
cb_wait_front(arg2)     // DEADLOCK

// Pass 2 (lines 79-90):
cb_reserve_back(arg2)   // red.reserve again
...reduce_sum, pack(arg2)...
cb_push_back(arg2)      // <-- compensating
cb_push_back(arg2)      // <-- doubled (absorbs the missing one from pass 1)
```

The same relocation happens on `acc_dfb` (CB3): the push from pass 1's `acc.store(rv)` is also missing and appears as a double-push later in pass 2.

## Environment

- Real hardware (Blackhole), not simulator.
- Reproduced on one chip.

Related: #518 (different assertion in the same compiler pipeline).

zoecarver:
Fixes #519. 

Root cause: `emitTileStores()` scanned all uses of a CB value to find `cb_push` ops to move after the newly created `ttl.compute`. When two compute blocks wrote to the same CB, the second compute block's processing would find and re-relocate the first compute block's push (already moved after compute1, but still before compute2), grouping both pushes together after the last compute.                                                                                                 
                                                                                                                                                                          
Fix: Walk forward from each store to find only the first `CBPushOp` on the same CB (the one that belongs to that specific store) rather than scanning all CB uses.    

Introduced in 474e10cd: [ttl] Thread local DFBs (#507)

### DIFF
--- a/lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
@@ -147,18 +147,18 @@ static void emitTileStores(PatternRewriter &rewriter, Location loc,
         rewriter, loc, tileResult, storeOp.getView(), indices);
     storesToErase.push_back(storeOp);
   }
-  // The DSL emits cb_push after each store. When multiple stores are
-  // absorbed into a single compute body, earlier pushes end up before
-  // the compute and execute before data is packed. Move them after the
-  // compute so the push executes after pack_tile writes the data.
+  // Move each store's cb_push after the compute so the push executes
+  // after pack_tile writes the data. Walk forward from each store to
+  // find the first matching CBPushOp on the same CB.
   for (StoreOp s : storesToErase) {
     Value viewCB = getAttachedCB(s.getView());
     if (viewCB) {
-      for (auto &use : viewCB.getUses()) {
-        if (auto pushOp = dyn_cast<CBPushOp>(use.getOwner())) {
-          if (pushOp->getBlock() == computeOp->getBlock() &&
-              pushOp->isBeforeInBlock(computeOp)) {
+      for (Operation *op = s->getNextNode(); op != nullptr;
+           op = op->getNextNode()) {
+        if (auto pushOp = dyn_cast<CBPushOp>(op)) {
+          if (pushOp.getCb() == viewCB) {
             pushOp->moveAfter(computeOp);
+            break;
           }
         }
       }
--- a/test/ttlang/Dialect/TTL/Transforms/convert_ttl_to_compute_shared_dfb.mlir
+++ b/test/ttlang/Dialect/TTL/Transforms/convert_ttl_to_compute_shared_dfb.mlir
@@ -1,0 +1,73 @@
+// Verifies that convert-ttl-to-compute preserves cb_push ordering when two
+// reduce+store sequences write to the same CB (fixes #519).
+
+// RUN: ttlang-opt %s --pass-pipeline='builtin.module(func.func(convert-ttl-to-compute))' | FileCheck %s
+
+// The input mirrors the reproducer from #519: two back-to-back reduce+store
+// sequences sharing red_dfb (cb2). Each store has a cb_push immediately after.
+// After lowering, each cb_push must remain after its own ttl.compute block;
+// they must not be grouped together after the second compute.
+
+// CHECK-LABEL: func.func @two_reduces_shared_dfb
+// CHECK:       ttl.cb_reserve %[[RED:.*]] :
+//
+// First compute: reduce → tile_store, then push for pass 1.
+// CHECK:       ttl.compute
+// CHECK:       ttl.cb_push %[[RED]]
+//
+// Intervening wait/reserve/store/pop on other CBs.
+// CHECK:       ttl.cb_pop
+// CHECK:       ttl.cb_wait %[[RED]]
+//
+// Second compute: reduce → tile_store, then push for pass 2.
+// CHECK:       ttl.cb_reserve %[[RED]]
+// CHECK:       ttl.compute
+// CHECK:       ttl.cb_push %[[RED]]
+// CHECK:       ttl.cb_pop
+// CHECK:       ttl.cb_wait %[[RED]]
+func.func @two_reduces_shared_dfb()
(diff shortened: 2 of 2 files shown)


brnorris03 (approved):
Thanks for catching and fixing that!

brnorris03:
Maybe add a runtime test?

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -147,18 +147,18 @@ static void emitTileStores(PatternRewriter &rewriter, Location loc,
...
-          if (pushOp->getBlock() == computeOp->getBlock() &&
-              pushOp->isBeforeInBlock(computeOp)) {
+      for (Operation *op = s->getNextNode(); op != nullptr;
+           op = op->getNextNode()) {
+        if (auto pushOp = dyn_cast<CBPushOp>(op)) {
+          if (pushOp.getCb() == viewCB) {
             pushOp->moveAfter(computeOp);
+            break;
           }
         }
```
brnorris03, line 163: Ran into issues with this in #525. The unconditional `moveAfter(computeOp)` regresses multi-store-per-reserve patterns (one reserved slot written by multiple stores in a single `with reserve as v:` scope, e.g. layernorm's `v.store(fill(0)); for _: v += reduce_sum(...); v.store(broadcast(v) * ms)`). The frontend emits one `cb_push` for the whole scope; every store walks forward to it and moves it to after its own `computeOp`, so the first store wins and the push lands before the accumulation loop and the final store — the consumer reads `fill(0)` instead of the final mean.

Suggested fix: keep the pre-#519 guard so we only move pushes that are currently before `computeOp`. Pushes already placed later (by the frontend at end-of-`with:` scope) are already correctly positioned.

```cpp
for (Operation *op = s->getNextNode(); op != nullptr;
     op = op->getNextNode()) {
  if (auto pushOp = dyn_cast<CBPushOp>(op)) {
    if (pushOp.getCb() == viewCB &&
        pushOp->getBlock() == computeOp->getBlock() &&
        pushOp->isBeforeInBlock(computeOp)) {
      pushOp->moveAfter(computeOp);
      break;
    }
  }
}
```

Regression test for the multi-store case in #525: `test/python/test_layernorm.py::test_layernorm_minimal_dfbs` (uses `with mean_dfb.reserve() as mean_blk: mean_blk.store(fill(0)); for _: mean_blk += ...; mean_blk.store(broadcast(mean_blk) * ms)`).

zoecarver:
Subsumed by #525

PR #531, opened 2026-04-22
Simulator: Add all_gather

FConstantinos:


python/sim/ttnnsim.py
```diff
@@ -1939,6 +1939,71 @@ def all_reduce(
     return result_tensor
 
 
+def all_gather(
+    input_tensor: Tensor,
+    dim: int,
+    cluster_axis: Optional[int] = None,
```
phizalev-TT, line 1945: Let's add support for `cluster_axis` here and in `all_reduce` soon after this. It will be important for 2D meshes.

phizalev-TT (approved):
LGTM!

Issue #533, opened 2026-04-24
reduce_max/reduce_sum dims=[1] returns zeros for fp32 tiles (fp32_reduce_acc path)


zoecarver:
## Summary
`ttl.math.reduce_max` and `ttl.math.reduce_sum` with `dims=[1]` silently return all zeros when the input tile is `fp32`. The same kernel on `bfloat16` tiles returns correct per-row values. `dims=[0]` and `dims=[0, 1]` work correctly in fp32; only `dims=[1]` is affected.

The bug is in the fp32-accumulation lowering for reduce: passing `options=\"--no-ttl-reduce-full-fp32\"` to `@ttl.operation` makes fp32 `dims=[1]` return correct values.

## Environment
- ttlang 1.0.0.dev4 (from the Python error traceback)
- Blackhole hardware (`bh-qbae-15` / `sterling-all` container)
- `layout=ttnn.TILE_LAYOUT`, tested with both `DRAM_MEMORY_CONFIG` and `L1_MEMORY_CONFIG`

## Minimal repro

```python
import torch, ttnn, ttl
TILE = 32

@ttl.operation(grid=(1, 1))
def reduce_max_kernel(x, scaler, out):
    x_dfb   = ttl.make_dataflow_buffer_like(x,      shape=(1, 1), block_count=2)
    sc_dfb  = ttl.make_dataflow_buffer_like(scaler, shape=(1, 1), block_count=1)
    red_dfb = ttl.make_dataflow_buffer_like(x,      shape=(1, 1), block_count=2)
    out_dfb = ttl.make_dataflow_buffer_like(out,    shape=(1, 1), block_count=2)

    @ttl.compute()
    def compute():
        sc = sc_dfb.wait(); xb = x_dfb.wait()
        red_dfb.reserve().store(ttl.math.reduce_max(xb, sc, dims=[1]))
        mb = out_dfb.reserve()
        mb.store(ttl.math.broadcast(red_dfb.wait(), mb, dims=[1]))

    @ttl.datamovement()
    def dm_read():
        ttl.copy(scaler[0, 0], sc_dfb.reserve()).wait()
        ttl.copy(x[0, 0],      x_dfb.reserve()).wait()

    @ttl.datamovement()
    def dm_write():
        ttl.copy(out_dfb.wait(), out[0, 0]).wait()

# Input: row r filled with value r -> row max should be r.
[...]

PR #534, opened 2026-04-26
[fix][ttl] Work around llk all-0 result for fp32 row reduction on Blackhole (#533)

brnorris03:
Problem: Row-reduce with full_fp32 and dtype=f32 produces 0 on BH

Current approach:
- Disabled full_fp32 lowering for REDUCE_ROW while issue #533 is open.
- Avoided auto-enabling fp32_dest_acc_en for bf16 ROW reduce in TTLSetComputeKernelConfig on Blackhole only (which required adding and propagating device info to the compiler).
- Preserved full_fp32 on generated reduce_uninit when the preceding reduce uses full-fp32.
- Expanded `test/python/test_reduce.py` so every reduce pytest runs both bf16 and fp32.
- Added MLIR coverage for:
  - ROW reduce staying non-full-fp32.
  - non-ROW reduce still enabling full-fp32 behavior.
  - reduce-full-fp32=0.
  - reduce_uninit() {full_fp32}.


.github/workflows/call-test-hardware.yml
```diff
@@ -104,25 +104,25 @@ jobs:
...
         run: |
           source build/env/activate
           python3 -m pytest test/python -v --tb=long --junitxml=build/test/pytest-report.xml
 
-      - name: Simulator tests
+      - name: ME2E tests (MLIR input end-to-end tests)
         run: |
           source build/env/activate
-          python3 -m pytest test/sim -v -n auto --tb=short --junitxml=build/test/sim-report.xml
+          python3 -m pytest test/me2e -v --tb=short --junitxml=build/test/me2e-report.xml
```
brnorris03, line 125: Reordered them based on how long tests take and which are most likely to fail first.

third-party/patches/ttmlir-reduce-deinit-arg.patch
brnorris03, line 1: will create a tt-mlir PR and remove this patch after it merges.

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
```diff
@@ -688,7 +688,13 @@ struct TTLTileReduceToTTKernel : OpConversionPattern<TileReduceOp> {
         scalerIdx, setup->dstIdx,
         ttk::ReduceTypeAttr::get(op.getContext(), ttkReduceType),
         ttk::ReduceDimAttr::get(op.getContext(), op.getReduceDim()));
-    if (fullFp32) {
+
+    bool useFullFp32 = fullFp32;
+    // TODO(#533): Blackhole REDUCE_ROW full-fp32 produces incorrect results.
+    if (isBlackholeTarget(op) && op.getReduceDim() == ttk::ReduceDim::Row) {
+      useFullFp32 = false;
```
zoecarver: I don't love this work around because we lose precision... I think it's an improvement on current behavior, but what about using matmul instead? Apparently it's faster too?

brnorris03: It's an llk bug (99% sure, will file it anyway), so this is kind of out of our control specifically. I agree about matmul support, but that's not a bug fix for reduce, I think it's a separate option -- and yes, I agree we should add it.

zoecarver: Yeah sorry if my original message wasn't clear, I think this is an OK bandaid but we should file an issue/fix this long term (with matmul if that's the best we can do *sigh*)

zoecarver: I wonder if we shouldn't emit a warning/error though, I can imagine it being extremely frustrating if you are debugging pcc and don't know why the flag doesn't work on specific hw.

brnorris03: I added a warning (and test), don't think a human can see it easily, but claude and company will for sure

zoecarver (approved):

PR #535, opened 2026-04-27
Simulator: Implict push pop

FConstantinos:
- Adds ability to do implicit pushes, pops and copy waits
- Added ast analysis
- Language patterns that can't be analyzed for implicit copy waits are rejected
- Compiler tests that are relevant to implicit push/pop/copy waits run with the simulator as well


python/sim/auto_push_pop.py
```diff
@@ -0,0 +1,1244 @@
...
+            and isinstance(node.value, ast.Call)
+        ):
+            continue
+        call = node.value
+        func = call.func
+        is_copy = (
+            _is_ttl_copy_call(call)
+            or (isinstance(func, ast.Name) and "copy" in func.id.lower())
+            or (isinstance(func, ast.Attribute) and "copy" in func.attr.lower())
+        )
```
brnorris03: Shouldn't it match `ttl.copy` only (this may match other unrelated copy methods)?

python/sim/program.py
```diff
@@ -216,57 +222,97 @@ def _run_cooperative(
...
-                # Emit operation_end for each node now that all kernels completed.
-                for core in range(total_cores):
-                    trace("operation_end", node=core)
-
-                # Validate all DataflowBuffers have no pending blocks
-                self._validate_dataflow_buffers(all_core_contexts)
-            finally:
-                # Clear scheduler
-                set_scheduler(None)
+            # Analyse all three thread functions once before iterating over
```
brnorris03: Why is the new code not wrapped in try: finally?

python/sim/auto_push_pop.py
```diff
@@ -0,0 +1,1244 @@
...
+        entry = get_context().active_hooks.get(code)
+        if entry is None:
+            return None
+        _, return_ips = entry
+        if return_ips:
+            frame = sys._getframe(1)
+            for ip in return_ips:
+                _fire_injection(frame, ip)
+    except Exception:
+        pass
```
brnorris03: Same question as the above - what are the implications of ignoring the exception?

python/sim/auto_push_pop.py
```diff
@@ -0,0 +1,1244 @@
...
+        if entry is None:
+            return None
+        by_lineno, _ = entry
+        ips = by_lineno.get(line_number)
+        if ips:
+            frame = sys._getframe(1)
+            for ip in ips:
+                _fire_injection(frame, ip)
+    except Exception:
+        pass
```
brnorris03: What are the implications of ignoring the exception here?

python/sim/auto_push_pop.py
brnorris03, line 1: Some Claude review comments about this file:
- [high] Bare except Exception: pass in sys.monitoring 
  callbacks (auto_push_pop.py:1108-1122, 1134-1147). Any
  error inside _fire_injection — including the AccessState
  checks raising on a real bug — is silently dropped. The
  simulator continues with an inconsistent block-state
  machine and produces meaningless errors downstream. Either
  let the exception propagate (Python's monitoring tolerates
  this) or capture it on the context and re-raise once user
  code returns. As is, the analysis is undebuggable when it
  fails.
  - [high] tx.wait() is matched anywhere in the function
  (auto_push_pop.py:265-274). _find_copy_records disqualifies
   a copy auto-wait if any tx.wait() is seen, with no
  ordering check. Pattern tx.wait(); tx = ttl.copy(...)
  (e.g., wait at the top of a loop iteration matching the
  previous iteration's copy) silently disables the auto-wait,
   leaving the second copy un-awaited. Fix: require the
  matching wait to follow the assignment textually. Negative
  test:
  def dm():
      tx.wait()                # noqa: F821 — stale wait
  above
      tx = ttl.copy(a, b)      # this copy is now never
  awaited
  - [high] Loop-acquire return-trigger may double-release
  (auto_push_pop.py:880-911). When an in-loop acquire is
  kept, an unconditional trigger_on_return=True cleanup is
  appended for every acquire in the loop body — not just the
  last — so two reserves on the same DFB inside one for body
  register two return-side triggers and double-push. Fix:
  emit the return trigger only for the last acquire on a
  given DFB.
  - [high] Aliasing of acquire results is mishandled
  (auto_push_pop.py:178-180, 627-655). _is_explicit_release
  only matches <recorded_var>.push/pop. After blk =
  dfb.reserve(); other = blk; other.push(), the analysis
  fires its own blk.push() on top of the user's
  already-released block. Fix: either reject any rebinding of
   acquire results in the validator or perform proper
  liveness through assignments.
  - [high] with dfb.reserve(): (no as) silently passes
  validation (auto_push_pop.py:401-407).
  _find_allowed_dfb_acquire_ids accepts the with form
  regardless of an as clause; _find_acquire_records ignores
  with entirely; the runtime context manager is what does the
   push/pop. So with dfb.reserve(): (no target) is "allowed"
  but the body has no name to release through, producing
  confusing errors if the user assumed they captured the
  block. Fix: require the as <name> form in the validator.
  - [high] Loop-body detection by line number
  (auto_push_pop.py:567-589). _compute_loop_body_lines
  returns a set[int] of line numbers, then
  _find_acquire_records:652 checks stmt.lineno in
  loop_body_lines. Multi-statement lines (stmt1; stmt2) and
  any future code generator that shares lines across loop and
   non-loop statements break this. Fix: key by id(stmt) (node
   identity).

brnorris03 (commented):
Some initial comments by Claude and me. I am still going through the rest.

python/sim/auto_push_pop.py
```diff
@@ -0,0 +1,1244 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Automatic push/pop and copy-wait insertion for simulator thread functions.
+
+When a thread function acquires a block via ``dfb.reserve()`` or
+``dfb.wait()`` without pairing it with an explicit ``blk.push()`` /
```
brnorris03: nit: why `` for the code references? I think the usual markdown is `dfb.reserve()`

test/sim/test_analysis.py
brnorris03, line 1: Can you also test with kernels with more complex control flow (e.g., nested for, if inside for, etc)? I think it would be good to include tests similar to those in #536 .

python/sim/auto_push_pop.py
```diff
@@ -0,0 +1,1244 @@
...
+    return (
+        func.attr in method_set
+        and isinstance(func.value, ast.Name)
+        and func.value.id in name_set
+    )
+
+
+def _is_explicit_release(stmt: ast.stmt, var_name: str) -> bool:
+    """Return True if ``stmt`` is ``<var_name>.push()`` or ``<var_name>.pop()``."""
+    return _is_method_call_on_any(stmt, var_name, {"push", "pop"})
```
brnorris03: What if there is a push for this block after assignment to a different var?
```
blk = dfb.reserve()                                                               
alias = blk                                     
alias.push()
```

brnorris03 (commented):
Some more test ideas by Claude -- some may not be supportable, but then maybe it would be possible to emit errors as part of the AST validation. The compiler actually could probably 

1. Conditional explicit release does not suppress injection
   ```python
   blk = dfb.reserve()
   if cond:
       blk.push()
   ```
The presence of `blk.push()` anywhere suppresses auto-injection; `cond=False` leaves the reserve un-pushed at runtime.

2. Conditional acquire in if/else arms
   ```python
   if cond:
       blk = a.reserve()
   else:
       blk = b.reserve()
   blk.store(...)
   ```
Both branches produce acquire records; one DFB gets an injected push with no matching reserve at runtime.

3. `break`/`continue`/early `return` between acquire and release
   ```python
   for i in range(N):
       blk = dfb.reserve()
       if done: break
       blk.push()
   ```
BN: Not sure about that one; I am pretty sure the compiler would not allow the break.

4. Nested `for` with independent acquires in inner vs outer body — verify trigger placement does not cross the inner-loop boundary.

5. Loop `else` clause containing an acquire (right now marked `in_loop`, but `for/else` runs once after the loop, not per iteration).

6. `tx`-name collision: `tx = dfb.reserve(); ...; tx = ttl.copy(...)` — copy-handle and acquire records share the name.

BN: I was also wondering about name collision cases, e.g., same name variable shadowing but it seemed to me that it should be able to handle those correctly since they are in different scopes -- but worth checking.

python/sim/program.py
FConstantinos: It was unnecessary, it was there only for moving from one pytest to the next.

python/sim/auto_push_pop.py
FConstantinos: The code no longer ignores exceptions

python/sim/auto_push_pop.py
FConstantinos: The code no longer ignores exceptions

python/sim/auto_push_pop.py
FConstantinos: I think now that pushes/pops happen dynamically, these comments are no longer relevant

python/sim/auto_push_pop.py
FConstantinos: I think _is_ttl_copy_call will now match exactly ttl.copy

FConstantinos:
> Some more test ideas by Claude -- some may not be supportable, but then maybe it would be possible to emit errors as part of the AST validation. The compiler actually could probably
> 
> 1. Conditional explicit release does not suppress injection
>    ```python
>    blk = dfb.reserve()
>    if cond:
>        blk.push()
>    ```
> 
> The presence of `blk.push()` anywhere suppresses auto-injection; `cond=False` leaves the reserve un-pushed at runtime.
> 
> 2. Conditional acquire in if/else arms
>    ```python
>    if cond:
>        blk = a.reserve()
>    else:
>        blk = b.reserve()
>    blk.store(...)
>    ```
> 
> Both branches produce acquire records; one DFB gets an injected push with no matching reserve at runtime.
> 
> 3. `break`/`continue`/early `return` between acquire and release
>    ```python
>    for i in range(N):
>        blk = dfb.reserve()
>        if done: break
>        blk.push()
>    ```
> 
> BN: Not sure about that one; I am pretty sure the compiler would not allow the break.
> 
> 4. Nested `for` with independent acquires in inner vs outer body — verify trigger placement does not cross the inner-loop boundary.
> 5. Loop `else` clause containing an acquire (right now marked `in_loop`, but `for/else` runs once after the loop, not per iteration).
> 6. `tx`-name collision: `tx = dfb.reserve(); ...; tx = ttl.copy(...)` — copy-handle and acquire records share the name.
> 
> BN: I was also wondering about name collision cases, e.g., same name variable shadowing but it seemed to me that it should be able to handle those correctly since they are in different scopes -- but worth checking.

I think 1, 2, 3, 5 are no longer interesting test cases, given that the push/pop happens at next acquisition or at end of scope of the block

4 and 6 have tests now

python/sim/auto_push_pop.py
FConstantinos: This is no longer relevant

python/sim/auto_push_pop.py
FConstantinos: Fixed

test/sim/test_analysis.py
FConstantinos: I've added some tests with more complex control flow, including nested for loops and if inside the for loop. However I'm not sure which tests you refer to for #536

python/sim/program.py
```diff
@@ -216,57 +222,106 @@ def _run_cooperative(
...
+
+            # Run scheduler
+            scheduler.run()
+
+            # Emit operation_end for each node now that all kernels completed.
+            for core in range(total_cores):
+                trace("operation_end", node=core)
+
+            # Validate all DataflowBuffers have no pending blocks
+            self._validate_dataflow_buffers(all_core_contexts)
```
brnorris03, line 325: Suggest adding `set_scheduler(None)` at the end of `__call__` (being extra defensive), because removing the old `finally:` leaves `ctx.scheduler` pointing at a stale (run-completed) scheduler between `Program()` invocations — anything that calls `get_scheduler()` between runs (pytest teardown, fixture assertions) sees the old object instead of `None`.

python/sim/program.py
```diff
@@ -216,57 +222,106 @@ def _run_cooperative(
...
+                        getcurrent()._sim_core = c  # type: ignore[attr-defined]
+                        fn()
+                        # Auto-push/pop any blocks still pending when the thread
+                        # function returns normally (final-iteration cleanup).
+                        # This must not run during exception propagation, so it
+                        # is placed after fn() rather than in a finally block.
+                        for _val in core_ctx.values():
+                            if isinstance(_val, DataflowBuffer):
+                                _val.auto_push_block()
+                                _val.auto_pop_block()
```
brnorris03, line 299: The simulator reports a "simulator bug" error when the user's `fn()` raises an exception -- maybe skip validation (and reporting sim bug) on abnormal exit?

python/sim/analysis.py
```diff
@@ -0,0 +1,567 @@
...
+        if isinstance(node, ast.Expr) and isinstance(node.value, ast.Call):
+            call = node.value
+            if _is_ttl_copy_call(call):
+                allowed.add(id(call))
+            elif (
+                isinstance(call.func, ast.Attribute)
+                and isinstance(call.func.value, ast.Call)
+                and _is_ttl_copy_call(call.func.value)
+            ):
+                allowed.add(id(call.func.value))
```
brnorris03, line 269: The method-chain branch allows any attribute call (e.g., `ttl.copy(src, dst).foo()` would be accepted), should it check for `call.func.attr == "wait"`?

python/sim/analysis.py
```diff
@@ -0,0 +1,567 @@
...
+                _collect(stmt.orelse)
+            elif isinstance(stmt, ast.With):
+                _collect(stmt.body)
+            elif isinstance(stmt, ast.Try):
+                _collect(stmt.body)
+                for handler in stmt.handlers:
+                    _collect(handler.body)
+                _collect(stmt.orelse)
+                _collect(stmt.finalbody)
+            # Do NOT recurse into nested def/class bodies.
```
brnorris03, line 349: A helper defined at module scope (e.g., `def h(a, b): return ttl.copy(a, b)`) called from a thread function bypasses both the validator (only walks the thread function's AST) and the runtime auto-wait — no error is emitted and the copy is never awaited. Should the module-scope case be documented as unsupported, or detected at the `copy()` call site?

python/sim/dfb.py
```diff
@@ -1394,7 +1426,9 @@ def validate_no_pending_blocks(self) -> None:
             nxt = [op.name for op in block.expected_ops]
             nm = _name_phrase_for_error(block)
             errors.append(
-                f"Block{nm} is reserve() acquired at kernel exit; producer kernel needs to push() before exit.\n\n"
+                f"Block{nm} from reserve() was not pushed before kernel exit. "
+                f"This is a simulator bug — auto-push injection should have fired. "
```
brnorris03, line 1430: Same concern as https://github.com/tenstorrent/tt-lang/pull/535#discussion_r3230608229 — the "simulator bug" is not accurate when user's `fn()` failed.

brnorris03 (commented):
Left a few minor comments, looks good in general. If you think they are appropriate, a few more test suggestions related to the comments:

- `ttl.copy(src, dst).foo()` should produce a `PatternViolation` (matches the method-chain question above).
- Module-scope helper that wraps `ttl.copy()` -- confirm it works or detect and produce error
- Maybe overkill: multi-line bare `ttl.copy(\n  src,\n  dst,\n)` as a bare expression — confirm `frame.f_lineno` and `stmt.lineno` agree.
- Two threads share a DFB; one yields with a pending `reserve`, the other's exit cleanup must not push it.

python/sim/analysis.py
FConstantinos: Fixed — added and call.func.attr == "wait" to the condition.

python/sim/program.py
FConstantinos: This is addressed by cleaning up simulator context after each ttl.operation run

FConstantinos:
> Left a few minor comments, looks good in general. If you think they are appropriate, a few more test suggestions related to the comments:
> 
> * `ttl.copy(src, dst).foo()` should produce a `PatternViolation` (matches the method-chain question above).
> * Module-scope helper that wraps `ttl.copy()` -- confirm it works or detect and produce error
> * Maybe overkill: multi-line bare `ttl.copy(\n  src,\n  dst,\n)` as a bare expression — confirm `frame.f_lineno` and `stmt.lineno` agree.
> * Two threads share a DFB; one yields with a pending `reserve`, the other's exit cleanup must not push it.

I've written tests for all these cases

python/sim/analysis.py
FConstantinos: Thanks! I have expanded the analysis to include module level functions, nested functions (think what is defined for if_src/if_dst) and generally whatever can be reached by walking the AST and the calls in it. I've also added tests for that.

python/sim/program.py
FConstantinos: Sorry, I'm not following this, if fn() raises an exception, that propagates (whether it is simulator bug or user error) and this is what we want, correct?

Also, which validation do you refer to? If it's _validate_dataflow_buffers, then this is called after the call to fn(), so if fn throws then we would not reach that validation.

python/sim/dfb.py
FConstantinos: Please see above

python/sim/program.py
brnorris03: So when scheduler.run() is called, that runs the user function -- if that raises an exception (in its thread), what exception is produced exactly when the simulator stops running -- the specific one from the user function or a generic simulator one? If it's the former, then ignore this comment, I thought it may be the latter.

brnorris03 (approved):
LGTM! Thank you!

PR #537, opened 2026-04-28
[fix][ttl] Fix live interval boundary computation (#536)

### LINKED ISSUE #536: Auto-inserted cb_pop_front hoisted past subsequent cb_wait_front in same DM thread
## Summary

When a single datamovement thread does N consecutive `cb.wait()` calls on the same DFB, all auto-inserted `cb_pop_front` calls are hoisted to AFTER the last use. Without intervening pops, `get_read_ptr()` returns the same slot 0 pointer for every wait, so every consumer use sources from slot 0; later slots are correctly produced by compute but never read.

Single core, single compute + single dm_write thread, no pipes, no multi-thread CB sharing. Deterministic.

## Reproducer

```python
import torch
import ttnn
import ttl

TILE = 32


def make_kernel():
    @ttl.operation(grid=(1, 1))
    def repro(out):
        shape = (1, 1)
        shared_cb = ttl.make_dataflow_buffer_like(out, shape=shape, block_count=2)
        out_cb = ttl.make_dataflow_buffer_like(out, shape=shape, block_count=2)

        @ttl.compute()
        def compute():
            with shared_cb.reserve() as v:
                v.store(ttl.math.fill(v, 7.0))
            with shared_cb.wait() as src, out_cb.reserve() as dst:
                dst.store(src)

            with shared_cb.reserve() as v:
                v.store(ttl.math.fill(v, 8.0))
            with shared_cb.wait() as src, out_cb.reserve() as dst:
                dst.store(src)

        @ttl.datamovement()
        def dm_read():
            pass

        @ttl.datamovement()
        def dm_write():
            blk = out_cb.wait()
            ttl.copy(blk, out[0:1, 0:1]).wait()
            # blk.pop()  # workaround: explicit pop here makes it pass
            blk = out_cb.wait()
            ttl.copy(blk, out[1:2, 0:1]).wait()
            # blk.pop()

    return repro


def _to_dev(t, device):


if __name__ == "__main__":
    try:
        k = make_kernel()
        H = 2 * TILE
        out_t = _to_dev(torch.full((H, TILE), -42.0, dtype=torch.bfloat16), device)
        k(out_t)
        out_h = ttnn.to_torch(out_t)
        vals = [out_h[0, 0].item(), out_h[TILE, 0].item()]
        expected = [7.0, 8.0]
        print(f"out={vals}  expected={expected}  "
              f"{'PASS' if vals == expected else 'FAIL (auto cb_pop_front hoisted)'}")
    finally:
        ttnn.close_device(device)
```

Output:
```
out=[7.0, 7.0]  expected=[7.0, 8.0]  FAIL (auto cb_pop_front hoisted)
```

## Generated metal kernel (dm_write)

```cpp
cb_wait_front(out_cb, 1);
ptr_a = get_read_ptr(out_cb);            // slot 0 (L1 base)
noc_async_write_tile(0, accessor, ptr_a); // DRAM[0] <- slot 0  (= 7.0)
noc_async_write_barrier();

cb_wait_front(out_cb, 1);                // no pop yet — front still slot 0
ptr_b = get_read_ptr(out_cb);            // SAME pointer as ptr_a
noc_async_write_tile(1, accessor, ptr_b); // DRAM[1] <- slot 0  (= 7.0)
noc_async_write_barrier();

cb_pop_front(out_cb, 1);  // both pops emitted at the very end
cb_pop_front(out_cb, 1);
```

`compute()`'s codegen is fine — slot 0 receives 7.0 and slot 1 receives 8.0, both visible after the pushes. The bug is purely in the consumer-side scheduling: with no `cb_pop_front` between the two `cb_wait_front`/`get_read_ptr` pairs, the read pointer never advances past slot 0.

## Workaround

Adding explicit `blk.pop()` after each use in `dm_write` makes the reproducer pass — it forces the pop into the IR between the two waits, so the read pointer advances correctly.

## Suggested fix

Auto-insert `cb_pop_front` immediately after the last use of each `cb.wait()` block (or before the next `cb.wait()` on the same DFB), rather than coalescing all pops at the end of the function.

brnorris03:
Bug:
Auto-inserted `cb_pop` / `cb_push` could be placed after a later acquire on the same DFB. In a DM writer with consecutive `cb_wait` calls, this left the read pointer on the first slot and caused both writes to read the same tile.

Changes:
- Bound DFB release matching and live-use analysis by the next same-kind acquire on the same DFB.
- Tightened `ttl-insert-cb-sync` documentation/comments around direct DFB uses.

Tests:
- Added MLIR coverage for consecutive waits/reserves and later explicit release cases.
- Added device pytest coverage for the issue 536 reproducer and the later explicit pop case.

Fixes #536

### DIFF
--- a/docs/development/DFBManagement.md
+++ b/docs/development/DFBManagement.md
@@ -63,23 +63,81 @@ correct DFB interval boundary.
 
 The pass treats every acquire as opening a DFB live interval. The interval
 starts at `cb_reserve` or `cb_wait` and ends after the last operation that can
-use the acquired slot. A later same-kind acquire on the same DFB bounds release
-matching and use discovery, because its release belongs to a different live
-interval.
+use the acquired slot. A later acquire in the same DFB sync class bounds
+release matching and use discovery, because its release belongs to a different
+live interval.
+
+DFB sync classes separate the producer side from the consumer side:
+`cb_reserve`/`cb_push` form producer intervals, and `cb_wait`/`cb_pop` form
+consumer intervals. Producer acquires bound other producer intervals; consumer
+acquires bound other consumer intervals.
 
 The pass finds owned uses from two sources:
 
 - Tensor-form uses follow the result of `cb_reserve` or `cb_wait` through
   `ttl.attach_cb`, `ttl.store`, and compute operations.
 - Direct DFB uses follow `ttl.copy` operations where the DFB operand direction
-  matches the acquire kind. `cb_reserve` live intervals include copies into the
-  DFB; `cb_wait` live intervals include copies from the DFB. This is required
-  for data movement kernels, where copies do not use the tensor value returned
-  by the acquire op.
+  matches the interval's DFB sync class. Producer intervals include copies into
+  the DFB; consumer intervals include copies from the DFB. This is required for
+  data movement kernels, where copies do not use the tensor value returned by
+  the acquire op.
 
 Uses inside descendant regions are projected to their ancestor operation in the
 acquire's block. This conservatively places the release after the enclosing
 structured op when the exact use is nested in an `scf.for` or `scf.if` body.
+
+### Slot State Model
+
+The pass models producer and consumer acquires as separate slot lifetimes:
+
+```
+Producer side:
+
+  free slot
+      |
+      | cb_reserve
+      v
+  reserved slot
+      |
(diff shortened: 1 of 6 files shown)


zoecarver (approved):
Thank you!!

PR #538, opened 2026-04-28
Simulator: Better error messages

FConstantinos:
This adds more descriptive error messages for the block state machine


python/sim/dfb.py
```diff
@@ -1310,33 +1391,37 @@ def validate_no_pending_blocks(self) -> None:
 
         if self._pending_reserved_block is not None:
             block = self._pending_reserved_block
+            nxt = [op.name for op in block.expected_ops]
+            nm = _name_phrase_for_error(block)
             errors.append(
-                f"Pending reserved block: Block(acquisition={block.acquisition.name}, "
-                f"thread={block.thread_type.name}, access={block.access_state.name}, "
```
phizalev-TT, line 1315: I think we should use "kernel" instead of "thread" in messages (and eventually in the code).

phizalev-TT (changes requested):

python/sim/dfb.py
FConstantinos: Sorry, I thought I'd already done that, but seems I missed some spots!

Fixed this for error messages, but for the code itself I'll do it in another commit to avoid a large breaking change

examples/errors/copy_source_lock_error.py
```diff
@@ -0,0 +1,128 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
brnorris03, line 1: I get this, too, and not sure whether we are supposed to keep using 2025 or for new files use 2026? Do you know?

FConstantinos: I was thinking we should just update all files to write 2025-2026 and do that update once a year. I guess the LLM tries to copy what the other files do and it never updates itself? :P

phizalev-TT (approved):

PR #540, opened 2026-04-29
[fix][ttl] Materialize tensor loop state before compute lowering (#527)

### LINKED ISSUE #527: [external bug report][from claude] ttl.add inside scf.for is dropped when result is not carried out of the loop
## Summary

Inside a `scf.for` loop, a Python-level accumulator reassignment like
`acc = acc + recv_cb.wait()` is silently dropped: the `ttl.add` op is
generated but its SSA result is unused outside the loop, so the store
after the loop writes the pre-loop `acc` instead of the accumulated sum.

The equivalent single-statement expression works correctly.

## Reproducer

Single core, no pipes, 32x32 tile. All data read from DRAM.

Expected `out[0, 0] == 34.0` (matmul partial `= 32` + recv value `= 2`).
Observed `out[0, 0] == 32.0` — the recv contribution is dropped.

```python

import torch
import ttnn
import ttl

TILE = 32


def make_kernel():
    @ttl.operation(grid=(1, 1))
    def kern(a, w, v, out):
        a_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
        b_cb = ttl.make_dataflow_buffer_like(w, shape=(1, 1), block_count=2)
        partial_cb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
        recv_cb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)

        @ttl.compute()
        def compute():
            # partial_cb <- matmul-accumulate (one iter)
            p = partial_cb.reserve()
            a_blk = a_cb.wait()
            b_blk = b_cb.wait()
            p += a_blk @ b_blk
            a_blk.pop()
            b_blk.pop()
            p.push()

            # BUG: this loop body's add result is dropped.
            acc = partial_cb.wait()
            for _ in range(1):
                acc = acc + recv_cb.wait()
            out_cb.reserve().store(acc)

            # WORKS: out_cb.reserve().store(partial_cb.wait() + recv_cb.wait())

        @ttl.datamovement()
        def dm_read():
            with a_cb.reserve() as a_blk:
                ttl.copy(a[0:1, 0:1], a_blk).wait()
            with b_cb.reserve() as b_blk:
                ttl.copy(w[0:1, 0:1], b_blk).wait()
            with recv_cb.reserve() as r:
                ttl.copy(v[0:1, 0:1], r).wait()

        @ttl.datamovement()
        def dm_write():
            with out_cb.wait() as o:
                ttl.copy(o, out[0:1, 0:1]).wait()

    return kern


def to_dev(t, device):
    return ttnn.from_torch(t.contiguous(), dtype=ttnn.bfloat16,
                           layout=ttnn.TILE_LAYOUT, device=device,
                           memory_config=ttnn.DRAM_MEMORY_CONFIG)


if __name__ == "__main__":
    try:
        a_torch = torch.ones(TILE, TILE, dtype=torch.bfloat16)
        w_torch = torch.ones(TILE, TILE, dtype=torch.bfloat16)
        v_torch = torch.full((TILE, TILE), 2.0, dtype=torch.bfloat16)

        k = make_kernel()
        a_tt = to_dev(a_torch, device)
        w_tt = to_dev(w_torch, device)
        v_tt = to_dev(v_torch, device)
        out_tt = to_dev(torch.zeros(TILE, TILE, dtype=torch.bfloat16), device)
        k(a_tt, w_tt, v_tt, out_tt)

        val = ttnn.to_torch(out_tt).float()[0, 0].item()
        print(f"out[0,0] = {val:.1f}  expected 34.0")
    finally:
        ttnn.close_device(device)
```

## Broken vs. working pattern

**Broken** — add is dropped:

```python
acc = partial_cb.wait()
for _ in range(1):
    acc = acc + recv_cb.wait()
out_cb.reserve().store(acc)
```

**Works** — single expression:

```python
out_cb.reserve().store(partial_cb.wait() + recv_cb.wait())
```

Both should be equivalent for `range(1)`.

## Output

```
out[0,0] = 32.0  expected 34.0
```

## Initial MLIR (relevant excerpt)

The `ttl.add` is emitted inside the `scf.for` body, but the result `%19`
has no users, and the `ttl.store` after the loop uses `%13` (the pre-loop
`partial_cb.wait()`):

```mlir
%12 = ttl.cb_wait %3 : <[1, 1], !ttcore.tile<32x32, bf16>, 2> -> tensor<1x1x!ttcore.tile<32x32, bf16>>
%13 = ttl.attach_cb %12, %3 : (...) -> tensor<1x1x!ttcore.tile<32x32, bf16>>
%c1_i64 = arith.constant 1 : i64
%c1 = arith.constant 1 : index
%14 = arith.index_cast %c1_i64 : i64 to index
scf.for %arg0 = %c0 to %14 step %c1 {
  %17 = ttl.cb_wait %4 : <[1, 1], !ttcore.tile<32x32, bf16>, 2> -> tensor<1x1x!ttcore.tile<32x32, bf16>>
  %18 = ttl.attach_cb %17, %4 : (...) -> tensor<1x1x!ttcore.tile<32x32, bf16>>
  %19 = ttl.add %13, %18 : ..., ... -> tensor<1x1x!ttcore.tile<32x32, bf16>>   // <-- UNUSED
}
%15 = ttl.cb_reserve %2 : <[1, 1], !ttcore.tile<32x32, bf16>, 2> -> tensor<1x1x!ttcore.tile<32x32, bf16>>
%16 = ttl.attach_cb %15, %2 : (...) -> tensor<1x1x!ttcore.tile<32x32, bf16>>
ttl.store %13, %15 : ..., ...                                                  // <-- stores pre-loop value
```

## Final MLIR (codegen)

No `add` op in the generated compute kernel — just a `copy_tile` of
`partial_cb` straight into the output CB. `recv_cb` is waited and popped
but its data is discarded:

```
cb_wait_front(partial_cb)
cb_wait_front(recv_cb)
cb_pop_front(recv_cb)            // consumed but never read
cb_reserve_back(out_cb)
copy_tile_init(partial_cb)
copy_tile(partial_cb, ...)       // <-- only partial; no add
pack_tile(..., out_cb, ...)
cb_pop_front(partial_cb)
cb_push_back(out_cb)
```

## Root cause (hypothesis)

The Python tracer treats `acc = acc + recv_cb.wait()` inside the loop as
a local rebind that does not escape the loop body. After the loop, `acc`
still refers to the pre-loop value (`%13`). The add op is emitted as a
side-effect but its SSA result is unused and DCE'd, while the store
after the loop uses the stale pre-loop value.

## Context

- Tested on Blackhole hardware (via `run-test.sh --hw`)
- TTL DSL
- Observed originally in a k-split matmul where the same pattern is used
  to accumulate `K_parts - 1` received partials; caused PCC ≈ 0.706 (only
  root's own partial landing).

brnorris03:
### Problem description

Python kernels can express tensor recurrences by rebinding a tensor variable to an expression that reads its previous value:

```python
acc = acc + x
```

The value after the loop must be the final recurrence value. Before this change, the compiler did not consistently represent and lower that tensor state through control flow, so compute lowering could not consume loops that produced tensor results.

Additive recurrences also have a performance requirement. They should preserve the L1 accumulation strategy that `+=` already uses on a reserved block, instead of storing every intermediate accumulator value in a compiler-allocated DFB. After this PR, `acc = acc + x` and `out_blk += x` lower to the same in-loop accumulating `ttl.store`; `+=` is now a special case of self-rebinding.

### What's changed
- Added `ttl-materialize-loop-state`, a TTL transform that removes ranked-tensor `scf.for` iter args before compute lowering. Additive recurrences of the form `ttl.add(%iter_arg, %contribution)` (and the commuted form) lower to a pre-loop initial store and an in-loop accumulating `ttl.store`; non-additive recurrences lower through a compiler-allocated DFB state slot.
- Extended Python AST lowering to emit `scf.for` iter args for self-rebinding loop variables (including tuple targets like `a, b = a + d, b + d`) and resultful `scf.if` for tensor variables assigned in branches.
- Factored shared compiler-allocated DFB materialization helpers so `ttl-insert-intermediate-dfbs` and `ttl-materialize-loop-state` use the same allocation convention.
- Fixed `TTKernelInsertL1Accumulation::precededByNonAccumulatingPack` so a `cb_reserve_back` or `cb_push_back` for a CB already covered by a closer pack no longer aborts the scan. Loops that accumulate into multiple CBs now detect their pre-loop init packs and reconfig L1 acc as "enable" before the loop.
- Lit coverage for additive, commuted additive, unary and binary recurrences, mixed tensor states, scalar iter args, multiple final-result users, multi-use add fallback, conditional recurrence, and zero-trip loop semantics. Device coverage (dtype-parametrized bf16/f32) for additive, non-additive, tuple-target, three-accumulator, multi-tile-block, and zero-trip recurrences.
- Conditional tensor rebinding inside a loop is correct at the materialize-loop-state level but is left as `xfail` end-to-end: `ttl-assign-dst` does not yet descend into nested regions, tracked as #587.

Fixes #527


lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
brnorris03, line 1: Code just moved into this common file from `TTLInsertIntermediateDFBs.cpp`, no changes

brnorris03:
Superseded by the multiple PRs split from the [bnorris/accumulation-scope-redesign](https://github.com/tenstorrent/tt-lang/pull/651) branch.

PR #543, opened 2026-04-29
Add input type verification

vtsilytskyiTT:
### Problem description
If user provides tensors with incompatible data types (eltwise operation with bf16 and fp32) tt-lang produces a misleading error message.
Before change the message is:
`ttl.diagnostics.TTLangCompileError: error: shape mismatch between (1, 1) f32 tensor and (1, 1) bf16 tensor; note: you can use ttl.math.broadcast() to expand the smaller tensor`

After change:
`ttl.diagnostics.TTLangCompileError: error: binary operation requires matching tensor operand types; got lhs type tensor<1x1x!ttcore.tile<32x32, bf16>> and rhs type tensor<1x1x!ttcore.tile<32x32, f32>>`

### What's changed
Discover situation with incompatible tensor types and report it clearly.

### Ticket
None

### Checklist


python/pykernel/_src/utils.py
```diff
@@ -14,18 +14,59 @@
...
+    if not match:
+        return None
+
+    shape_str, dtype = match.groups()
+    try:
+        shape = tuple(int(dim) for dim in shape_str.split("x"))
+    except ValueError:
+        return None
+    return shape, dtype
+
```
brnorris03: there are MLIR python type accessors .shape and .element_type already, there is no need for using regex here.

brnorris03 (commented):
C++ MLIR verifiers already enforce the same invariant for matmul and store; copy has a documented TODO that should be implemented, rather than patched around in Python. I think it would achieve the desired effect of better errors without duplicating functionality and possible inconsistencies between the python frontend and the rest of the compiler.

brnorris03:
For elementwise, there are a couple of options -- you could add a verify to `TTL_ElementWiseBinary` or convert `TTL_BinaryElementwiseOpTrait` into a real verifying trait similar to the upstream `SameOperandsAndResultElementType`/`SameTypeOperands`.

The easiest is to just add the `SameTypeOperands` trait (upstream MLIR) to                `TTL_ElementWiseBinary`, but I have some vague memory of it not working in our case (too constraining) -- still should be the first thing to try.

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -197,37 +197,60 @@ mlir::LogicalResult mlir::tt::ttl::CopyOp::verify() {
...
+                           << ")";
+    }
+  }
+
+  if (transferTensorTy.getElementType() != cbTy.getElementType()) {
+    return emitOpError() << "tensor element type ("
+                         << transferTensorTy.getElementType()
+                         << ") must match CB element type ("
+                         << cbTy.getElementType() << ")";
+  }
```
brnorris03, line 253: Nice!

python/pykernel/_src/kernel_ast.py
```diff
@@ -528,7 +536,19 @@ def visit_BinOp(self, node):
...
+        )
+        if same_shape_tensor_type_mismatch:
+            message = (
+                "binary operation requires matching tensor operand types; "
+                f"got lhs type {lhs.type} and rhs type {rhs.type}"
+            )
+            if hasattr(self, "_raise_error"):
+                self._raise_error(node, message)
+            raise TypeError(message)
+        if types_differ:
```
brnorris03: Is this still needed? Shouldn't the verifiers catch this now? I am guessing something is sneaking through -- do you have an example where the MLIR verifiers are allowing this (if you didn't have this check here)?

brnorris03 (commented):
Thanks for addressing the comments, just had one more question.

lib/Dialect/TTL/IR/TTLOps.cpp
vtsilytskyiTT: There was a TODO, now there is the validation logic.

python/pykernel/_src/kernel_ast.py
vtsilytskyiTT: Agree, as I can see this code snippet can be safely removed.

brnorris03 (approved):
lgtm, thank you!!

Issue #541, opened 2026-04-29
[claude] grid="auto" produces incorrect output when work shape does not exactly fill the device grid


brnorris03:
### Problem description

`@ttl.operation(grid="auto")` deterministically produces incorrect output whenever the work (M_BLOCKS x N_BLOCKS) does not exactly fill the device's compute grid. The kernel runs on every core of the auto-allocated grid, including cores that have no valid block to process. Those "idle" cores still execute the dm_read / dm_write / compute loops with out-of-bounds block indices, which corrupts the multicast handshake (and likely also corrupts adjacent DRAM via OOB writes).

This appears to be the underlying cause of:
- #513 -- `test_balanced_matmul` reported as flaky (output values such as `mean = -1.4e29`, consistent with uninitialized memory).
- #502 -- `singlecore_matmul.py` and `multicore_matmul.py` fail on hardware (both use `grid="auto"`).
- All three tests in `test/python/pipe/test_mcast_matmul.py` (`test_mcast_matmul`, `test_balanced_matmul`, `test_balanced_matmul_relu`) failing on hardware -- not only `test_balanced_matmul`.

The reported flakiness is consistent with OOB DRAM writes landing on different addresses across runs. On the wormhole_b0 setup tested here the failure is fully deterministic.

### Reproduction

Reproduced on wormhole_b0 (8x7 compute grid) and on blackhole (different grid; see below). Run from a built tree:

```bash
source build/env/activate
pytest test/python/pipe/test_mcast_matmul.py -v
```

All three tests fail in ~24s on wormhole. The minimal repro is M = K = N = 256:

```python
import torch, ttl, ttnn
from ttlang_test_utils import to_dram

TILE = 32
BLOCK_M = BLOCK_N = BLOCK_K = 8
BLOCK_SIZE = BLOCK_M * TILE  # 256

def make_kernel(M, K, N, grid):
    M_BLOCKS = M // BLOCK_SIZE
    N_BLOCKS = N // BLOCK_SIZE
    K_BLOCKS = K // BLOCK_SIZE

    @ttl.operation(grid=grid)
    def mm(a, w, out):
        NUM_COLS, NUM_ROWS = ttl.grid_size(dims=2)
        # ... same body as test_mcast_matmul.make_mcast_kernel ...
    return mm

device = ttnn.open_device(device_id=0)
a = to_dram(torch.randn(256, 256, dtype=torch.bfloat16) * 0.02, device)
[...]

PR #544, opened 2026-04-29
[fix][ttl] Fix incorrect results with `grid=auto` when work does not fill node grid (#541)

brnorris03:
Fixes issue #541 by migrating multicast pipe kernels to host-side callable grids and rejecting `grid="auto"` with multicast pipes at compile time.

Updates tests, docs, and skills that use non-portable hardcoded grid sizes, e.g., `(8,8)`.

Future (planned) work: the compile-time check in this PR rejects `grid="auto"` combined with multicast pipes but does not prevent the same problem when the grid is supplied as a callable returning the full device extent. The current simple syntactic check should be replaced with a real semantic verifier that runs after grid resolution on the traced MLIR and enforces three invariants: tensor index expressions remain within tensor shape on every active core; each pipe send/receive executes only on cores in the pipe's source/destination set; and no active core executes kernel code without membership in every pipe referenced on its execution path. Kernels violating any invariant should be rejected with diagnostics identifying the violated invariant, the offending cores, and a suggested correction.


claude-slash-commands/ttl-import.md
```diff
@@ -429,48 +429,63 @@ For tensors larger than 32x32, process multiple tiles. **Use multinode and loops
 
 ### IMPORTANT: Match the User's Target Data Size
```
zoecarver, line 430: Thanks for updating this, unfortunately these skills moved to tt-forge repo. I need to delete these from this repo.

python/sim/ttnnsim.py
```diff
@@ -911,9 +911,17 @@ def compute_with_storage_grid_size(self) -> CoreCoord:
...
 
+_DEFAULT_DEVICE = Device()
+
+
 def open_device(device_id: int = 0) -> Device:
-    """Open a simulated device (no-op)."""
-    return Device(device_id)
+    """Open a simulated device (no-op).
+
+    Sim has no real per-device state, so all calls share the module-level
```
FConstantinos, line 920: If we do that, and there is user code that checks for equality between two devices, would that break their code, or introduce a difference on how their code behaves between simulator and compiler?

brnorris03: I am not sure, and I don't really like the dummy device idea (at least not this very simplistic one). There were other things I don't like about this partial fix, so I implemented a more general pipenet approach in https://github.com/tenstorrent/tt-lang/pull/547 (which eliminates the need for this). Can you please take a look (maybe starting by reading the [doc](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipenet-refactor/docs/development/PipeNets.md))?

examples/metal_examples/1d_mcast_matmul/ttlang/1d_tt_lang.py
```diff
@@ -10,7 +10,14 @@
 TS = ttnn.TILE_SIZE  # 32
 
 
-@ttl.operation(grid="auto")
+def _device_grid_2d(*tensors):
+    # Full device extent. The kernel uses per-node guards to handle inactive
+    # nodes (see `if node_index < num_working_nodes`).
+    g = tensors[0].device().compute_with_storage_grid_size()
```
zoecarver, line 16: Can we use the ttl grid api instead?

brnorris03:
PR #447 would supersede this PR.

Issue #505, opened 2026-04-14
Support overlapping multicast destinations in PipeNet


zoecarver:
Currently, multicast pipes in a PipeNet that share destination cores error at construction time (see `PipeNet._validate_no_overlapping_destinations` in `python/ttl/pipe.py`). This blocks patterns like scatter-gather/all-to-all where every core multicasts to every other core.

The fix: switch the multicast receiver signaling from `noc_semaphore_set_multicast` (overwrite) to `noc_semaphore_inc_multicast` (atomic increment), matching how unicast gather already works with cumulative `semaphore_wait_min`. The `noc_semaphore_inc_multicast` API exists in tt-metal (`dataflow_api.h:2290`) but needs to be added to the TTKernel dialect first.

PR #547, opened 2026-05-02
[ttl] Pipenet enhancements, fix #541

### LINKED ISSUE #541: [claude] grid="auto" produces incorrect output when work shape does not exactly fill the device grid

brnorris03:
### Problem

Issue #541. When the launch grid is larger than a PipeNet's work extent, unguarded PipeNet-coupled work (DFB↔pipe copies, `cb_wait` on a pipe-coupled buffer, `with cb.reserve()` blocks carrying pipe role traffic) runs on launched nodes outside the work extent, indexing tensors out of bounds and breaking the multicast handshake — silent miscompilation.

### Change

The launch grid and the PipeNet active extent are decoupled. Pipes are sized from the work extent; the user wraps PipeNet-coupled regions in `is_src` / `is_dst` / `is_active` guards (or `if_src` / `if_dst` callbacks); a new compiler pass verifies the guards and rejects unguarded work with a precise diagnostic.

The new functionality is mostly in a new pass, `ttl-verify-pipenet-guards` (~990 lines), but there are various other source changes for adding the predicates (total ~2000 lines source changes, the rest are tests and documents).

#### Compiler

`ttl-verify-pipenet-guards` walks each kernel-thread function and proves that each PipeNet-coupled op's effective execution domain ⊆ the role domain it requires:

* `ttl.copy(buffer, pipe)` — reachable only from the pipe's source.
* `ttl.copy(pipe, buffer)` — reachable only from the pipe's destination range.
* `cb_wait` on a pipe-coupled DFB — reachable only from the union of `cb_push` producers for the same DFB index (cross-thread, since indices are module-global).
* `ttl.pipenet_scope` (frontend-emitted region marker) — execution domain ⊆ declared role union.

Domains are `std::set<Coord>` over the launch grid (mandatory `ttl.launch_grid` module attribute). The walker recognizes `scf.if` / `affine.if` / `scf.for` / `scf.while` / `affine.for` / `linalg.generic` / `ttl.if_src` / `ttl.if_dst` / `ttl.pipenet_scope`; conditions decompose via the new `ttl.is_src` / `ttl.is_dst` / `ttl.is_active` predicate ops or per-coord evaluation of `core_x`/`core_y` arithmetic and `affine.if` IntegerSets. Errors accumulate; one `signalPassFailure` at the end. After verification, an inline-and-erase sweep removes the `pipenet_scope` markers before TTL→TTKernel.

The pass is wired into the C++, Python frontend (`ttl_api.py`), and me2e pipelines, after `ttl-annotate-cb-associations` / `ttl-finalize-dfb-indices` and before `convert-ttl-to-ttkernel`.

`CreatePipeOp::verify()` requires `dstStart <= dstEnd` per the spec's [contiguous-hypercube rule](https://github.com/tenstorrent/tt-lang/blob/<spec-commit>/docs/sphinx/specs/TTLangSpecification.md#L633), so destination domains read directly without normalization. Construction-time PipeNet validation (empty, mcast dst overlap, mixed unicast/multicast per the spec's [`PipeNet[DstT]` signature](https://github.com/tenstorrent/tt-lang/blob/<spec-commit>/docs/sphinx/specs/TTLangSpecification.md#L653)) is centralized in `OperationPipeNets.validate()`, so a malformed `PipeNet([...])` errors as early as possible for both compiler and sim.

#### Diagnostics

Every user-facing diagnostic embeds the offending PipeNet id and a concrete suggested fix in the primary message; structured notes attach a witness coordinate, a `PipeNet <N> declared here` note pointing at the user's `ttl.PipeNet([...])` call site, and the suggested guard. The Python frontend formatter renders the primary error and each note as separate source-context blocks. See [the diagnostics table](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipenet-refactor/docs/development/PipeNets.md#user-facing-diagnostics) for the full list.

#### Simulator parity

Both compiler and sim consume one operation-level data structure, `OperationPipeNets` (in `python/_pipenets/__init__.py`), built from the operation closure plus each thread's wrapped function. PipeNet ids are operation-local. The simulator implements `is_src` / `is_dst` / `is_active` and the `if_src` / `if_dst` callbacks with the same per-pipe role semantics the verifier checks against, so a kernel that compiles also runs on the simulator. Captured PipeNets (closure-defined) are discovered by the same closure walk and contribute to role domains identically.

#### Spec

`## Pipe` now states explicitly that node ranges must lie within `grid_size` but need not span it, and pipes should be sized from work extent. `### Pipe net` allows closure-captured PipeNets, and the active-set paragraph requires implementations to verify user guards on PipeNet-coupled regions imply the correct role. The Gather example uses explicit work-extent shape names instead of `ttl.grid_size()`.

See [docs/development/PipeNets.md](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipenet-refactor/docs/development/PipeNets.md) for the verifier algorithm, IR additions, invariants, worked examples, and future work.

### Ticket

Closes #541

### Test coverage

See [Test coverage](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipenet-refactor/docs/development/PipeNets.md#test-coverage) for the per-behavior matrix (Device / Sim / Lit). The same pytest files run on hardware and on the simulator via `test/scripts/ttlang-sim-pytest`. Negative coverage for every diagnostic lives in `test/ttlang/Dialect/TTL/Transforms/verify_pipenet_guards_invalid.mlir` (under `--verify-diagnostics`); the end-to-end Python lit `test/python/invalid/invalid_pipenet_full_grid_no_guard.py` mirrors the issue #541 reproduction and confirms the diagnostic is rendered correctly through the Python frontend.

### Repaired pre-existing tests

`test/python/pipe/test_mcast_matmul.py` was incorrect on n150 hardware and silently passing on Blackhole. Pipes were sized against the launch grid (`NUM_ROWS`, `NUM_COLS`) but work was distributed via `_even_split`, which returns a smaller `num_*_used` extent when the launch does not divide work evenly. Nodes outside the work subgrid wrote out-of-bounds blocks, corrupting DRAM. The kernels now size pipes against `num_rows_used` × `num_cols_used` and wrap thread bodies in explicit `is_active()` guards. Verified on Blackhole; n150 CI should now pass.

The collectives kernels in `test/python/pipe/test_pipenet_collectives.py` similarly now wrap their thread bodies in `is_active()` so launched nodes outside the active set short-circuit instead of stalling on unproduced DFBs.

### Future work

* Issue #505 (within-PipeNet multicast destination overlap): the validator still rejects two multicast pipes targeting an overlapping node within one PipeNet (single shared semaphore pair).
* Full future-work list (parametric PipeNets, interprocedural analysis across `func.call`, cross-chip PipeNets, larger-grid domain representations): see `docs/development/PipeNets.md`.


docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -628,7 +628,9 @@ Blocks have a life cycle that starts with acquisition by using dataflow buffer's
...
+A *pipe* is a communication primitive for organizing the passing of data between data movement kernels on different nodes. A pipe is used as a source or a destination in the `ttl.copy`. The pipe is constructed with source node coordinate (`src`) and destination (`dst`), which is either a single node coordinate for unicast or *node range* for multicast. The node range uses a combination of dimension slices and values to describe a contiguous hypercube.
+
+A node range has the same number of dimensions as `grid_size(dims=N)`, and each coordinate `c_i` lies within the corresponding grid extent `G_i` (i.e. `0 <= c_i < G_i`). The range may be smaller than the grid: pipes are not required to span every node along any dimension. Pipe coordinates should be sized from the operation's *work extent*, which may be smaller than the launch extent given by `@ttl.operation(grid=...)`.
```
FConstantinos: Should we consider defining what "work extent" and "launch extent" mean here, or at least give an example?

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -642,7 +644,9 @@ A *pipe net* is a communication primitive that groups pipes into a network. A pi
...
+A pipe net is constructed either in the scope of an operation function or in an enclosing scope and captured by the operation function. It can only be used with its `if_src` and `if_dst` functions inside of a data movement kernel function. The corresponding  `ttl.copy` where a pipe is a source or a destination can be called only inside of a condition body function. Calls into `if_src` and `if_dst` can be nested within condition functions for different pipe nets.
+
+The *active set* of an operation is the union, over every pipe in every pipe net in scope of the operation (constructed in its body or captured from an enclosing scope), of the pipe's source coordinate and its destination coordinate or range. Nodes outside the active set do not participate in pipe communication and the implementation must ensure they skip the bodies of every kernel thread function for that operation. This decouples pipe extent from launch extent: an operation launched with `grid="auto"` on a larger grid than its work shape executes only on the active subset.
```
FConstantinos: This to me seems to say that if a ttl.operation has pipes defined, then only the active set of nodes defined by those pipes will run the kernels.

Are we worried about the implicitness here? It is not evident in the kernel code itself whether it will run in a particular node or not.

brnorris03: Yeah, I think that's a valid concern. We could alternatively not do anything in the compiler in terms of ensuring that kernel functions only run when active, but detecting when a user has NOT properly guarded their receiver's code is a much harder problem (on the compiler side).

docs/sphinx/specs/TTLangSpecification.md
brnorris03: I think that auto always meaning "full" is at best a bad name. The `test/python/pipe/test_pipenet_active_guard.py` is one of a number of tests that uses auto but does not multicast to ALL nodes in the grid. Yes, user code can be fully defined with conditionals everywhere to always check if a node should be participating in the multicast (based on the size of the data), but there is no way for the compiler right now to detect whether this was actually done, and we end up with incorrect results as in the failing matmul tests in main. 

Apart from whether the compiler should try to prevent some of these errors (missing ifs in user code to ensure no node reads or writes out of bounds as a result of a full-grid multicast), I think also having "auto" always mean "full" grid is misleading at best. To me auto implies that there will be some automatic correctness or optimization done by the compiler (and sim) that would be correct without requiring users to write their kernels in a specific different way.

brnorris03: So for the current "auto", user code would have to (always) look like something like this (subset of matmul example):

```python
  @ttl.operation(grid="auto")                                                                  
  def small_mcast_matmul(a, w, out):                                                           
      ttl.PipeNet([...])                                                                       
      @ttl.compute()                                                                           
      def compute():                               
          x, y = ttl.node(dims=2)                                                              
          if x >= N_BLOCKS or y >= M_BLOCKS:                                                   
              return                                                                           
          ...                                                                                  
      @ttl.datamovement()                                                                      
      def dm_read():                                                                           
          x, y = ttl.node(dims=2)                  
          if x >= N_BLOCKS or y >= M_BLOCKS:                                                   
              return                                                                           
          ...                                      
      @ttl.datamovement()                                                                      
      def dm_write():                             
          x, y = ttl.node(dims=2)                  
          if x >= N_BLOCKS or y >= M_BLOCKS:
              return                                                                           
          ...
```

whereas with compiler-inserted guards, there is no boilerplate x,y checks in each function or part of the code that needs to respect the active set (explicitly).

I think it's fine to decide not to automate the guards (e.g., not to have the new pass in this PR), but then how do you suggest we deal with user code that happens to be called on sizes that end up with less than full grid occupancy?

FConstantinos:
These changes make a simulator function redundant, I've pushed a commit that removes that function.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -628,7 +628,17 @@ Blocks have a life cycle that starts with acquisition by using dataflow buffer's
...
+
+| Value | Launch extent |
+| :---- | :---- |
+| `"auto"` | When the operation uses pipe nets, the per-axis bounding box of every pipe coordinate (the work extent); otherwise the device compute grid. |
+| `"full"` | The device compute grid, regardless of pipe coordinates. |
+| Tuple | Used verbatim. |
+
+Under `grid="full"` (or any explicit tuple wider than the work extent), the user must guard pipe-coupled regions with `net.is_src()` / `net.is_dst()` / `net.is_active()` (or equivalent coordinate predicates) so that nodes outside the corresponding role skip the pipe-coupled work. The implementation must verify these guards statically and reject programs in which a pipe-coupled operation is reachable from a node outside its declared role. Under `grid="auto"`, the launch coincides with the active set and the verification is trivially satisfied.
```
phizalev-TT: Looks like here there is a distiction between "auto" and "full", but my understanding that there is none right now.

If this is true I suggest only documenting "full" for now.

brnorris03: Good catch, I rewrote that to make it clearer that they are the same.

docs/sphinx/specs/TTLangSpecification.md
brnorris03: I rewrote the definitions, can you take another look? Also auto is the same as full now.

docs/sphinx/specs/TTLangSpecification.md
brnorris03: Sorry for misunderstanding the request in my earlier responses. I added some (hopefully) clearer definitions to that part of the spec, let me know if it's still not sufficient.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -642,13 +658,18 @@ A *pipe net* is a communication primitive that groups pipes into a network. A pi
...
+
+The *active set* of an operation is the union, over every pipe in every pipe net in scope of the operation (constructed in its body or captured from an enclosing scope), of the pipe's source coordinate and its destination coordinate or range. The *role domain* of a pipe net is its per-net active set; `pipe_net.is_src()`, `pipe_net.is_dst()`, and `pipe_net.is_active()` are predicates that evaluate to `True` on the source role, destination role, and their union, respectively. Implementations must verify that user-written guards on pipe-coupled regions imply the correct pipe net role: a `ttl.copy` with a pipe source must be reachable only from the pipe's source coordinate; a `ttl.copy` with a pipe destination must be reachable only from the pipe's destination range; a dataflow buffer wait paired with a pipe-coupled push must be reachable only from a node where some producer pushes. Implementations are not required to skip non-pipe-coupled work outside the active set: under `grid="full"` users may legitimately run plain SPMD work over the full launch grid alongside pipe-coupled regions guarded by `is_src` / `is_dst` / `is_active`.
```
phizalev-TT: Is *active set* same as *work extent*? If so we may want to use *active extent* to clarify.

phizalev-TT: I am not sure I understand why we use "plain SPMD work" here. SPMD is concerned with multi-device operations.

brnorris03: The generic SPMD term is about any program running in parallel -- whether all parallel threads of control execute the same program or not (when using is_src, etc, they do not). I can try rewording this.

brnorris03: > Is active set same as work extent? If so we may want to use active extent to clarify.

They have different purposes. But I think perhaps not both are needed in the spec, just active set, so I will remove work extent since it mainly serves the implementation.
                                                                                          
Active set is the exact set of pipe-role nodes — the union of each pipe's source coordinate with its destination coordinate/range. It is what is_src() / is_dst() / is_active() are evaluated against, and is used as runtime guards on pipe-coupled regions. Helps answer "which exact nodes are allowed to execute pipe-coupled code?".

Work extent is a bounding box — the smallest axis-aligned rectangle covering every source coordinate and every destination range across all pipe nets. The implementation uses it to reason about the geometric region the operation touches. Helps answer "what rectangle do pipe coordinates live in?"

phizalev-TT: > The generic SPMD term is about any program running in parallel -- whether all parallel threads of control execute the same program or not (when using is_src, etc, they do not). I can try rewording this.

Agree on general SPMD meaning, but in the spec it is used in multi-device sense that follows TT-NN documentation. So maybe let's not use it here to avoid confusion.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -642,13 +658,20 @@ A *pipe net* is a communication primitive that groups pipes into a network. A pi
...
+Any non-pipe-related code segment can execute on all nodes of
+the operation's grid or have its own guards that are independent of pipes.
 
 | Function | Description |
 | :---- | :---- |
 | `ttl.PipeNet[DstT](pipes: List[ttl.Pipe[DstT]]) -> ttl.PipeNet[DstT]` | Constructs pipe net. |
 | `ttl.PipeNet[DstT].if_src(self, cond_fun: Callable[[ttl.SrcPipeIdentity[DstT]], None])` | Call condition function for each pipe in the pipe net that is a source. |
 | `ttl.PipeNet[DstT].if_dst(self, cond_fun: Callable[[ttl.DstPipeIdentity], None])` | Call condition function for each pipe in the pipe net that is a destination. |
+| `ttl.PipeNet[DstT].is_src(self) -> bool` | Predicate: `True` on the current node iff it is a source coordinate of any pipe in the pipe net. |
```
zoecarver: nit: "iff"

lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
```diff
@@ -0,0 +1,1025 @@
...
+  }
+  return {result, nullptr};
+}
+
+// Return whichever of `lhs`/`rhs` has the lexicographically smaller
+// stringified `getLoc()`. Used to pick a deterministic op when the dataflow
+// solver could otherwise carry either of two unanalyzable predicates through
+// the lattice; without this, diagnostic notes change op across runs depending
+// on solver order. Either argument may be null.
+Operation *firstByLocation(Operation *lhs, Operation *rhs) {
```
zoecarver: This seems... unideal. I don't see why this would be non-deterministic, but I understand we maybe want to consistently put a diagnostic on the first op. Anyway, maybe we can compare getLoc directly instead of stringifying? Or better yet use some dominance info or another mlir utility?

zoecarver (approved):
Thank you for all the improvements, looks good to me!

docs/sphinx/specs/TTLangSpecification.md
brnorris03: expanded into "if and only if"

lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
brnorris03: Refactored this, I think it improves it (and added tests for unanalyzable cases, had to fix a bug in the location of the user guard when present in such cases).

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -642,13 +658,20 @@ A *pipe net* is a communication primitive that groups pipes into a network. A pi
...
+
+The *active set* of an operation is the union, over every pipe in every pipe net in scope of the operation (constructed in its body or captured from an enclosing scope), of the pipe's source coordinate and its destination coordinate or range. The *role domain* of a pipe net is its per-net active set; `pipe_net.is_src()`, `pipe_net.is_dst()`, and `pipe_net.is_active()` are predicates that evaluate to `True` on the source role, destination role, and their union, respectively. Implementations must verify that user-written guards on pipe-coupled regions imply the correct pipe net role: a `ttl.copy` with a pipe source must be reachable only from the pipe's source coordinate; a `ttl.copy` with a pipe destination must be reachable only from the pipe's destination range; a dataflow buffer wait paired with a pipe-coupled push must be reachable only from a node where some producer pushes. The active predicates are only required for code that includes pipe-coupled computations.
+Any non-pipe-related code segment can execute on all nodes of
+the operation's grid or have its own guards that are independent of pipes.
```
phizalev-TT: Should these two separate lines be part of the paragraph above?

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -642,13 +658,20 @@ A *pipe net* is a communication primitive that groups pipes into a network. A pi
...
+
+The *active set* of an operation is the union, over every pipe in every pipe net in scope of the operation (constructed in its body or captured from an enclosing scope), of the pipe's source coordinate and its destination coordinate or range. The *role domain* of a pipe net is its per-net active set; `pipe_net.is_src()`, `pipe_net.is_dst()`, and `pipe_net.is_active()` are predicates that evaluate to `True` on the source role, destination role, and their union, respectively. Implementations must verify that user-written guards on pipe-coupled regions imply the correct pipe net role: a `ttl.copy` with a pipe source must be reachable only from the pipe's source coordinate; a `ttl.copy` with a pipe destination must be reachable only from the pipe's destination range; a dataflow buffer wait paired with a pipe-coupled push must be reachable only from a node where some producer pushes. The active predicates are only required for code that includes pipe-coupled computations.
```
phizalev-TT: Can we expand on "a dataflow buffer wait paired with a pipe-coupled push must be reachable only from a node where some producer pushes"?

phizalev-TT (approved):
Formatting nit and a question. Otherwise LGTM!

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -628,7 +628,23 @@ Blocks have a life cycle that starts with acquisition by using dataflow buffer's
...
+
+The launch extent is selected by the value passed to `grid=`:
+
+| Value | Launch extent |
+| :---- | :---- |
+| `"full"` | The device compute grid, regardless of pipe coordinates. |
+| Tuple | Used verbatim. |
+| `"auto"` (future) | Currently the same as `"full"`. In future may provide grid-sizing-related optimizations. |
+
+Whenever the launch extent is wider than the active set (which is always the case under `"full"` or any explicit tuple, and also currently under `"auto"`), the user must guard pipe-coupled regions with `net.is_src()` / `net.is_dst()` / `net.is_active()` (or equivalent coordinate predicates) so that nodes outside the corresponding role skip the pipe-coupled work. The implementation must verify these guards statically and reject programs in which a pipe-coupled operation is reachable from a node outside its declared role.
```
FConstantinos: > The implementation must verify these guards statically and reject programs in which a pipe-coupled operation is reachable from a node outside its declared role.

I think this means two things:

1. First that the sources and destinations of pipes can be known at compile time (e.g they are not set during runtime by the user)
2. More important from the simulator perspective is that this statement demands that the simulator as well needs to do those static checks, since the simulator itself is an implementation of TTLang

Are we OK with that? I think the simulator can do those checks if 1) holds, but it will be a lot of work (constant folding and propagation passes?).The simulator can do those checks at runtime though fairly easily, but that's not a static analysis.

So perhaps better phrasing would be "either statically or dynamically" or maybe "the implementation should make a best effort to" or something like that

python/sim/ttnnsim.py
```diff
@@ -1790,6 +1790,22 @@ def multiply(
     return Tensor(a_t * b_t)
 
 
+def matmul(
+    a: Union[Tensor, torch.Tensor],
```
FConstantinos: why Union here? Do the corresponding ttnn functions also work with torch.Tensor, not just ttnn.Tensor?

docs/sphinx/specs/TTLangSpecification.md
brnorris03: This PR implements the static verification of guards (that's the main contribution here). I asked Claude to check tt-metal for any kernels that have dynamic communication patterns (basically the equivalent of our pipe sources and destinations being dependent on computed values), and it did not find any.

FConstantinos (approved):

docs/sphinx/specs/TTLangSpecification.md
brnorris03: I removed the implementation requirements parts and joined the summary paragraph with the preceding (initially I didn't want it to get missed in the detailed decsription).

docs/sphinx/specs/TTLangSpecification.md
brnorris03: As we discussed, I removed the implementation "requirements" parts.

python/sim/ttnnsim.py
brnorris03: Good catch! Changed to `Tensor`.

docs/sphinx/specs/TTLangSpecification.md
brnorris03: I ended up removing that part since it's an implementation-related suggestion, which we agreed should be done separately.

PR #549, opened 2026-05-03
[ttl][pipe] support overlapping multicast destinations in PipeNet (#505)

### LINKED ISSUE #505: Support overlapping multicast destinations in PipeNet

brnorris03:
## Problem

Closes #505. PipeNets whose pipes share a destination core were rejected at construction by `_validate_no_overlapping_destinations` (added in #496) in `python/ttl/_pipenets/__init__.py`. This requires tt-lang users to split logically-related broadcasts (e.g. a row broadcast and a column broadcast in a 2D matmul) into separate PipeNets and synchronize them by hand, even though the spec (`ttl.PipeNet[DstT](pipes: List[ttl.Pipe[DstT]])`) imposes no such restriction.

A concrete example. On a 2x2 grid, give one PipeNet two multicast pipes:

```python
ttl.PipeNet([
    ttl.Pipe(src=(0, 0), dst=(slice(0, 2), 0)),   # P1: row 0
    ttl.Pipe(src=(0, 1), dst=(slice(0, 2), 0)),   # P2: row 0 again
])
```

Both pipes target the same row, so the receiver at (0, 0) sees arrivals from two senders within one PipeNet. Before this PR, construction raises an exception; the receiver-side handshake was not able to disambiguate the two arrivals.

## The handshake — before and after

Old protocol (one-shot write; some args elided for brevity, see `lib/Dialect/TTL/Transforms/PipeLowering.cpp` on `origin/main` for the full lowering):

```
sender:    noc_async_write_multicast(data, recv_slot_addr, num_dests)
           noc_async_write_barrier()
           noc_semaphore_set_multicast(recv_sem, 1, num_dests)   // remote sems <- 1
receiver:  experimental::semaphore_wait(recv_sem, 1)             // blocks until recv_sem == 1
           consume tile
           noc_semaphore_set(recv_sem, 0)                        // reset for next round
```

`set_multicast` writes a fixed value, not an increment. After any one sender's `set_multicast(1)`, the receiver's semaphore reads 1; a second sender writing 1 again is observably a no-op. The receiver's `semaphore_wait(recv_sem, 1)` returns on the first arrival, the receiver consumes the corresponding slot and resets the semaphore — by the time the second sender writes 1, the receiver may have already moved on, or, if the second sender wrote between the first wait-returning and the reset, the receiver loses that signal entirely. Either way the receiver cannot count arrivals.

New protocol (cumulative counter):

```
                // one per (receiver, PipeNet); kernel-local memref<1xi32>
                int32_t recv_counter[1] = {0};

sender:    noc_async_write_multicast(data, recv_slot_addr, num_dests)
           noc_async_write_barrier()
           noc_semaphore_inc_multicast(recv_sem, +1, num_dests)
           noc_async_atomic_barrier()                          // order: data, then inc
receiver:  ++recv_counter[0]
           experimental::semaphore_wait_min(recv_sem, recv_counter[0])
           consume tile
```

`inc_multicast` adds to the remote semaphore. N senders each call `inc(+1)` once per round, so the remote counter records the number of arrivals. The receiver advances its local expectation by 1 per round and waits for the remote counter to reach at least that value. Each sender writes its data to a distinct slot in the receiver's CB (slot assignment from the existing slot-coloring in `PipeGraph`), so the data writes also do not collide.

With this in place, two senders to one receiver no longer race, and the `ttl.PipeNet(...)` constructor no longer rejects overlapping multicast destinations.

## What else changed

- The remaining PipeNet construction checks are kept or added: empty PipeNets, mixed unicast+multicast in one PipeNet, `slice.step != 1`, mixed coordinate ranks across pipes.

- `PipeGraph::build` previously warned and continued when a Pipe receiver was not actually a `CircularBufferType` (or its `BindCBOp` was untraceable). In that case the sender wrote to its own `write_ptr` instead of the receiver's. Both cases are now hard errors.

- New design doc `docs/development/PipeOptimizations.md`. With the `ttl.PipeNet(...)` constructor no longer rejecting overlapping multicast destinations, several follow-up rewrites become achievable as separate work: forwarding-chain lowering for 1->K broadcast, ring decomposition for N->1 reductions, wave decomposition for the case where the receiver's `block_count` would exceed the per-core CB cap of 32. None of those rewrites is in this PR; the doc captures the motivation and sketches the approaches.

- Updated `docs/development/PipeNets.md` with a "Within-PipeNet receiver semantics" section documenting slot-per-sender allocation, the per-(receiver, PipeNet) arrival counter, and a proof that overlapping multicasts retain the parallelism of non-overlapping ones (no compiler-inserted serialization between senders).

- Mechanical refactors with no behavioral effect: `PipeGraph::build` moved into its own `.cpp` (co-locate with the rest of `PipeGraph`; shrink `ConvertTTLToTTKernel.cpp`); the sender / receiver semaphore-index encoding moved to a header (so the counter-allocation code and the pattern lowerings share one encoding); module pipes pre-indexed once into a `PipeNetIndex` (avoid re-walking per pattern match); `DenseMapInfo` sentinel keys extended to fill every field (prior `{INT64_MAX, 0, 0, ...}` sentinels could collide with real keys).

## Tests

End-to-end coverage on hardware and simulator: a two-sender minimum repro (issue #505 base), multi-tile blocks with partial-overlap pipes (asymmetric arrival counts per receiver), N-to-N all-to-all on a 1D row and 2D per-column (matching the spec scatter-gather example), and a 4096x4096x4096 matmul with one row-broadcast and one column-broadcast pipe per axis (mirroring tt-metal `minimal_matmul::test_linear`; the chain-based lowering tt-metal uses is in the design doc as future work). Negative tests for construction-time rejection of `slice.step != 1` and mixed coord ranks, with simulator parity for the slice-step check.

## Follow-ups filed

- bug #585 — `test_minimal_matmul_pipes` is skipped on Wormhole because the 8-dest multicast on the A pipe leaves dispatch FW in `RUN_MSG_INIT` and the next `OpenDevice` times out. Same signature as tt-metal#43511.


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -417,61 +418,7 @@ struct TileStoreLowering : OpConversionPattern<TileStoreOp> {
 
 } // namespace
 
-//===----------------------------------------------------------------------===//
-// PipeGraph implementation
-//===----------------------------------------------------------------------===//
-
-FailureOr<PipeGraph> PipeGraph::build(ModuleOp mod) {
-  PipeGraph graph;
-
```
brnorris03, line 426: Moved to its own source file.

test/python/pipe/test_pipenet_overlap.py
```diff
@@ -0,0 +1,204 @@
...
+
+@ttl.operation(grid=(4, 1))
+def overlap_sum_kernel(inp, out):
+    """Senders (0, 0) and (1, 0) each multicast one tile to receivers
+    (2, 0) and (3, 0). Each receiver sums the two received tiles.
+    """
+    net = ttl.PipeNet(
+        [
+            ttl.Pipe(src=(0, 0), dst=(slice(2, 4), 0)),
+            ttl.Pipe(src=(1, 0), dst=(slice(2, 4), 0)),
```
zoecarver, line 34: Sorry to be annoying, but could you add a test for:
```
            ttl.Pipe(src=(1, 0), dst=(slice(2, 4), 0)),
            ttl.Pipe(src=(0, 0), dst=(slice(2, 4), 0)),
```

Is there a guarantee that `if_dst` iterates in slot order? if so would this disagree with `assignGatherSlotIndices` and cause issues (i.e., `if_dst` gets 1,0 then 0,0 but `assignGatherSlotIndices` assigns 0,0 -> slot 0 and 1,0 -> slot 1)? Could that cause a synchronization issue/hang? It might not, just want to make sure this is considered and tested as I'm not sure I completely follow the slot logic.

zoecarver (approved):
Seems like the description is out of date now, but that's fine. Maybe just update the squash commit. This LGTM! Excited to see what this new pattern unlocks.

test/python/pipe/test_pipe_patterns.py
```diff
@@ -131,61 +131,60 @@ def dm_write():
...
-# TODO(#505): blocked on noc_semaphore_inc_multicast support in TTKernel.
-# Multicast pipes in a PipeNet share a semaphore pair, so overlapping
-# multicast destinations corrupt the handshake. Once #505 lands, this
-# kernel can run with a single PipeNet.
 
 N_SG = 4
 
 
 @ttl.operation(grid=(N_SG, 1))
 def scatter_gather_kernel(inp, out):
```
zoecarver: Do we need to update the spec examples based on this?

python/ttl/pipe.py
```diff
@@ -106,6 +106,11 @@ def _validate_slice(s: slice, name: str):
                 f"dst {name} slice start must be < stop, "
                 f"got slice({s.start}, {s.stop})"
             )
+        if s.step is not None and s.step != 1:
+            raise ValueError(
+                f"dst {name} slice step must be 1 or None "
+                f"(strided multicast is not supported), got step={s.step}"
```
FConstantinos, line 112: Is that reflected in the simulator as well?

test/python/pipe/test_pipenet_overlap.py
brnorris03: The declaration order doesn't matter. Slot assignment is order-independent by construction — assignGatherSlotIndices sorts pipes by `(srcX, srcY, dstStartX, dstStartY, pipeNetId)` before assigning slots.

`if_dst` does iterate in list order, but the slot index isn't iteration-dependent — each receiver's `ttl.copy` uses `receiverInfo->gatherSlotIdx`, so each block lands in its assigned slot regardless of which `if_dst` block runs first. The cumulative-wait semaphore at the receiver counts arrivals, not slot order. 

I added the reverse order test variant.

test/python/pipe/test_pipe_patterns.py
brnorris03: This one is an all-to-all, the spec's one is per column (intentionally), so they are just doing different things. But @phizalev-TT can decide which is more useful to mention in the spec. And I added the spec example as a separate test (renamed the original to test_scatter_gather_1d and the new one is test_scatter_gather_2d).

python/ttl/pipe.py
brnorris03: It is now, please take a look

brnorris03:
> Seems like the description is out of date now, but that's fine. Maybe just update the squash commit. This LGTM! Excited to see what this new pattern unlocks.

Sorry, accidentally updated with a different PR's overview, corrected now.

zoecarver (approved):
Discussed further offline. Looks great!

FConstantinos (approved):

PR #550, opened 2026-05-05
Simulator: Add pipnet predicate functions

FConstantinos:
pipe_net.is_src/is_dst/is_active


brnorris03 (approved):
Looks great, thanks for adding so quickly. Would it also be possible to add a multicast negative test, too?

Issue #264, opened 2026-01-22
[TTL] Handle mixed f32/bf16 dtypes in compute ops


brnorris03:
## Summary
Multiple compiler passes make conservative assumptions when handling mixed dtypes (f32 + bf16 together), treating everything as f32 if any f32 is present. This wastes DST capacity and may reduce efficiency.

### 1. TTLSetComputeKernelConfig (lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp:28-45)
**Issue:** `hasF32TileArgs()` returns true if **ANY** block argument is f32, enabling `fp32_dest_acc_en` for the **entire** compute operation.

```cpp
static bool hasF32TileArgs(ComputeOp computeOp) {
  for (BlockArgument arg : body->getArguments()) {
    if (tileType.getElementType().isF32()) {
      return true;  // ← Enables fp32 for ENTIRE compute op
    }
  }
}
```

**Impact:** Forces all operations in the compute body to use fp32 accumulation, even if some operations only involve bf16 inputs and could use bf16 precision.

### 2. TTLAssignDST (lib/Dialect/TTL/Transforms/TTLAssignDST.cpp:76-79, 118-120)
**Issue:** When `fp32_dest_acc_en=true`, uses f32 DST capacity (4 tiles double-buffered, 8 single-buffered) for the entire compute operation.

```cpp
// TODO at line 76-79:
// TODO: Handle mixed dtypes - if compute block has both f32 and bf16 tile
// arguments, should we use f32 capacity (4 tiles, safer) or error? Consider
// emitting warning for mixed dtypes since f32 config reduces available DST
// capacity and may cause unexpected capacity overflow.

if (fp32DestAccEn && !elementType.isF32()) {
  elementType = mlir::Float32Type::get(computeOp.getContext());
}
```

**Impact:** Reduces available DST registers from 8→4 (double-buffer) or 16→8 (single-buffer), potentially causing capacity overflow for operations that could have used the larger bf16 capacity.

### 3. ConvertTTLToCompute (lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp:161-168)
**Issue:** Assumes all inputs/outputs in a fused compute operation have the same element type.

```cpp
Type scalarType = type.getElementType();  // ← Gets from output type
Type tileType = ttcore::TileType::get(scalarType);
for (size_t i = 0; i < trace.rootInputs.size(); ++i) {
[...]

PR #553, opened 2026-05-07
Typecast implementation

vtsilytskyiTT:
### Problem description
Typecast is not yet supported.

### What's changed
Add typecast operation.

### Limitations
Fusion is not yet supported. User needs to materialize typecast result into DFB.

### Ticket
None

### Checklist


vtsilytskyiTT:
Fix python test.

brnorris03 (approved):
This looks good to me overall. I would approve as is, but wanted to suggest an alternative -- up to you whether to consider it or not.

I think you can possibly reuse `TTL_TileUnaryOp` by changing it slightly with an extra parameter, something like:

```c++
// Was: AllTypesMatch<["input", "result"]> always added.                                    
  // New: caller chooses; default keeps existing behavior.                                    
  class TTL_TileUnaryOp<string mnemonic, list<Trait> traits = [],                             
                       bit allowDtypeChange = 0>                                              
      : TTL_TileOp<mnemonic, !listconcat(                                                     
            traits,                                                                           
            [TTL_TileUnaryOpTrait, TTL_DSTInputsTrait, TTL_InPlaceOpTrait,                    
             TTL_DstResultOpTrait],
            !if(allowDtypeChange,                                                             
                // Shape-only check; dtype may differ. Verifier in .cpp enforces
                // that input and result are TileTypes with equal shape.                      
                [],                                                                           
                [AllTypesMatch<["input", "result"]>]))> {                                     
    let summary = "Unary tile operation";                                                     
    let arguments = (ins TTL_TileType:$input, Index:$dst_index);                              
    let results = (outs TTL_TileType:$result);                                                
    let assemblyFormat =                                                                      
        "$input `into` `dst` `[` $dst_index `]` attr-dict `:` "                               
        "type($input) `->` type($result)";                                                    
  }                                                                                           
                                                                                              
  multiclass TTL_UnaryElementwisePair<string mnemonic, string kernelOp,
                                      bit allowDtypeChange = 0> {                             
    def Op : TTL_ElementWiseUnary<mnemonic> { ... }      
    def TileOp : TTL_TileUnaryOp<"tile_"#mnemonic, [], allowDtypeChange> {                    
      let description = "Tile-level "#mnemonic#". Maps to ttkernel."#kernelOp#".";            
      let hasVerifier = !if(allowDtypeChange, 1, 0);                                          
    }                                                                                         
  }        
```

Again, completely up to you.

brnorris03:
Also some more test ideas; I don't expect all of them to be appropriate (this was just me iterating with claude, I did edit them as much as I could, but sometimes it does propose things that are somewhat unrealistic or redundant, so use your judgement).

- e2e correctness pytest (in `test/python`) with `assert_pcc` against torch, parametrized over multiple dtype pairs (`bf16->f32`, `f32->bf16`, `int32->f32`, `f32->int32`, `bf16->int32`). The compile-time logic is identical for every pair (just threads `DataTypeAttr` through), so dtype parametrization adds no compile-time coverage — but each pair maps to a different SFPU LLK kernel in tt-metal. Also I think adding validation to the simple_typecast.py python lit test (instead of keeping it compile-only) would be good.
- Typecast whose input is another op's intermediate result with no `attach_cb` between them, e.g. `out = ttl.math.typecast(ttl.math.add(a, b), bf16)`. This is the only DSL-level case that exercises the `tryFusion` fallback in `LowerTypecastToCompute` (`getAttachedCB(op.getInput()) == nullptr` branch). Currently zero coverage.
- Typecast result feeding a same-dtype op in the same compute body, e.g. `add(typecast(a, f32), b_f32)`. Distinct from the above because here the typecast is the *producer* fused with a downstream consumer; the previous case has the typecast as the *consumer* of an upstream op. The fusion machinery may treat the two directions asymmetrically.
- A kernel with a `ttl.tile_typecast` plus an unrelated mixed-dtype path that needs full bf16 DST capacity — exercises the false-positive branch of the new `computeDSTCapacity` heuristic, which currently walks for *any* `TileTypecastOp` and bypasses the rejection. This is the only test that targets the newly loosened check, not the case where it correctly accepts an intentional conversion.

vtsilytskyiTT:
@brnorris03 thanks for your input!
1. typecast via TileUnaryOp feels unnatural. Like adding and exceptional case to a well-defined part of code, which doesn't require it. A whole flag for a single use case.
2. I added 2 tests and they actually discovered a bug. I forgot about unpack_to_dest_fp32. This caused data truncation and catastrophic accuracy loss. Now we have actual runtime tests for bf16->fp32 and fp32->bf16.
3. Your suggested test number 2 and 3 (fusion) will be useful, but they are limited by "TODO(#264)". I can take a look at this as a separate PR.
Thank you once again!

include/ttlang/Dialect/TTL/IR/TTL.h
```diff
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent USA, Inc.
```
brnorris03: I wouldn't change the copyright, I don't think we are supposed to change the wording (or the year). New files can be created with the correct year, but I think the rest has to be as above.

include/ttlang/Dialect/TTL/IR/TTL.h
```diff
@@ -35,6 +35,7 @@ constexpr llvm::StringLiteral kCBIndexAttrPrefix("ttl.cb_index.");
 constexpr llvm::StringLiteral kTargetArchAttrName("ttl.target_arch");
 constexpr llvm::StringLiteral kFp32DestAccEnAttrName("fp32_dest_acc_en");
 constexpr llvm::StringLiteral kDstFullSyncEnAttrName("dst_full_sync_en");
+constexpr llvm::StringLiteral kUnpackToDestFp32AttrName("unpack_to_dest_fp32");
```
brnorris03: I think all of these should have `ttl.` prefix since they are discardable attributes (a couple from before seem to have slipped in without). I don't expect you to change the older ones, but suggest changing the new one to start with `ttl.`.

python/ttl/operators.py
```diff
@@ -643,6 +643,48 @@ def fill(output: TensorBlock, value) -> TensorBlock:
...
 
+@syntax("typecast")
+def typecast(input: TensorBlock, dtype) -> TensorBlock:
+    """
+    Elementwise typecast: convert each element of ``input`` to ``dtype``.
+
+    Args:
+        input: Input tensor (CB-attached). Each element is a tile.
+        dtype: Target data type. Accepts a ``ttcore.DataType`` enum value
+            or a torch/ttnn dtype convertible via ``dtype_utils``.
```
zoecarver, line 671: Hmm does it make sense to accept a torch.dtype here?

vtsilytskyiTT: I don't see a reason not to. It is trivially convertible to TTNN type.
update: I see 2 reasons pro torch types:
1. There is a special conversion function already existing `tensor_dtype_to_ttcore_datatype`
2. For `typecast` it is natural to cast types. This time we notify user that function accepts ttcore types, but may take TTNN/torch and convert them to expected type.

include/ttlang/Dialect/TTL/IR/TTL.h
vtsilytskyiTT: OK, I'll revert copyright changes. It is strange to have different policies across one organization. In tt-metal we always bump the year in copyright.

include/ttlang/Dialect/TTL/IR/TTL.h
vtsilytskyiTT: Done

include/ttlang/Dialect/TTL/IR/TTL.h
vtsilytskyiTT: Done

zoecarver (approved):
This is great, thank you Vlad! As a follow up maybe you can propose a spec + sim change as well? Or we can let someone else handle that.

test/python/test_typecast.py
```diff
@@ -0,0 +1,153 @@
...
+
+_CONFIGS = [
+    pytest.param(
+        _typecast_bf16_to_f32,
+        torch.bfloat16,
+        ttnn.bfloat16,
+        torch.float32,
+        ttnn.float32,
+        id="bf16_to_f32",
+    ),
```
zoecarver: Actually before we land, could you add a few more test cases? Here are some ideas:
* bf4 / bf8 / bf16 / f32 combinations/sweeps
* multiple tensor inputs used together with compute: eg input of bf16 and f32, cast bf16 -> f32, then add together cast + other arg (and maybe some fusions that test that our dst allocator still allocates the right size intermediates for longer fusions)
* larger block sizes, these are all 1x1 but we have some new logic in lowering to compute which we should make sure is tested

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -41,19 +42,50 @@ static bool hasF32TileArgs(ComputeOp computeOp) {
...
 }
 
-static bool hasF32InputTileArgs(ComputeOp computeOp) {
-  Block *body = &computeOp.getRegion().front();
-  if (!body) {
+static bool isF32CB0InputBlockArgument(Value value, ComputeOp computeOp) {
+  auto arg = dyn_cast<BlockArgument>(value);
+  if (!arg || arg.getOwner() != &computeOp.getRegion().front()) {
+    return false;
+  }
```
brnorris03, line 49: Slightly cleaner, IMO:
`if (!arg || arg.getParentRegion() != &computeOp.getRegion())`

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -41,19 +42,50 @@ static bool hasF32TileArgs(ComputeOp computeOp) {
...
+  unsigned argNumber = arg.getArgNumber();
+  if (argNumber >= computeOp.getNumInputs()) {
+    return false;
+  }
+  Value cb = getAttachedCB(computeOp.getInputs()[argNumber]);
+  if (!cb || getCBIndex(cb) != 0) {
     return false;
   }
+  std::optional<mlir::Type> elementType = getTileElementType(arg.getType());
+  return elementType && elementType->isF32();
```
brnorris03: nit: could skip non-f32 tiles before tracing CBs.
```c++
    std::optional<Type> elemTy = getTileElementType(arg.getType());
    if (!elemTy || !elemTy->isF32())
      return false;
  
    Value cb = getAttachedCB(computeOp.getInputs()[argIdx]);
    return cb && getCBIndex(cb) == 0;
```

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -1434,6 +1434,85 @@ struct LowerFillToCompute : OpRewritePattern<FillOp> {
...
 
+//===----------------------------------------------------------------------===//
+// Typecast Lowering
+//===----------------------------------------------------------------------===//
+
+/// Lowers ttl.typecast to ttl.compute with ttl.tile_typecast in the body.
+/// Cannot reuse buildComputeFromInputs because the input and output tile
+/// element types differ; that helper assumes a single tile element type for
+/// all block arguments.
+struct LowerTypecastToCompute : OpRewritePattern<TypecastOp> {
```
brnorris03, line 1453: This pattern duplicates most of `buildComputeFromInputs` above; the only real difference is that the input and output block args need different tile element types. Consider extending `buildComputeFromInputs` to derive each block-arg tile type from the corresponding input/output value's element type rather than reusing `outputType.getElementType()` for all of them. Then this pattern collapses to a single `buildComputeFromInputs(...)` call like the other unary lowerings.

python/ttl/_src/ttl_ast.py
```diff
@@ -601,6 +601,21 @@ def visit_Attribute(self, node, func_args=[], kwargs={}):
...
+                # downstream syntax handlers receive the real dtype rather
+                # than triggering the parent's "expression does not produce
+                # a value" diagnostic.
+                if (
+                    not func_args
+                    and not kwargs
+                    and isinstance(node.value, ast.Name)
+                    and node.value.id in self.fn_globals
+                    and hasattr(self.fn_globals[node.value.id], node.attr)
+                ):
```
brnorris03, line 617: How do we know that the underlying Python object is a dtype Value? What if the code is something like `torch.zeros` or really any other no-arg function (instead of a dtype value) -- I think this would treat any no-arg function as a Value and never evaluate the call?

What about something like:
```python
from .dtype_utils import tensor_dtype_to_ttcore_datatype
if (
    isinstance(node.value, ast.Name)
    and node.value.id in self.fn_globals
):
    obj = getattr(self.fn_globals[node.value.id], node.attr, None)
    try:
        tensor_dtype_to_ttcore_datatype(obj)
    except (ValueError, TypeError):
        pass
    else:
        return obj
```

I would also add a regression test for this, e.g., something like `ttl.math.typecast(x, torch.zeros)` (or any no-arg function that's not a dtype).

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -122,6 +173,23 @@ struct TTLSetComputeKernelConfigPass
     }
     funcOp->setAttr(kEnableFPUBinaryOpsAttrName,
                     BoolAttr::get(funcOp.getContext(), enableFPUBinaryOps));
+
+    bool needsUnpackFp32 = false;
+    funcOp->walk([&](ComputeOp computeOp) {
+      if (needsUnpackFp32) {
+        return WalkResult::interrupt();
+      }
+      if (needsUnpackToDestFp32(computeOp)) {
```
brnorris03: this `if (needsUnpackFp32)` is dead code, can remove

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -41,6 +42,52 @@ static bool hasF32TileArgs(ComputeOp computeOp) {
...
+  return elementType && elementType->isF32();
+}
+
+static bool isSFPUTileStrategyOp(Operation *op) {
+  if (!isTileComputeOp(op)) {
+    return false;
+  }
+  if (isa<TileReduceOp, TileMatmulBlockOp>(op)) {
+    return false;
+  }
```
brnorris03, line 71: The function name is a bit misleading, this checks for a bunch of things, not just SFPUTileStrategy ops. 

Also, I think the intent is to filter FPU ops? If so, perhaps we should define add `TTLFPUOpTrait` to the above ops and simplify the check to `op->hasTrait<TTLSFPUOpTrait>() &&  !isFPUEligibleBinaryOp(op)`? (for completeness should probably also have `TTLSFPUOp` trait, as well). WDYT?

(If you agree, can be just a TODO, not part of this PR)

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -459,10 +459,21 @@ inline FailureOr<std::uint32_t> computeDSTCapacity(ComputeOp computeOp) {
...
+    // conversion (ttl.tile_typecast); fall through to f32 capacity, which is
+    // conservative and safe for that pattern. Otherwise, reject the mix
+    // because the kernel may have been authored assuming bf16 capacity.
+    bool hasIntentionalDtypeConversion = false;
+    computeOp.getRegion().walk([&](TileTypecastOp) {
+      hasIntentionalDtypeConversion = true;
+      return WalkResult::interrupt();
+    });
+    if (!hasIntentionalDtypeConversion) {
+      return computeOp.emitOpError(
```
brnorris03:
```suggestion
if (!computeOp.containsOp<TileTypecastOp>()) {
  return computeOp.emitOpError(...);
```

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -965,6 +965,48 @@ defm TTL_Signbit : TTL_UnaryElementwisePair<"signbit", "signbit_tile">;
...
+//===----------------------------------------------------------------------===//
+// Typecast Operations (Tensor + Tile)
+//===----------------------------------------------------------------------===//
+//
+// `ttl.typecast` is a unary elementwise op that converts the data type of
+// every element. It does not fit the standard unary elementwise multiclass
+// because input and result element types differ; the multiclass requires
+// AllTypesMatch on input and result. The destination dtype is encoded in the
+// result type and is threaded to ttkernel.typecast_tile and its init op
+// during lowering by reading it from the result tile element type.
```
brnorris03, line 977: The "multiclass requires AllTypesMatch" reason mixes tile and tensor sides — `TTL_ElementWiseUnary` doesn't have AllTypesMatch, only `TTL_TileUnaryOp` does. Can we trim this? The "threaded to typecast_tile during lowering" part also seems like it belongs in the .cpp rather than the .td.

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -284,6 +284,20 @@ mlir::LogicalResult mlir::tt::ttl::CopyTileOp::verify() {
   return success();
 }
 
+mlir::LogicalResult mlir::tt::ttl::TileTypecastOp::verify() {
```
brnorris03, line 287: Should we also reject identity casts and non-float element types here? An integer-to-float `tile_typecast` parses but lowers to invalid hardware.

brnorris03 (commented):
A few more mostly minor things, plus some more tests to consider:
- Negative AssignDST case: `tile_typecast` plus an unrelated stray dtype mix (should still be rejected).
- `unpack_to_dest_fp32` positive: f32-to-bf16 `tile_typecast` from CB0 sets the attribute.
- `unpack_to_dest_fp32` negative: bf16-to-f32 `tile_typecast` from CB0 does not set it.
- TTLToTTKernel: a case where the `tile_typecast` result is actually consumed (chained SFPU or copy_dst), since the current test leaves it unused.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -459,10 +459,21 @@ inline FailureOr<std::uint32_t> computeDSTCapacity(ComputeOp computeOp) {
...
-    return computeOp.emitOpError(
-        "mixed f32 and non-f32 tile arguments; "
-        "DST capacity uses f32 limits (4 tiles) which may produce "
-        "incorrect results");
+    // Mixed dtypes are intentional when the body performs an explicit dtype
+    // conversion (ttl.tile_typecast); fall through to f32 capacity, which is
+    // conservative and safe for that pattern. Otherwise, reject the mix
+    // because the kernel may have been authored assuming bf16 capacity.
+    bool hasIntentionalDtypeConversion = false;
+    computeOp.getRegion().walk([&](TileTypecastOp) {
```
brnorris03: This accepts any typecast anywhere in the body — what about a body that has a legitimate typecast and also an unrelated stray mix (e.g. `tile_add` on f32+bf16 args)? We would still want to catch that. Could we check that the block-arg element type set equals `{src, dst}` for some typecast and reject otherwise?

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -1434,6 +1434,85 @@ struct LowerFillToCompute : OpRewritePattern<FillOp> {
...
+  LogicalResult matchAndRewrite(TypecastOp op,
+                                PatternRewriter &rewriter) const override {
+    auto inputType = getTensorType(op.getInput());
+    auto resultType = getTensorType(op.getResult());
+    if (!inputType || !resultType) {
+      return failure();
+    }
+
+    if (!getAttachedCB(op.getInput())) {
+      return tryFusion(op, rewriter);
```
brnorris03, line 1464: `TypecastOp` carries `TTL_UnaryElementwiseOpTrait`, so `traceFusionToRoots` will recurse through it, but `emitTileOpFor` has no `TypecastOp` case and returns nullptr. So fusion of any chain containing typecast hits "unsupported op type" and fails. `o.store(typecast(exp(x), f32))` is one example. Is that the intended limitation, or am I missing a case that handles it elsewhere?

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -41,6 +42,52 @@ static bool hasF32TileArgs(ComputeOp computeOp) {
   });
 }
 
+static bool isF32CB0InputBlockArgument(Value value, ComputeOp computeOp) {
```
brnorris03, line 45: The `getCBIndex(cb) != 0` check here, combined with `_set_unpack_to_dest_fp32` only configuring CB0, seems to imply that an f32 SFPU consumer reading from any non-CB0 CB never gets `UnpackToDestFp32` mode — the unpack would use `Default` and possibly some f32 inputs would lose precision silently at runtime? Maybe the attribute can include the set of CB indices that need the mode (e.g. `DenseI32ArrayAttr`) and the Python side iterate it? Or at least a diagnostic when an f32 SFPU consumer on a non-CB0 CB is detected. 

Also what happens if that CB is used by an FPU op? Does something ensure it isn't used (I may have missed it elsewhere)?

Also, could you add a docstring on this function explaining what it checks and why — the name says `isF32CB0InputBlockArgument` but the role in the unpack-mode decision chain is not obvious from the call site.

python/ttl/operators.py
```diff
@@ -643,6 +643,48 @@ def fill(output: TensorBlock, value) -> TensorBlock:
...
+        Result tensor with the same shape as ``input`` but with the element
+        type derived from ``dtype``.
+    """
+    from ttl.dialects import ttcore
+    from .dtype_utils import tensor_dtype_to_ttcore_datatype
+
+    if isinstance(dtype, ttcore.DataType):
+        ttcore_dtype = dtype
+    else:
+        ttcore_dtype = tensor_dtype_to_ttcore_datatype(dtype)
```
brnorris03, line 683: Should we reject non-float dtypes here with a `ValueError`? The test file says integer-to-float is not supported but the wrapper accepts anything.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: Agree, this is cleaner.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: Yes, I totally support fine-grained traits! Let me add a TODO.

vtsilytskyiTT: Does `isDstInputTileComputeOp` function name better reflect the purpose?

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
vtsilytskyiTT: Agree, containsOp is a cleaner to do a sanity check.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
brnorris03: Yes, I like `isDstInputTileComputeOp` better.

test/python/test_typecast.py
vtsilytskyiTT: Bullets 1 and 3 - OK, implemented.
Bullet 2 is tricky. User needs to materialize result into DFB to pass it in fused computation. For now I mark the test as `xfail` and suggest a separate PR (because it changes compiler in general, not specific to typecast).

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Yes, you are right, this is an awful code duplication. Fixed.

zoecarver:
> Bullet 2 is tricky. User needs to materialize result into DFB to pass it in fused computation. For now I mark the test as xfail and suggest a separate PR (because it changes compiler in general, not specific to typecast).

I'm OK doing this as a follow up, but I don't see the XFAIL'd test here. I'm confused as to why this wouldn't work if we do something like: 

```
bf16_cb = make_dfb_like(bf16_tensor)
f32_cb = make_dfb_like(f32_tensor)
out = make_dfb_like(f32_tensor)

f32_cb.reserve().store(typecast(bf16_cb.wait()))
out.reserve().store(f32_cb.wait() + fill(42.0))
```

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
vtsilytskyiTT: Fixed, added test with mixed-in input.

lib/Dialect/TTL/IR/TTLOps.cpp
vtsilytskyiTT: Agree, accept only floats. Done.

python/ttl/operators.py
vtsilytskyiTT: Agree, done

PR #554, opened 2026-05-08
[fix][ttl] ttl-insert-cb-sync: Allow tensor SSA uses past the next-acquire boundary (#536)

### LINKED ISSUE #536: Auto-inserted cb_pop_front hoisted past subsequent cb_wait_front in same DM thread

### LINKED ISSUE #556: [ttl] Coalesce consecutive cb_wait acquires into multi-tile cb_wait_front(N) with per-acquire src_idx
## Problem

A sequence of `N` consecutive `cb_wait` ops on the same DFB lowers to `N` separate `cb_wait_front(1)` / `cb_pop_front(1)` calls. The buffer exposes a single FIFO front pointer, so the auto-pop pass must place `cb_pop` ops in the same op order it observes consumer uses; out-of-declaration-order consumes (consumer reads tile 2 before tile 1) cannot be expressed without violating FIFO monotonicity of the front pointer.

## Reproducer

`test/python/test_auto_pop_push.py::test_reordered_consumes_violate_fifo_xfail` (branch `bnorris/auto-push-pop-fixes`).

`@pytest.mark.xfail(strict=True)`. Flips to PASS once this lands.

## Proposed change

Detect strictly-consecutive acquires on the same DFB sync class (same block, only `attach_cb` / arith constants between them) and lower the group as one `cb_wait_front(N)` plus per-acquire `src_idx` in `[0, N)` plus one `cb_pop_front(N)`. Each acquire then addresses its tile by index, decoupling consume order from release order.

TableGen extension: optional `num_tiles` and `tile_offset` attributes on `CBWaitOp` / `CBPopOp` (and their producer-side mirrors `CBReserveOp` / `CBPushOp`). `num_tiles=0` marks a subsumed acquire whose `cb_wait_front` is elided; the SSA result still anchors downstream `src_idx` indexing.

Lowering: `ConvertTTLToTTKernel.cpp` reads the new attributes; `ConvertTTLTileOpsToTTKernel.cpp` folds `tile_offset` into `flatSrcIndex`. Independent of #555, but the unified ownership simplifies the coalescing analysis.

Related: #536.

brnorris03:
Problem:
`ttl-insert-cb-sync` inserts `cb_pop` / `cb_push` after the last use of each acquired slot, but only searched up to the next same-class acquire on the same DFB. Tensor SSA uses can legitimately appear after a later acquire (a tile from `cb_wait t1` may be consumed after `cb_wait t2`). Those uses were missed, the pop was inserted right after the acquire, and the read pointer advanced before the data was read. The case_a / case_b reproducers from the issue 536 follow-up [comment](https://github.com/tenstorrent/tt-lang/issues/536#issuecomment-4372107754) miscompiled silently.

Independently, deferred-consume patterns (`t1 = cb.wait(); t2 = cb.wait(); use(t1); use(t2)`) lowered to N independent `cb_wait_front(k)` / `cb_pop_front(k)` calls. Because metal's wait/pop primitives are non-cumulative, the first pop advances the front before the producer has pushed enough tiles to satisfy the next read.

Fixes (~530 LOC):
1. Split `findLastOwnedUse`: direct CB uses keep the boundary, tensor SSA uses ignore it. Extend `findOwnedReleases`' upper bound to the last owned use.
2. New pass `ttl-coalesce-dfb-acquires` rewrites a maximal run of same-DFB acquires + their matched releases into the canonical tt-metal cumulative-wait shape (`cb_wait_front(N*k)` + per-block `tensor.extract_slice` views + `cb_pop_front(N*k)`). `addSliceOffset` already folds the slice offsets into the per-tile `src_idx` / `dst_idx`, so no lowering changes. Same template handles producer-side `cb_reserve`/`cb_push`. The detection rule is locally-checkable: an op between members terminates the group iff it touches our DFB or consumes a group member's result, or carries a region (the conservative envelope of "could trigger a release on the DFB"). This means matmul-style patterns where acquires on two DFBs are interleaved (`a1, b1, a2, b2`) coalesce per-DFB independently, since an acquire on a different DFB is benign. See `docs/development/DFBManagement.md` for the full correctness argument. Fixes #556.

Tests:
- test/python/test_auto_pop_push.py: device coverage including the case_a / case_b reproducers, an interleaved-baseline regression guard, and axis-coverage cases (fanout, mixed immediate/deferred, long chains, multi-DFB, multi-tile geometry, tight block_count, single-slot, long DM loops, producer/consumer in scf.for, multiple direct DFB uses on one acquire). The reordered-consume case (formerly xfail #556) now runs as a passing test.
- test/ttlang/Dialect/TTL/Transforms/insert_cb_sync.mlir: IR-level coverage of the deferred-use pattern for compute-thread consumers (flat and inside scf.for) and producers, plus a regression guard for the scf.for iter_args pattern.
- test/ttlang/Dialect/TTL/Transforms/coalesce_dfb_acquires.mlir: positive (consumer / producer / loop-body / matmul-style two-DFB interleaved), negative (interleaved consume, single-acquire-per-DFB alternation, pre-existing num_tiles), and idempotency coverage for the new pass.
- docs/development/DFBManagement.md: extended "DFB Sync Insertion" with the ownership criteria, release-placement invariants, and idempotency.

Future work:
- #555: encode DFB ownership in SSA on `ttl.copy` operands. Would collapse `findLastOwnedUse` to a single SSA walk and lift the DM-thread xfails. Requires (mechanical) updates to a large number of lit tests.

Fixes #536 follow-up. Fixes #556.


zoecarver (approved):
LGTM!

arichinsTT (approved):
Looks good!

PR #559, opened 2026-05-11
Simulator: Default float32 promotion

FConstantinos:
The simulator promotes all ttnn datatypes to float32 so that we get a consistent experience over multiple platforms.
Also adds a flag that prevents this promotion.


docs/sphinx/simulator.md
```diff
@@ -56,6 +56,62 @@ include them (the hardware CI always does; the GitHub-hosted sim CI does not):
...
+
+- `examples/matmul_1d.py`
+- `examples/matmul_1d_mcast.py`
+- `examples/metal_examples/single_node_matmul/ttlang/single_node_matmul.py`
+- `examples/metal_examples/multinode_matmul/ttlang/multinode_matmul.py`
+
+**L1 memory budget.** The simulator uses the declared dtype for all
+`DataflowBuffer` capacity accounting so the reported footprint always matches
+what the hardware would allocate, regardless of whether float32 promotion is
+active. If the total buffer capacity for a core exceeds the L1 limit, the
```
brnorris03, line 100: I don't quite understand what that means -- if the user code has bf16, does the sim have proportionally the same L1 capacity after f32 promotion as would be available on hw with bf16 or is the capacity shrunk to what f32 on hardware would allow?

python/sim/ttnnsim.py
```diff
@@ -813,39 +813,90 @@ def nd_sharded(
...
+# can compare against them.
+bfloat16: torch.dtype = _PROMOTABLE_FLOAT_DTYPES["bfloat16"]
+float16: torch.dtype = _PROMOTABLE_FLOAT_DTYPES["float16"]
 float32: torch.dtype = torch.float32
 
-# Original value saved so set_matmul_promote_bf16 can restore it.
-_original_bfloat16: torch.dtype = torch.bfloat16
+# When True (the default), tensor creation functions promote bfloat16 and
+# float16 backing to float32 for accurate computation on all host architectures.
+# The declared dtype is always preserved in Tensor._dtype for L1 accounting.
```
brnorris03, line 850: Oh, I think that answers my question above; the promotion does not shrink the effective simulated L1 for the user, right?

FConstantinos: yes exactly. So what essentially happens is that the declared type is preserved and propagated and then using that type we can effectively compute how much L1 space would have been used if we didn't do the promotion. This is also similar to how bfloat8_b works for computing used L1 space (there is no native bfloat8_b type in torch so float16 is used)

docs/sphinx/simulator.md
FConstantinos: See below

brnorris03 (approved):
LGTM!

PR #561, opened 2026-05-12
[fix][ttl] Restrict unpack_to_dest_fp32 to SFPU consumers, prefix kernel attributes with ttl., and create separate TTLMarkFPUBinaries pass

brnorris03:
### Problem description

`TTLSetComputeKernelConfig` enables `ttl.unpack_to_dest_fp32` whenever a `ttl.compute` has any f32 input block argument. The attribute selects `UnpackToDestMode::UnpackToDestFp32` on CB index 0, which is incompatible with the SRCA/SRCB unpack stream. FPU consumers (`tile_reduce`, `tile_matmul_block`, and `tile_add`/`sub`/`mul` lowered as FPU binary ops) read from SRCA/B; setting the mode on a CB they read silently zeros the unpacked value. The criterion has to be the consumer's data-movement strategy, not the element type of the block argument.

Separately, the C++ constant `kUnpackToDestFp32AttrName` is `"ttl.unpack_to_dest_fp32"` but `_compile_ttnn_kernel` in `ttl_api.py` looks the attribute up as `"unpack_to_dest_fp32"`, so the runtime call that applies the mode is never made. `fp32_dest_acc_en` and `dst_full_sync_en` agree on both sides only because neither carries the prefix.

### What's changed

`TTLSetComputeKernelConfig` sets `ttl.unpack_to_dest_fp32` only when an SFPU-strategy op in the body reads an f32 input block argument. Operands sourced from DST (e.g. an SFPU op consuming an upstream FPU result in the same body) are excluded, because no CB unpack occurs for them.

For that consumer-strategy check to be available before `TTLAssignDST` runs, FPU eligibility of `tile_add`/`sub`/`mul` is marked in a new function-level pass `TTLMarkFPUBinaries` scheduled right after `convert-ttl-to-compute` and the equivalent detection code is removed from `TTLAssignDST`.

All three runtime kernel attributes (`fp32_dest_acc_en`, `dst_full_sync_en`, `unpack_to_dest_fp32`) carry the `ttl.` dialect prefix in both C++ and Python.

Lit coverage:
- `MarkFPUBinaries/mark_fpu_binary.mlir`: positive and negative cases for the eligibility predicate, with and without `enable-fpu-binary-ops`.
- `SetComputeKernelConfig/set_compute_kernel_config.mlir`: new `UNPACK` and `UNPACK-NOFPU` RUN lines covering f32 typecast (set), f32 reduce/matmul/fpu-add (not set), and the operand-on-DST mixed-strategy case (not set).

CI: https://github.com/tenstorrent/tt-lang/actions/runs/25713215760


docs/development/ComputeKernelConfig.md
```diff
@@ -0,0 +1,133 @@
...
+zeros the value the FPU reads. The pass must therefore consult the
+consumer's strategy, not just the dtype of the block argument. This
+mirrors tt-metal's `preserve_fp32_precision` policy in
+`ttnn/operations/eltwise/unary_ng/unary_ng.cpp` and
+`ttnn/operations/copy/typecast/typecast.cpp`, which sets the flag only for
+SFPU-strategy ops.
+
+## FPU Binary Marking
+
+`tile_add`, `tile_sub`, and `tile_mul` carry `TTLDSTInputsTrait` by default,
```
vtsilytskyiTT, line 103: This is a strange decision. As I understand tile_add is a higher-level OP that can be lowered to either FPU or SFPU implementation. If so, only SFPU version should be marked with `TTLDSTInputsTrait` trait.

vtsilytskyiTT (commented):
While this code solves the problem, it is somewhat convoluted. In MLIR paradigm `tile_add` should be an OP, that is lowered by a separate path to either `sfpu_tile_add` or `fpu_tile_add` with all required traits and properties.
I do not oppose this PR, just think that long-term solution should be different.

docs/development/ComputeKernelConfig.md
brnorris03: Yes, I see your point -- but here is a counter question -- when only some versions of the hardware support add/sub/mul on FPU, then what? I really don't want low-level hardware intricacies to be part of the top-level op definition.

brnorris03: To elaborate, my thinking on this for the dialect design is that fundamentally add/sub/mul are binary elementwise ops with register inputs, like all other elementwise ops. The ability to execute them on FPU (on some TT devices) is an optimization, and hence not really part of their fundamental semantics.

brnorris03:
> While this code solves the problem, it is somewhat convoluted. In MLIR paradigm `tile_add` should be an OP, that is lowered by a separate path to either `sfpu_tile_add` or `fpu_tile_add` with all required traits and properties. I do not oppose this PR, just think that long-term solution should be different.

Agreed about separate lowering, but that can be achieved by separate conversion patterns for the fpu vs regular versions in ttltottkernel -- there is no reason to define completely separate ops (e.g., two different elementwise adds) to accomplish that. I am not aware of any upstream (llvm-project, iree, etc) MLIR examples that do this, but happy to learn if you point me to such.

PR #562, opened 2026-05-12
Use pass to choose between FPU and SFPU binary ops

vtsilytskyiTT:
### What's changed
Remove `ttl.fpu_binary` attribute, use clear pass strategy.

### Ticket
None

### Checklist


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -1296,6 +1297,20 @@ struct TTLConvertTTLToTTKernelPass
...
+      OpPassManager pm(ModuleOp::getOperationName());
+      TTLLowerBinaryTilesOptions lowerOpts;
+      lowerOpts.enableFPUBinaryOps = enableFPUBinaryOps;
+      pm.addNestedPass<func::FuncOp>(createTTLLowerBinaryTiles(lowerOpts));
+      if (failed(runPipeline(pm, mod))) {
+        signalPassFailure();
+        return;
+      }
+    }
+
```
brnorris03, line 1313: I don't think a standalone pass here is necessary. Just a conversion in TTLToTTKernel is sufficient IMO because we'd never need to apply just these three binary ops' lowering on its own. Could just add a special conversion pattern for the FPU case (leaving the existing binary op one as is with benefit=1):

```// FPU pattern: matches AddTileOp only when isFpuBinaryEligible holds.
  template <typename SourceOp, typename InitOp, typename TTKernelComputeOp>
  struct TTLTileBinaryFPUToTTKernel : OpConversionPattern<SourceOp> {
    TTLTileBinaryFPUToTTKernel(const TypeConverter &tc, MLIRContext *ctx)
        : OpConversionPattern<SourceOp>(tc, ctx, /*benefit=*/2) {}

    LogicalResult matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
                                  ConversionPatternRewriter &rewriter) const override
   {
      auto computeOp = op->template getParentOfType<ComputeOp>();
      if (!computeOp || !isFpuBinaryEligible(op, computeOp,
                                             readEnableFpuFlag(op))) {
        return rewriter.notifyMatchFailure(op, "not FPU-eligible");
      }
      // ... existing FPU lowering body ...
    }
  };```

vtsilytskyiTT:
Closing in favor of https://github.com/tenstorrent/tt-lang/pull/568

PR #570, opened 2026-05-12
Update ksplit matmul benchmark kernel to use the new pipe APIs

zoecarver:
Makes it compile again, also cleaner logic :)


phizalev-TT (approved):
LGTM!

Issue #572, opened 2026-05-12
[ttl] element_read scan kernel compares bf16 values as raw i32 bit patterns


arichinsTT:
### Problem description

The `element_scan_kernel` in `test/python/simple_element_access.py` demonstrates
an argmax-like pattern: it reads bf16 elements via `element_read` (which returns
raw bits as `i32`), then compares them using Python `==` (lowered to `arith.cmpi`
on i32). While equality comparison of bit patterns is correct for bf16, magnitude
comparisons (`>`, `<`) on raw i32 bit patterns do not give correct results for
bf16 values because:

1. **Negative values**: bf16 uses sign-magnitude representation. Comparing raw
   unsigned bit patterns treats negative values as larger than positive values
   (e.g., -1.0 bf16 = 0xBF80 > 0.5 bf16 = 0x3F00 when compared as unsigned
   integers).

2. **Mixed signs**: A negative bf16 value has the sign bit set (bit 15), making
   its i32 representation larger than any positive bf16 value when compared as
   unsigned integers.

The current test only uses `==` comparison, which works correctly for exact
bit-pattern matching. However, if users extend this pattern to implement argmax
(finding the index of the maximum element), the comparison would be incorrect
for inputs containing negative values or mixed-sign values.

### What should change

- Document this limitation clearly in the `element_read` / `element_write` API
  documentation
- Consider adding helper functions or built-in comparisons that handle bf16
  sign-magnitude ordering correctly (e.g., converting to sign-magnitude-aware
  comparison or using float comparison after reconstruction)
- The current `element_scan_kernel` test is a compile-only test that verifies
  lowering produces correct C++ code; it does not test numerical correctness
  on hardware

### Affected files

- `test/python/simple_element_access.py` - `element_scan_kernel` function
- `python/ttl/operators.py` - `element_read()` / `element_write()` API

gHashTag:
## What this is

An async information drop on the bf16 sign-magnitude comparison question
in this issue. No PR attached, no merge requested, no action required.

---

## Background

There is a small public side-repository at:

  https://github.com/gHashTag/tt-lang-t27

It ships an 84-format numeric catalog (`formats_catalog.json`,
SHA-256 `01cd5d0b83b091bbd08345233a878c3402f6bac61db52a7b6f14b9c033677398`,
v0.3.1 on PyPI) and a handful of conformance packs with HONEST `abs_error`
where a value is not bit-exact. The catalog row for bfloat16 reads:

[...]

PR #575, opened 2026-05-12
Element Read/Write

arichinsTT:
### Problem description                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                   
  tt-lang DM (datamovement) threads had no way to read or write individual scalar elements from circular buffer (CB) blocks without host readback. This made on-device reductions like argmax impossible to express — the compute result had to be 
  copied back to the host and re-issued as a dispatch, adding significant latency and complexity.
      
### What's changed                                                                                                                                                                                                                  
 - Adds element_read and element_write operations that allow DM threads to access individual bf16/f32 elements within a CB block at a given tile coordinate [row, col], operating entirely on-device.                                               
  - New IR ops (TTL_ElementReadOp, TTL_ElementWriteOp): Read/write a single element from a CB-attached tensor at a tile coordinate. Elements are returned/accepted as i32 (raw bits).                                                                                                                                                                                                                                                                            
  - EmitC lowering pass (ttl-lower-element-access-to-emitc): Lowers the TTL ops to inline C++ lambdas that resolve the CB's L1 address (distinguishing get_read_ptr vs get_write_ptr based on whether the block came from cb.wait() or cb.reserve()),
   compute the hardware face-based tile offset, and emit the memory access.                                                                                                                                                                                                                                                                                                                                                                                                                         
  - Python frontend: Adds element_read(block, row, col) and element_write(block, row, col, value) syntax in DM kernel bodies. Scalar integer values are auto-cast to i32 at operation boundaries. Scalar variables assigned inside loop/if bodies    
  that need to outlive the scope are backed by memref<1xi32> allocated at function entry, preserving values across iterations without violating SSA dominance.                                                                                                                                                             
  - Known limitation: element_read returns raw i32 bit patterns. Equality comparison (==) on these values is correct for bf16, but magnitude comparisons (>, <) are not — bf16 uses sign-magnitude representation, so raw unsigned integer comparison
   produces incorrect ordering for negative values or mixed-sign inputs. This is tracked in #572; helper functions for bf16-aware comparison are follow-on work.

### Ticket          
https://github.com/tenstorrent/tt-lang/issues/572
### Checklist


lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
```diff
@@ -0,0 +1,410 @@
...
+///   TODO(#572): validate f32 face ordering on hardware with a runtime test.
+static void addElementAccessHelpers(func::FuncOp func, OpBuilder &builder,
+                                    bool needsBF16, bool needsF32) {
+  builder.setInsertionPointToStart(&func.getBody().front());
+  auto loc = func.getLoc();
+
+  // Use lambdas (valid inside function bodies in C++17) instead of nested
+  // function definitions, which are not allowed in C++.
+  if (needsBF16) {
+    emitVerbatim(loc,
```
zoecarver: I'm all for verbatum emitc but I wonder if there's an existing ttkernel op we can use?

zoecarver: Or maybe this can all be expressed in GEP/memref ops?

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
```diff
@@ -0,0 +1,410 @@
...
+    if (failed(elemType)) {
+      return rewriter.notifyMatchFailure(
+          op, "element access block must be a ranked tensor of tiles");
+    }
+    std::string helperName =
+        elemType->isBF16() ? "_ttl_elem_read_bf16" : "_ttl_elem_read_f32";
+
+    auto i32Type = rewriter.getI32Type();
+
+    auto *ctaOp = createLiteral(
```
zoecarver: Do these need to factor in base cta index?

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
```diff
@@ -0,0 +1,410 @@
...
+    emitVerbatim(
+        loc,
+        "auto _ttl_elem_write_bf16 = [](uint32_t l1_addr, uint32_t row,"
+        " uint32_t col, uint32_t val) {"
+        " ASSERT(row < 32 && col < 32);"
+        " volatile tt_l1_ptr uint16_t* base ="
+        " reinterpret_cast<volatile tt_l1_ptr uint16_t*>(l1_addr);"
+        " uint32_t face = (row >= 16 ? 2 : 0) + (col >= 16 ? 1 : 0);"
+        " uint32_t offset = face * 256 + (row % 16) * 16 + (col % 16);"
+        " base[offset] = (uint16_t)val;"
```
zoecarver: should we error rather than truncating?

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+    auto *readOp = createCallOpaque(
+        loc, helperName, i32Type, ValueRange{l1Addr, rowI32, colI32}, rewriter);
+
+    rewriter.replaceOp(op, readOp->getResults());
+    return success();
+  }
+};
+
+struct ElementWriteLowering : OpConversionPattern<ElementWriteOp> {
```
zoecarver: optional: most of the code here is duplicated with `ElementReadLowering` could refactor into a helper

python/ttl/operators.py
```diff
@@ -651,5 +738,9 @@ def fill(output: TensorBlock, value) -> TensorBlock:
     "grid_size",
     "signpost",
     "fill",
+    "reduce_sum",
+    "reduce_max",
```
zoecarver, line 867: unrelated?

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
arichinsTT: truncating for the type?

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
arichinsTT: I agree, looking into it rn

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
arichinsTT: probably, but will be switching to ttkernel op for getting that

lib/Dialect/TTL/Transforms/LowerElementAccessToEmitC.cpp
arichinsTT: I think the ttkernel route is the best option, tho would require one new op for the dereference for load/store, which we could to as a verbatim for now

brnorris03:
Can you explain more this design and the reasoning behind it? It would have been good to discuss this before implementing, but at least it should be documented in the docs/development/DFBManagement.md doc, focusing on the overall approach and how it fits in the rest of the compiler.

At a high level, I don't understand the rationale for using mutable `memref<1xi32>` cells for cross-scope scalars instead of SSA form. A number of analyses will not see the value dataflow, so any TTL pass that reasons about consumers of CB-attached blocks has to be patched individually to recognize the new ops, wherever an analysis walks SSA to find consumers or upstream dataflow analyses are used (having to change many passes as a result of adding an op is usually a sign of needing to rethink the design).

For example, in `ttl-coalesce-dfb-acquires`, `mayReleaseDFB` would not consider `element_read` to be a dfb consumer and the pass would incorrectly coalesce cb_waits that have element reads inbetween (after which the element read would be reading the wrong element). I can provide an example of how this results in incorrect IR. 

You may be able to alleviate some of that by adding the -mem2reg to the pipeline(s) early on, but that's more of a workaround than a better design.

There are also a lot of changes to the frontend that seem unrelated to element read/write semantics (e.g., scopes helpers and control flow related changes). I think this PR should only focus on the element access ops (definition and lowering) and not modify the front end for unrelated purposes unless they clearly fit in the design. BTW, #540 implements frontend support for the control flow in things like argmax.

PR #578, opened 2026-05-13
[fix][ttl][pipe] Fix write pointer tracking on pipe sender across iterations + verifier for DFB single-producer/consumer

### LINKED ISSUE #574: [external bug report][from claude] multi-iteration compute loop with cross-core gather+bcast produces garbage on stripe > 0
## Summary

A 2-core kernel that uses a `gather_net` + `bcast_net` pair with an outer `for _ in range(num_of_stripes)` loop produces correct results for stripe 0 but garbage for stripe 1 when `num_of_stripes >= 2`. With `num_of_stripes == 1`, the kernel works as expected.

The kernel compiles cleanly past `ttl-verify-pipenet-guards` (with `if node_col == 0:` / `if node_col > 0:` integer-constant coord guards), and it does not crash — the output tensor is just wrong on the second stripe.

This was originally hit during rmsnorm backward work; I'm filing as the reduced reproducer the user has been carrying.

## Reproducer

`grid=(2, 1)`, output shape `(64, 64)` bf16, expected to be filled with `2.0` everywhere (each core computes `1.0 + 1.0` and writes its column). With `num_of_stripes = 2`, the second 32 rows come back as a mix of `1.0` and subnormal/garbage values:

```python

import torch
import ttnn
import ttl
TILE = 32

def make_kernel():
    @ttl.operation(grid=(2, 1))
    def issue3(
        out
    ):
        out_rows = out.shape[0] // TILE
        col_cores = min(2, out_rows)
        row_cores = 1

        # change to 1 or 2 (doesn't work when set to 2 - when for loop has more than 1 iterations)
        num_of_stripes = 2

        rb = out_rows // num_of_stripes
        row_shape = (rb, 1)
        partial_cb = ttl.make_dataflow_buffer_like(out, shape=row_shape, block_count=2)
        recv_cb = ttl.make_dataflow_buffer_like(out, shape=row_shape, block_count=col_cores)
        sum_cb = ttl.make_dataflow_buffer_like(out, shape=row_shape, block_count=2)
        bcast_cb = ttl.make_dataflow_buffer_like(out, shape=row_shape, block_count=2)
        gather_net = ttl.PipeNet([
            ttl.Pipe((x, y), (0, y))
            for x in range(1, col_cores)
            for y in range(row_cores)
        ])
        bcast_net = ttl.PipeNet([
            ttl.Pipe((0, y), (slice(1, col_cores), y))
            for y in range(row_cores)
        ])

        block_shape = (rb, 1)
        out_cb = ttl.make_dataflow_buffer_like(out, shape=block_shape, block_count=2)

        @ttl.compute()
        def compute():
            node_col, node_row = ttl.node(dims=2)
            for _ri in range(num_of_stripes):
                with partial_cb.reserve() as partial_blk:
                    partial_blk.store(ttl.math.fill(partial_blk, 1.0))
                if node_col == 0:
                    partial_blk = partial_cb.wait()
                    blk = recv_cb.wait()
                    with sum_cb.reserve() as sum_blk:
                        sum_blk.store(blk + partial_blk)
                    sum_blk = sum_cb.wait()
                    with out_cb.reserve() as out_blk:
                        out_blk.store(sum_blk)
                    sum_cb.reserve().store(sum_blk)
                else:
                    blk = bcast_cb.wait()
                    with out_cb.reserve() as out_blk:
                        out_blk.store(blk)

        @ttl.datamovement()
        def dm_read():
            node_col, node_row = ttl.node(dims=2)
            for _ri in range(num_of_stripes):
                if node_col > 0:
                    blk = partial_cb.wait()
                    def send(pipe):
                        tx = ttl.copy(blk, pipe); tx.wait()
                    gather_net.if_src(send)
                    def recv(pipe):
                        b = bcast_cb.reserve()
                        tx = ttl.copy(pipe, b); tx.wait()
                    bcast_net.if_dst(recv)
                else:
                    def recv(pipe):
                        b = recv_cb.reserve()
                        tx = ttl.copy(pipe, b); tx.wait()
                    gather_net.if_dst(recv)
                    blk = sum_cb.wait()
                    def send(pipe):
                        tx = ttl.copy(blk, pipe); tx.wait()
                    bcast_net.if_src(send)

        @ttl.datamovement()
        def dm_write():
            node_col, node_row = ttl.node(dims=2)
            for _ri in range(num_of_stripes):
                r0 = node_row + _ri * 1
                out_blk = out_cb.wait()
                ttl.copy(out_blk, out[r0 : r0 + 1, node_col : node_col + 1]).wait()

    return issue3

def _to_dev(t, device):

if __name__ == "__main__":
    try:
        k = make_kernel()
        st = 64
        out_1 = _to_dev(torch.full((st, st), -42.0, dtype=torch.bfloat16), device)
        k(out_1)
        out_1 = ttnn.to_torch(out_1)[:st, :st]
        out_2 = torch.full((st, st), 2.0, dtype=torch.bfloat16)
        torch.testing.assert_close(out_1, out_2)
        print("all checks passed.")
    finally:
        ttnn.close_device(device)
```

## Expected vs observed

Expected: every element of the `(64, 64)` output is `2.0` (`1.0` produced on each core, gather+sum+bcast → both cores write `2.0` to their column for both stripes).

Observed: top half (rows `0..31`, the stripe-0 write) is all `2.0` as expected. Bottom half (rows `32..63`, the stripe-1 write) is a mix of `1.0` and subnormal/garbage:

```
[ 1.0000e+00,  1.0000e+00,  1.0000e+00,  ...,  3.6551e-38,
  2.9786e-11,  6.3477e-36],
[ 1.0000e+00, -1.5286e+38,  1.0000e+00,  ...,  2.4796e-38,
 -5.9373e-26,  4.2061e-38],
[ 1.0000e+00,  1.0000e+00,  1.0000e+00,  ...,  1.2398e-38,
  3.5345e-17,  9.0654e-35]
```

`Mismatched elements: 2048 / 4096 (50.0%)`.

Setting `num_of_stripes = 1` makes the assertion pass.

## Notes

- The kernel does compile and run after the recent PipeNet guard verifier changes — i.e. the verifier-side fix that the ksplit kernel needed is sufficient for this kernel; the remaining failure is purely a runtime/semantics issue on stripe `_ri > 0`.
- Possibly related to #527 (loop-reassignment dropping ops) but the structure here uses `sum_blk.store(blk + partial_blk)` inside a `with`-reserve, not a Python-level accumulator reassignment.
- The CB block_count for every CB is `>= num_of_stripes`, so it's not an obvious slot-starvation case.

## Environment

- Hardware: BH (Blackhole), card1
- ttlang `1.0.5.dev23`

brnorris03:
## Problem

Fixes #574: in a multi-iteration gather/multicast kernel, the pipe sender's local view of the destination DFB's write pointer did not advance, so iterations after the first wrote into stale DFB slots.

Also adds a new verifier (`ttl-verify-dfb-spsc`) that rejects any DFB used by more than one producer or consumer thread. tt-metal CBs maintain a single `pages_received` / `pages_acked` counter pair shared by both sides; two threads incrementing either counter over-count the available slots and let the producer reuse them while a consumer is still reading.
> Metal reference:  https://github.com/tenstorrent/tt-metal/blob/cf7232ab47227167f091c3c3cf6fdbed63c8b51c/tt_metal/hw/inc/internal/atomic_rwptr.h#L43
> ```// Atomic using only atomic load and store for single producer / single consumer FIFO read & write pointers.```)

## What changed

- Sender lockstep: the pipe sender now reserves and pushes on its local view of the destination DFB around each NOC write, keeping its write pointer in step with the receivers' across iterations. Loopback (sender is also a destination) is skipped because the user's receive callback on that core already advances the DFB.
- SPSC verifier: a module-level pass groups every `cb_reserve` / `cb_wait` by `cb_index` and enclosing kernel thread and emits a diagnostic when any DFB has more than one producer or consumer thread.
- Test refactors: a few existing tests shared a DFB across kernel threads. Before this PR nothing enforced SPSC, and those kernels worked at runtime because the threads happened to serialize their accesses on the small inputs the tests use. The new verifier rejects the structure regardless, so the tests are refactored to per-consumer DFBs (or, for `simple_add_loop.py`, to the spec-current `+=` accumulator).

## Follow-ups filed

- enhancement #580 — consolidate the three duplicated `ttl-to-ttkernel-pipeline` definitions.
- enhancement #581 — auto-split user DFBs whose consumers span multiple kernel threads.
- bug #582 — multi-source gather has a per-iteration write-pointer rate mismatch.
- bug #583 — cross-DFB multicast loopback still drops the stripe-1 write to the destination DFB; tracked by an `xfail` test in this PR.

### DIFF
--- a/docs/development/DFBManagement.md
+++ b/docs/development/DFBManagement.md
@@ -18,11 +18,14 @@ ttl-insert-cb-sync             (FuncOp)   Insert cb_push / cb_pop
   ... compute lowering, DST assignment, loop lowering ...
 ttl-finalize-dfb-indices       (Module)   Index reuse + limit check
 ttl-annotate-cb-associations   (FuncOp)   Copy CB indices to tile ops
+ttl-verify-dfb-spsc            (Module)   Reject DFBs shared across threads
 convert-ttl-to-ttkernel        (Module)   Lower to TTKernel dialect
 ttkernel-insert-inits          (Module)   Insert hardware init calls
 ```
 
 `ttl-finalize-dfb-indices` must precede `ttl-annotate-cb-associations` because annotation copies the `cb_index` attribute from `BindCBOp` onto tile operations (`bcast`, `reduce`, `transpose`). If annotation runs before finalization, the copied indices become stale after reuse rewrites them.
+
+`ttl-verify-dfb-spsc` must run after `ttl-finalize-dfb-indices` so every `bind_cb` carries its final `cb_index`. The pass asserts on unresolvable indices.
 
 ## DFB Lifecycle
 
--- a/docs/development/DFBManagement.md
+++ b/docs/development/DFBManagement.md
@@ -45,6 +45,77 @@ For compiler-allocated DFBs, `InsertIntermediateDFBs` creates the full sequence 
 
 A DFB's L1 memory is reclaimable after its last `cb_pop`. This defines the interval used for index reuse.
 
+## Single-producer Single-consumer Semantics
+
+### Contract
+
+Each DFB has exactly one producer thread and exactly one consumer thread. A *thread* here is a `func.func` carrying the `ttl.kernel_thread` attribute (compute, noc, ethernet); ops in untagged functions are outside the contract.
+
+The rule is inherited from tt-metal: its CB protocol is not multi-writer safe on either side. Each CB has two shared counters in `dataflow_api.h`:
+
+- `pages_received`, incremented by `cb_push_back` (producer side),
+- `pages_acked`, incremented by `cb_pop_front` (consumer side).
+
+`cb_reserve_back` blocks until `pages_received - pages_acked < block_count`. `cb_wait_front` blocks until `pages_received > pages_acked`. The protocol is correct only when exactly one thread writes each counter; the counters are not atomic with respect to multiple writers and carry no per-thread identity.
+
+### Violation
+
+A two-consumer DFB inside a stripe loop:
+
+```python
+buf = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+@ttl.compute()
+def compute():
+    for _ in range(num_stripes):
+        with buf.reserve() as b:
+            b.store(...)
+        with buf.wait() as b:       # consumer A: compute
+            ...
+
+@ttl.datamovement()
+def dm_read():
+    for _ in range(num_stripes):
+        with buf.wait() as b:       # consumer B: dm_read
+            ...
+```
+
+Per iteration, the producer pushes once (`pages_received += 1`) and each consumer pops once (`pages_acked += 2`). After iteration 0, the producer's `cb_reserve_back` on iteration 1 sees two free slots when only one has actually been consumed; it writes slot 0 while the late consumer is still reading slot 0's old data. The symmetric failure occurs with two producers: each `cb_push_back` advances the shared write pointer, and a consumer reads a partially-written slot.
+
+A single-iteration test masks this — exactly one push and two over-pops do not corrupt data when the producer never refills — so the rule must be enforced statically rather than left to test coverage.
+
+### Correct form
+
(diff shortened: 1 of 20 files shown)


zoecarver (approved):
Great improvement, thank you!

PR #589, opened 2026-05-14
Support scalar constants via fill

vtsilytskyiTT:
### Problem description
Current reduce OP accepts scaler only as a tile. This is a HW limitation, which should not bother end user.

### What's changed
1. Allow float/int scalar value as a scaler. It is implicitly transformed into a tile using ttl.fill.
2. Improve error message in `kernel_ast.py` if the value is undefined. 

### Ticket
None

### Checklist


zoecarver (approved):
This is great. Do we need a spec change?

vtsilytskyiTT:
> This is great. Do we need a spec change?

Done. Please verify whether it is clear and fits the document.

PR #590, opened 2026-05-14
Simulator: Implement dry run

FConstantinos:
- Allow runs where no computations take place, just validations
- Dry runs affect operations only inside ttl.operations
- Assumption: Computed results do not change control flow
- matmul tutorial steps 0 and 2-7 added to CI as dry runs with assertions deacticated
- Various optimizations that improve times to a third of original.

As an example, here are some performance results of a run for step_7 of matmul-tutorial on our QB:

result for regular run:
(ᴛᴛʟᴀɴɢ) kfragkiadakis@bh-qbae-15:~/tt-lang$ time tt-lang-sim examples/matmul-tutorial/step_7_multidevice_shard_k_all_reduce.py 
PCC 1.000000
PCC 1.000000
PCC 1.000000
PCC 1.000000

real    1m0.656s
user    13m32.365s
sys     0m11.490s

result for dry-run:
(ᴛᴛʟᴀɴɢ) kfragkiadakis@bh-qbae-15:~/tt-lang$ time tt-lang-sim examples/matmul-tutorial/step_7_multidevice_shard_k_all_reduce.py --dry-run

(ᴛᴛʟᴀɴɢ) kfragkiadakis@bh-qbae-15:~/tt-lang$ time tt-lang-sim examples/matmul-tutorial/step_7_multidevice_shard_k_all_reduce.py --dry-run
PCC nan
PCC nan
PCC nan
PCC nan

real    0m22.943s
user    0m40.891s
sys     0m11.905s

On the main branch:
(ᴛᴛʟᴀɴɢ) kfragkiadakis@bh-qbae-15:~/tt-lang$ time tt-lang-sim examples/matmul-tutorial/step_7_multidevice_shard_k_all_reduce.py
PCC 1.000000
PCC 1.000000
PCC 1.000000
PCC 1.000000

real    2m46.371s
user    16m12.349s
sys     11m32.285s
(ᴛᴛʟᴀɴɢ) kfragkiadakis@bh-qbae-15:~/tt-lang$ 

So it appears that the optimizations benefit regular runs as well.


python/sim/block.py
```diff
@@ -420,10 +459,13 @@ def transpose(block: Block) -> Block:
             f"transpose requires a 2-D block grid, got shape {block._shape}"  # type: ignore[attr-defined]
         )
 
+    M, N = block._shape  # type: ignore[attr-defined]
+
+    if _is_dry_run():
```
brnorris03, line 447: No dry run guard for squeeze and unsqueeze above...

python/sim/ttnnsim.py
```diff
@@ -1556,14 +1632,42 @@ def to_torch(self) -> torch.Tensor:
...
+        out_shape = torch.broadcast_shapes(self._tensor.shape, other._tensor.shape)
+        return Tensor(
+            torch.zeros(out_shape, dtype=self._tensor.dtype),
+            self._layout,
+            dtype=self._dtype,
+        )
+
+    def _zeros_matmul(self, other: "Tensor") -> "Tensor":
+        """Return zeros shaped by the matmul output of self @ other."""
+        out_shape = (*self._tensor.shape[:-1], other._tensor.shape[-1])
```
brnorris03: This shape formula is wrong for batched/broadcast matmul — (32,64)@(4,64,96) gives (32,96) in dry-run vs (4,32,96) real (tested with Claude). Infer the shape from meta tensors instead, so it matches torch and raises on bad dims?

python/sim/copy.py
```diff
@@ -98,24 +113,25 @@ def __init__(
...
-        )
+        if TRACE.enabled:
+            trace(
+                "copy_start",
+                src=type(src).__name__,
+                dst=type(dst).__name__,
+                **_copy_trace_fields(src, dst),
+            )
 
         if self._starts_on_copy():
```
brnorris03, line 136: The send (Block->Pipe) transfers here unguarded, but the receive is skipped in wait() under dry-run — so the pipe queue grows and the receiver never blocks. Dry-run pipe sequencing isn't really exercised. Guard this too, or push/pop a zero-payload marker?

python/sim/ttlang_sim.py
```diff
@@ -77,7 +82,12 @@ def execute_script_with_simulator(
     }
 
     try:
-        code = compile(script_path.read_text(), str(script_path), "exec")
+        code = compile(
+            script_path.read_text(),
+            str(script_path),
+            "exec",
+            optimize=1 if optimize else -1,
```
brnorris03, line 93: This only disables asserts in normal test mode not in dry-run mode, which would normally produce wrong results, right? Should asserts be always disabled in dry-run (the result verification ones)?

python/sim/math.py
```diff
@@ -19,13 +19,42 @@
...
 _ = matmul
 
 
+def _is_dry_run() -> bool:
+    return get_context().config.dry_run
+
+
+# Shared sentinel: a single zero-element Tensor used for all dry-run result blocks in
+# this module. Avoids any per-operation tensor allocation; content is irrelevant.
+_DRY_RUN_SENTINEL = Tensor(torch.empty(0))
```
brnorris03: The sentinel defaults to TILE layout, so a chained dry-run result reports the wrong layout — check_same_layout and the row-major guards would not be able to catch actual mismatches. Maybe include the source layout on the result?

test/scripts/tt-lang-sim-profile
```diff
@@ -0,0 +1,125 @@
...
+        file=sys.stderr,
+    )
+    print(
+        f"  checked: {_REPO_ROOT}/python_packages/ttl/sim/ttlang_sim.py (installed)",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+
+def _extract_save_arg(argv: list) -> tuple:
```
brnorris03, line 58: --save is searched across the whole argv, so `... -- --save out` steals the script's own --save. Split at the first -- and only look before it?

test/sim/test_dry_run.py
```diff
@@ -0,0 +1,258 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the simulator dry-run mode (set_dry_run).
+
+Dry-run mode skips all data movement and numerical computation while
+still exercising DFB sequencing, block state machine transitions,
```
brnorris03, line 7: Gaps vs what this docstring promises — deadlock detection under dry-run, `ttl.math` and `ttl.block` ops (e.g. the squeeze/unsqueeze issue), shape-changing ops, and dtype params (everything is float32 now).

docs/sphinx/simulator.md
```diff
@@ -177,3 +177,47 @@ Create a debug configuration in `.vscode/launch.json`:
...
+test/scripts/tt-lang-sim-profile examples/matmul-tutorial/step_1_single_node_single_tile_block.py --dry-run --save /tmp/stats.prof
+python -m pstats /tmp/stats.prof
+```
+
+The `--save FILE` option must appear after all other arguments to avoid ambiguity
+with simulator flags that also begin with `--`.
+
+## Dry-run mode
+
+Pass `--dry-run` to `ttlang-sim` (or call `sim.context.set_dry_run(True)` from Python)
```
brnorris03: nit: tt-lang-sim, not ttlang-sim

python/sim/greenlet_scheduler.py
```diff
@@ -303,150 +349,160 @@ def _initialization_phase(self) -> None:
         """
```
brnorris03, line 351: This is about existing code, specifically the " keep ts=0, giving them priority in fair scheduling." line below, which seems stale — there's no per-kernel ts=0 anymore; fair priority is position in _active now (according to the comment below that).

test/sim/test_dry_run.py
```diff
@@ -0,0 +1,258 @@
...
+        assert torch.all(result.to_torch() == 0)
+
+    def test_matmul_module_level(self):
+        a = ttnn.from_torch(torch.ones(32, 64))
+        b = ttnn.from_torch(torch.ones(64, 96))
+        result = ttnn.matmul(a, b)
+        assert result.shape == (32, 96)
+        assert torch.all(result.to_torch() == 0)
+
+    def test_relu_module_level(self):
```
brnorris03, line 150: relu(-ones) is 0 even without dry-run, so this passes either way. Suggest using a positive input so real=1, dry-run=0.

python/sim/dfb.py
```diff
@@ -1128,13 +1172,24 @@ def reserve(self, name: Optional[str] = None) -> Block:
...
-            ),
-            self.likeness_tensor.layout,
-            dtype=self.likeness_tensor.dtype,
-        )
+        #
+        # In dry-run we skip the ``torch.zeros`` + ``Tensor`` wrap entirely:
+        # the slot's bytes are never read because every downstream operation
+        # (``Block.store``, ``__matmul__``, ``ttl.math.*``, ``ttl.copy``) is
+        # itself dry-run-guarded and returns its own sentinel.  Same precedent
+        # as ``Block.__call__`` at the ``_DRY_RUN_SENTINEL if _is_dry_run()``
```
brnorris03: `Block.__call__` doesn't exist?

brnorris03 (changes requested):
Sorry it took so long! The biggest issue I see unless I missed it somehow, is that the CLI `--dry-run` path (main -> run_file) is not tested — everything here uses `set_dry_run()` or `execute_script_with_simulator(optimize=...)` directly, so things like correctness checks failing because asserts are still being executed is not caught. I think that it would be good to have some cli tests if that is available and documented as a usage path.

test/sim/test_dry_run.py
FConstantinos: fixed I believe

test/scripts/tt-lang-sim-profile
FConstantinos: fixed

python/sim/math.py
FConstantinos: fixed!

python/sim/ttlang_sim.py
FConstantinos: My intention was that asserts are NOT disabled during dry-run. But I do disable them for testing.

python/sim/copy.py
FConstantinos: I've moved this guarding to copyhandlers.py

python/sim/ttnnsim.py
FConstantinos: done, thanks!

python/sim/block.py
FConstantinos: fixed

docs/sphinx/simulator.md
FConstantinos: fixed

python/sim/greenlet_scheduler.py
FConstantinos: fixed

test/sim/test_dry_run.py
FConstantinos: fixed

python/sim/dfb.py
FConstantinos: fixed

FConstantinos:
> Sorry it took so long! The biggest issue I see unless I missed it somehow, is that the CLI `--dry-run` path (main -> run_file) is not tested — everything here uses `set_dry_run()` or `execute_script_with_simulator(optimize=...)` directly, so things like correctness checks failing because asserts are still being executed is not caught. I think that it would be good to have some cli tests if that is available and documented as a usage path.

Thank you for the thorough review!

I've added a CLI test. With regards to asserts, I expect them to not be disabled during dry-run. It's up to the user if they want to disable asserts (e.g through PYTHONOPTIMIZE=1). I do disable them but for tests only

brnorris03 (approved):
Thank you for the updates!

phizalev-TT (approved):

PR #597, opened 2026-05-19
[ttl] Support scalar multiplication (and fix reduce #594 bug)

### LINKED ISSUE #594: ttl.math.reduce_{sum,max}(inp, <numeric scalar>, dims=[0|1]) returns wrong values on multi-tile inputs
The numeric-scalar reduce API added in #589 produces incorrect output when the reduced dim leaves more than one output tile (i.e. the *non*-reduced dim spans multiple tiles). Single-tile inputs and the older tile-form scaler (`reduce(inp, scaler_tile, ...)`) are correct in these configurations.

**Observed:** `actual ≈ expected × 1.6` per output element. On `scaler=0.25`, `inp_shape=(2,2) tiles`, `dims=[0]`, bf16, `max_diff=2.97`.

**Standalone repro** (~5s):

```python
import torch, ttl, ttnn

TILE = 32

@ttl.operation(grid=(1, 1))
def reduce_kernel(inp, out):
    inp_dfb = ttl.make_dataflow_buffer_like(inp, shape=(2, 2), block_count=2)
    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 2), block_count=2)

    @ttl.compute()
    def compute_fn():
        with inp_dfb.wait() as inp_blk, out_dfb.reserve() as out_blk:
            out_blk.store(ttl.math.reduce_sum(inp_blk, 0.25, dims=[0]))

    @ttl.datamovement()
    def dm_read():
        with inp_dfb.reserve() as blk:
            ttl.copy(inp[0:2, 0:2], blk).wait()

    @ttl.datamovement()
    def dm_write():
        with out_dfb.wait() as blk:
            ttl.copy(blk, out[0:1, 0:2]).wait()

torch.manual_seed(0xBEEF)
inp_t = (torch.rand(2*TILE, 2*TILE, dtype=torch.bfloat16) - 0.5) * 4.0
out_t = torch.zeros(TILE, 2*TILE, dtype=torch.bfloat16)
inp = ttnn.from_torch(inp_t, layout=ttnn.TILE_LAYOUT, device=device, memory_config=ttnn.L1_MEMORY_CONFIG)
out = ttnn.from_torch(out_t, layout=ttnn.TILE_LAYOUT, device=device, memory_config=ttnn.L1_MEMORY_CONFIG)
reduce_kernel(inp, out)

expected = inp_t.float().sum(dim=0, keepdim=True) * 0.25
actual   = ttnn.to_torch(out)[:1, :].float()
print(f"max_diff = {(actual - expected).abs().max().item():.4f}")
ttnn.close_device(device)
```

**Scope** (bisected against pre-#589 commit `e4248ab7`; the numeric-scalar API itself doesn't exist there, so only the post-#589 behavior is directly testable):
- `dims=[0]` / `dims=[1]` multi-tile scalar-form → **broken, introduced by #589**
- `dims=[0]` / `dims=[1]` multi-tile *tile-form* → correct on both pre- and post-#589 (rules out the underlying reduce kernel; the bug is in the scalar-form post-reduce `mul_unary` path)
- `dims=[0,1]` multi-tile any form → pre-existing bug (separate issue to file)
- Single-tile non-1.0 scaler `dims=[0,1]` → pre-existing bug (separate issue to file)

**Likely root cause:** the post-reduce `mul_unary` driven by `ttl.reduce_scalar_multiplier` is applied to only a subset of result tiles when the output spans multiple tiles. Single-output-tile cases (`dims=[0,1]` collapses to one tile) hide this on the scalar form.

**Coverage:** `test/python/test_reduce_scalar.py::test_scalar_reduce_multi_tile` cross-product hits this; 24 cases currently fail. Suggest `xfail` markers referencing this issue until the fix lands.

brnorris03:
## Problem
Fixes #594. `ttl.math.reduce_{sum,max}(x, <number>, dims=[0,1])` returned `reduce(x) * c^2` instead of `reduce(x) * c`.

tt-metal `REDUCE_SCALAR` applies the scaler tile once per reduced dimension, so passing `c` directly to the LLK double-scaled 2D reductions. The first fix also put scaler rewriting in `ttl-insert-intermediate-dfbs`, mixing DFB materialization with reduce semantics.

## What's changed
`reduce_sum` and `reduce_max` no longer take a scaler argument. Scaling is now ordinary multiplication:

```python
0.5 * ttl.math.reduce_sum(x, dims=[0, 1])
```

Reduce lowering always passes `fill(1.0)` to the LLK. User scaling lowers as `ttl.mul_unary_const` / `ttl.tile_mul_unary_const`.

`ttl.mul_unary_const` is `Pure` and `TTL_UnaryElementwiseOpTrait`, so it fuses with elementwise producers. Its DFB interface materializes inputs only when the producer is reduce or matmul; elementwise and broadcast producers stay fused. This covers `c * reduce(...)`, `reduce(a) * reduce(b)`, `c * x`, `x * c`, `c * (a + b)`, chained host scalars such as `2 * 5 * (a + b)`, and scalar post-ops on matmul.

`ttl-insert-intermediate-dfbs` again only materializes operands required by `DFBInputOpInterface`; the scaler override and reduce-scaler attribute are gone.

Future work: runtime scalar scalers need separate IR/ABI support for host scalar args or DFB scalar-tile materialization for device-derived values.

## Tests
`test/python/test_scalar_mul.py` covers both reduce kinds, both dtypes, all dims, single- and multi-tile inputs, varied scalers, `fp32_dest_acc_en`, `reduce(a) * reduce(b)`, negative dims, scalar multiply over non-reduce tile blocks, fused add expressions, chained host scalars, and chained elementwise expressions. The fp32 `0.5 * x + y` non-reduce case is xfailed pending #612; bf16 coverage remains enabled. `test/python/test_matmul_fused_postops.py` covers scalar-constant matmul post-ops for 1x1x1, 2x2x2, and 1x2x4 blocks in bf16 and fp32.


test/python/test_reduce_scalar.py
```diff
@@ -0,0 +1,797 @@
...
+
+        @ttl.operation(grid=(1, 1), fp32_dest_acc_en=True)
+        def kernel(inp, out):
+            inp_dfb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=2)
+            out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+            @ttl.compute()
+            def compute_fn():
+                with inp_dfb.wait() as inp_blk, out_dfb.reserve() as out_blk:
+                    out_blk.store(scaler * ttl.math.reduce_sum(inp_blk, dims=[0, 1]))
```
vtsilytskyiTT: This is a language gap that for 1 line we should rewrite whole kernel.

vtsilytskyiTT (approved):
I must admit that this approach is much cleaner.

test/python/test_reduce_scalar.py
brnorris03: Which part? Always interested in addressing language gaps. :-)

vtsilytskyiTT: Kernel for `reduce_max` below differs by a single line (319). This is a code duplication without a good reason except language limitation.

brnorris03: Yeah. Although in this case it's mainly an artifact of lazy AI test generation; I will fix that in the tests. But you are right that reduce itself could take the operation as a parameter.

zoecarver (approved):

FConstantinos (approved):

PR #599, opened 2026-05-19
[compiler] Align bcast and fill with spec v0.17

brnorris03:
### Problem description

Spec v0.17 moves block-level operations into the `ttl.block` namespace and makes `broadcast` take an explicit result `shape`: `ttl.block.broadcast(expr, dims, shape)`. The compiler and examples still used the older output-operand form, which blocked migrating tutorials and runtime tests to the spec spelling.

### What's changed

`ttl.block.broadcast` now accepts `dims` and `shape` and produces a result tensor with the requested shape. With the output operand removed from the Python API, lowering resolves the output dataflow buffer from the downstream `ttl.store` user. Per-tile `BcastType` (Row/Col/Scalar/none) is derived from `(rank, dims)` instead of being supplied by the user.

The compiler, simulator, examples, and docs are migrated to the `ttl.block.broadcast(..., dims=..., shape=...)` spelling. The AST visitor resolves `.shape` on tensor blocks inside TTL kernels, so call sites can write `shape=y_blk.shape` instead of recomputing the shape locally.

`ttl.block.fill` remains supported in its current compiler form. Full fill dtype alignment and adding `shape` to compiler reduce calls are intentionally left for follow-up PRs.

### Follow-up work

- Align `ttl.block.fill` with the spec dtype decision once the dtype argument lands.
- Add explicit `shape` to compiler `ttl.math.reduce_sum` and `ttl.math.reduce_max` calls.
- Update examples and tests for the reduce `shape` argument after rebasing on the reduce update.

### Tests

Added ND broadcast lowering coverage in `test/ttlang/Conversion/TTLToCompute/` plus op-verifier negative tests for invalid broadcast shapes, dims, and element types. Runtime and simulator coverage was migrated to the new broadcast API.


include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1013,22 +1013,36 @@ def TTL_TileTypecastOp : TTL_TileOp<"tile_typecast",
 // Broadcast Operation
 //===----------------------------------------------------------------------===//
 
-def TTL_BcastOp : TTL_Op<"bcast", [
+def TTL_BlockBroadcastOp : TTL_Op<"block.broadcast", [
     DeclareOpInterfaceMethods<TTL_DFBInputOpInterface>]> {
-  let summary = "Broadcast a row/column/scalar tensor to full tiles";
+  let summary = "Broadcast a block over specified dimensions to a target shape";
```
arichinsTT, line 1018: is it possible to deduce the specified dimension from the input and output shape?

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1013,22 +1013,36 @@ def TTL_TileTypecastOp : TTL_TileOp<"tile_typecast",
...
-    Reads from CB, writes result to DST then packs to output CB.
-    Both input and output must be CB-attached tensors.
-
-    bcast_type specifies the broadcast dimension:
-    - Row: broadcast first row to all rows
-    - Col: broadcast first column to all columns
-    - Scalar: broadcast single value to all elements
+    Broadcasts `$input` over each dimension listed in `$dims` to produce a
+    result of shape `$shape`. For every dimension `d` in `$dims` the
+    corresponding `input.shape[d]` must equal 1; for every dimension not in
```
arichinsTT, line 1022: isn't it just a divisbility requirement of  output.shape[d] % input.shape[d] == 0, not that the input bcast dimension(s) are 1?

brnorris03: Not according to the spec.

include/ttlang/Dialect/TTL/IR/TTLOps.td
brnorris03: My very first bcast version did, so yes. This particular (stable) implementation does not (yet), but also goal is alignment with current spec, where we have to have that argument. I think eventually we will eliminate (or at least reduce) the need for the explicit shape arg.

include/ttlang/Dialect/TTL/IR/TTLOps.td
brnorris03: But would be cool to have eventually.

zoecarver (approved):

arichinsTT (approved):

FConstantinos (approved):

phizalev-TT (approved):

PR #602, opened 2026-05-19
Allow typecast fusion

vtsilytskyiTT:
### Problem description
`ttl.typecast` could not participate in fused `ttl.compute` regions, which forced dtype-changing expressions to break fusion or fail during TTL-to-compute lowering. This also made mixed dtype fusion behavior unclear, especially when f32 inputs required `unpack_to_dest_fp32` handling.

### What's changed
This change enables `ttl.typecast` fusion by deriving tile result types from each source op's tensor result instead of relying on a single default tile type. It updates compute lowering, Python AST/API type handling, and kernel config assignment so fused compute regions can preserve dtype-changing intermediates correctly.
The compute kernel config logic now tracks `unpack_to_dest_fp32` per CB and detects conflicts where the same f32 CB is consumed by both FPU and SFPU strategies. Tests were added and updated to cover typecast fusion, mixed dtype rejection, and `unpack_to_dest_fp32` positive, negative, and conflict cases.

### Ticket
https://github.com/tenstorrent/tt-lang/issues/264

### Checklist


python/ttl/ttl_api.py
```diff
@@ -680,6 +688,33 @@ def _get_kernel_bool_attr(module, kernel_name: str, attr_name: str) -> bool:
...
+
+    Returns an empty list when the attribute is missing. Used by the runtime
+    bridge to consume the per-CB UnpackToDestFp32 selection emitted by
+    `ttl-set-compute-kernel-config`.
+    """
+    operation = _lookup_kernel_func_op(module, kernel_name)
+    attr = operation.attributes.get(attr_name, None)
+    if attr is None:
+        return []
+    attr_text = str(attr).strip()
```
arichinsTT: instead of parsing the string why not use existing binding from mlir

```
from ttl.ir import DenseI32ArrayAttr

...
if attr is None:
          return []   
if not DenseI32ArrayAttr.isinstance(attr):                                                                                                                                                                                                   
       raise ValueError(
              f"Expected DenseI32ArrayAttr for '{attr_name}' on kernel "                                                                                                                                                                           
              f"'{kernel_name}', got {attr}"                                                                                                                                                                                                       
        )
return list(DenseI32ArrayAttr(attr))  
````

Comment (vtsilytskyiTT) [reply python/ttl/ttl_api.py]: Thanks, good point. I was thinking about pattern matching or regex. But your suggestion is perfect.

Comment (vtsilytskyiTT) [reply python/ttl/ttl_api.py]: Done

**Comment by vtsilytskyiTT** [comment]
Resolve conflicts.

**Review by zoecarver — APPROVED** [review]
Very cool! LGTM. Thank you, Vlad!

File: lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp [hunk 2]
```diff
@@ -557,7 +589,7 @@ static LogicalResult buildFusedCompute(Operation *sinkOp,
       Value inputTile = tensorToTile[bcastOp.getInput()];
       Value outputTile = body->getArguments().back(); // output block arg
       auto bcastTileOp = createTileOpWithPlaceholderDstIndex<TileBcastOp>(
```
brnorris03: This still uses `outputTileType`, which is the final fused sink type, not necessarily this particular op's result type. That breaks cases like `typecast(bcast(...))`, where the bcast result should keep its own dtype and the later `ttl.tile_typecast` should perform the conversion.

The same applies to the matmul special cases below. These special-case emitters should derive the result tile type from the current source op, like `emitTileOpFor()` does.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -76,19 +82,80 @@ static bool isDstInputTileComputeOp(Operation *op) {
...
+  }
+
+  llvm::SmallVector<int64_t, 2> conflicts;
+  for (int64_t cb : sfpuCBs) {
+    if (fpuCBs.contains(cb)) {
+      conflicts.push_back(cb);
+    }
+  }
+  for (int64_t cb : conflicts) {
+    fpuCBs[cb]->emitWarning()
```
brnorris03: Should this be a pass error instead of easy to miss warning (since could lead to losing f32 precision)?

brnorris03:
Can you maybe add a couple fused testcases like `typecast(matmul)` or `typecast(matmul + a)`?

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: IMHO, can't say this in an error, because application is still correct. Rather programmer's mistake. I changed the warning to error and fixed a few tests.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Fixed. Honestly, this feels like patching. Going through all OP combinations and picking wrong is error-prone. But I don't see how to change this code to be bullet-proof yet.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -177,18 +249,29 @@ struct TTLSetComputeKernelConfigPass
...
+    WalkResult walkResult = funcOp->walk([&](ComputeOp computeOp) {
+      auto cbs = collectF32SFPUInputCBs(computeOp);
+      if (failed(cbs)) {
         return WalkResult::interrupt();
       }
+      for (int64_t cb : *cbs) {
+        unpackFp32CBs.insert(cb);
+      }
       return WalkResult::advance();
     });
```
brnorris03: I would make this conflict detection function-level (not per `ttl.compute` as it is now). `ttl.unpack_to_dest_fp32` is emitted on `func.func` and applied as a per-kernel config, but `collectF32SFPUInputCBs(computeOp)` only checks conflicts inside one `ttl.compute`. If one compute consumes CB1 through FPU and another compute consumes CB1 through SFPU, this still emits `ttl.unpack_to_dest_fp32 = array<i32: 1>`, which configures CB1 as `UnpackToDestMode::UnpackToDestFp32` for the whole kernel, which is not correct. 

You could aggregate both sets across the whole `func.func` before setting the attr:

1. For each `ttl.compute`, collect:
   - f32 CBs consumed by FPU-style ops: reduce, matmul, FPU-eligible add/sub/mul.
   - f32 CBs consumed by SFPU ops that require `UnpackToDestFp32`.
2. Union both sets at `runOnOperation()` scope.
3. Diagnose `intersection(kernelFpuCBs, kernelSfpuCBs)` before setting `ttl.unpack_to_dest_fp32`.
4. Only if the intersection is empty, emit `ttl.unpack_to_dest_fp32 = array<i32: ...>` from the kernel-level SFPU set.
5. Add an invalid lit test where one `ttl.compute` consumes CB1 through `tile_add` and another `ttl.compute` in the same `func.func` consumes CB1 through `tile_exp`.

Otherwise this can emit `ttl.unpack_to_dest_fp32 = array<i32: 1>`, which configures CB1 as `UnpackToDestMode::UnpackToDestFp32` for the whole kernel even though the FPU consumer.

brnorris03 (changes requested):
Thank you for the updates! One major concern inline; some additional minor nits to address if you want:

nit: `test/ttlang/Dialect/TTL/Transforms/SetComputeKernelConfig/unpack_to_dest_fp32_conflict.mlir` is a negative test so its name should be something like `unpack_to_dest_fp32_conflict_invalid.mlir` to match the repo's conventions.

nit: in `python/ttl/ttl_api.py`, `kernel_bool_attrs` now stores both booleans and the `unpack_to_dest_fp32` int array. Rename it to kernel_config_attrs or similar since it's no longer just bools.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
brnorris03: Agreed. I do not think this PR needs to solve the general classification problem.

My concern here was correctness, not the clean long-term design: `ttl.unpack_to_dest_fp32` is a function-level kernel config, so incompatible CB requirements must be checked at function scope before emitting it. The new test covers that.

The op-list classification can stay local for now. If more unpack-mode cases are added later, we should move this to traits or an op interface, but I would not block this PR on that refactor, especially since AFAIK future architectures are dropping FPU support for elementwise ops.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
brnorris03: Yes, but we do want to produce errors for cases where we know that the compiler does NOT support perfectly valid syntax. This is true for many features at this point as we continue to implement them incrementally. So the error can be phrased to be clear that the compiler does not support it without implying logic errors. It's definitely the preferred approach rather than producing an easy to miss warning and incorrect results (or crash/hang). And normally there is an issue the error mentions that is used to track this limitation and eventually remove it.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: Agree with all 3 comments.

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
```diff
@@ -321,40 +321,43 @@ func.func @all_fpu_ops(%a: tensor<2x2x!ttcore.tile<32x32, f32>>,
...
 // CHECK-NEXT:      ttl.yield
 //
-// SFPU path: add also needs copy_tile for both operands
+// SFPU path: add also needs copy_tile for both operands. Both block args for A
+// are SFPU-only consumers, so no conflict diagnostic fires.
 // SFPU-LABEL: func.func @block_arg_fpu_and_sfpu
 // SFPU-DAG:       %[[C0:.*]] = arith.constant 0 : index
 // SFPU-DAG:       %[[C1:.*]] = arith.constant 1 : index
 // SFPU:           ttl.compute
-// SFPU-SAME:      ttl.unroll_factor = 4
```
brnorris03, line 352: Why remove the unroll factor check (perhaps accidental)?

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
```diff
@@ -416,38 +426,42 @@ func.func @block_arg_fpu_and_sfpu(%a: tensor<2x2x!ttcore.tile<32x32, f32>>,
...
+// need a second copy_tile for A.
+// Pattern:
+//   copy a_sfpu->0, abs->0, copy a_fpu->1, copy B->2, add(1,2)->3,
+//   add(0,3)->0 (sum1), mul(1,2)->1, add(0,1)->0 (sum2)
 // SFPU-LABEL: func.func @fpu_binary_no_dst_reuse
 // SFPU-DAG:       %[[C0:.*]] = arith.constant 0 : index
 // SFPU-DAG:       %[[C1:.*]] = arith.constant 1 : index
+// SFPU-DAG:       %[[C3:.*]] = arith.constant 3 : index
 // SFPU:           ttl.compute
-// SFPU-SAME:      ttl.unroll_factor = 2
```
brnorris03, line 443: Another accidental removal of unroll factor check?

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -76,19 +81,64 @@ static bool isDstInputTileComputeOp(Operation *op) {
...
+/// `computeOp` consumed by an FPU-style tile op (reduce, matmul, or
+/// FPU-eligible add/sub/mul). FPU consumers route their operand through
+/// SRCA/SRCB, which is incompatible with `UnpackToDestFp32` mode on the CB.
+static std::optional<int64_t> getF32FPUSrcCBIndex(Operation *op, Value operand,
+                                                  ComputeOp computeOp) {
+  if (!isa<TileReduceOp, TileMatmulBlockOp>(op) && !isFPUEligibleBinaryOp(op)) {
+    return std::nullopt;
+  }
+  return getF32InputCBIndexForBlockArg(operand, computeOp);
+}
```
brnorris03: **[medium] Silent miscompile: an f32 CB shared by an SFPU DST op and a `tile_bcast`/`tile_transpose` gets `UnpackToDestFp32` with no diagnostic.**

`wantsUnpackToDestFp32` correctly stops bcast/transpose from *requesting* the mode (tt-llk #1338), and the new conflict check covers FPU-vs-SFPU — but nothing stops the mode from being *enabled* on a CB that a bcast/transpose also reads when another SFPU op on the same CB requests it. That is the exact #1338 failure the comment above warns about.

Repro (`ttl-set-compute-kernel-config`): an f32 CB read by both `tile_exp` and `tile_bcast` emits `ttl.unpack_to_dest_fp32 = array<i32: 1>` with no error, corrupting the bcast. Fix: treat bcast/transpose f32-CB consumers as "must stay Default" so the existing conflict check fires. Self-contained — compiles alone, existing tests unchanged, verified locally:

```suggestion
/// Return the CB index when `value` is an f32 input block argument of
/// `computeOp` consumed by a tile op that must keep its source CB in `Default`
/// unpack mode. FPU-style ops (reduce, matmul, FPU-eligible add/sub/mul) route
/// the operand through SRCA/SRCB, and `tile_bcast` / `tile_transpose` lower to
/// LLK paths (unary_bcast, transpose_dest) that produce incorrect results under
/// `UnpackToDestFp32` on their source CB (tt-llk #1338). All are incompatible
/// with `UnpackToDestFp32`.
static std::optional<int64_t> getF32FPUSrcCBIndex(Operation *op, Value operand,
                                                  ComputeOp computeOp) {
  if (!isa<TileReduceOp, TileMatmulBlockOp, TileBcastOp, TileTransposeOp>(op) &&
      !isFPUEligibleBinaryOp(op)) {
    return std::nullopt;
  }
  return getF32InputCBIndexForBlockArg(operand, computeOp);
}
```

Optional follow-up (clearer, but touches the struct field `fpuCBConsumers`, the kernel-level map, and the two `expected-error` strings): rename `getF32FPUSrcCBIndex`->`getF32DefaultUnpackCBIndex` / `fpuCBConsumers`->`defaultUnpackCBConsumers` and generalize the message from "both FPU and SFPU strategies" to "both a default-unpack consumer (FPU or tile_bcast/tile_transpose) and an SFPU consumer". I verified that variant too.

**Regression tests (both verified against the fix):**

Append to `unpack_to_dest_fp32_conflict_invalid.mlir`:
```mlir
// -----

#map = affine_map<(d0, d1) -> (d0, d1)>

// Purpose: tile_bcast / tile_transpose must keep their source CB in Default
// unpack mode (tt-llk #1338). When an f32 CB feeds both an SFPU DST op
// (tile_exp) and a tile_bcast, the pass must diagnose the conflict instead of
// silently enabling UnpackToDestFp32 and corrupting the bcast.
func.func @f32_cb1_used_by_bcast_and_sfpu(
    %a: tensor<1x1x!ttcore.tile<32x32, f32>>,
    %b: tensor<1x1x!ttcore.tile<32x32, f32>>)
    -> tensor<1x1x!ttcore.tile<32x32, f32>>
    attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
  %init = tensor.empty() : tensor<1x1x!ttcore.tile<32x32, f32>>
  %cb0 = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb1 = ttl.bind_cb {cb_index = 1, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb2 = ttl.bind_cb {cb_index = 2, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %a_cb = ttl.attach_cb %a, %cb0 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %b_cb_sfpu = ttl.attach_cb %b, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %b_cb_bcast = ttl.attach_cb %b, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %init_cb = ttl.attach_cb %init, %cb2 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %out_view = ttl.cb_reserve %cb2 : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %res = ttl.compute
      ins(%a_cb, %b_cb_sfpu, %b_cb_bcast : tensor<1x1x!ttcore.tile<32x32, f32>>, tensor<1x1x!ttcore.tile<32x32, f32>>, tensor<1x1x!ttcore.tile<32x32, f32>>)
      outs(%init_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      {indexing_maps = [#map, #map, #map, #map], iterator_types = ["parallel", "parallel"]} {
    ^bb0(%a_tile: !ttcore.tile<32x32, f32>, %b_sfpu: !ttcore.tile<32x32, f32>, %b_bcast: !ttcore.tile<32x32, f32>, %out_tile: !ttcore.tile<32x32, f32>):
      %i = ttl.iter_index 0 : index
      %j = ttl.iter_index 1 : index
      %ex = ttl.tile_exp %b_sfpu into dst[%c0] : !ttcore.tile<32x32, f32> -> !ttcore.tile<32x32, f32>
      // expected-error @below {{f32 input from CB 1 is consumed by both FPU and SFPU strategies in the same kernel}}
      %bc = ttl.tile_bcast %b_bcast, %out_tile 2 : i32 into dst[%c0] : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
      ttl.tile_store %ex, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.tile_store %bc, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.yield
  } -> tensor<1x1x!ttcore.tile<32x32, f32>>
  return %res : tensor<1x1x!ttcore.tile<32x32, f32>>
}
```

Append to `unpack_to_dest_fp32_negative.mlir`:
```mlir
// -----

#map = affine_map<(d0, d1) -> (d0, d1)>

// Purpose: tile_bcast consuming an f32 CB must NOT enable UnpackToDestFp32
// (tt-llk #1338); with no SFPU consumer on that CB there is also no conflict.
// CHECK-LABEL: func.func @bcast_only_f32_does_not_set_unpack
func.func @bcast_only_f32_does_not_set_unpack(
    %a: tensor<1x1x!ttcore.tile<32x32, f32>>)
    -> tensor<1x1x!ttcore.tile<32x32, f32>>
    attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
  %init = tensor.empty() : tensor<1x1x!ttcore.tile<32x32, f32>>
  %cb0 = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb1 = ttl.bind_cb {cb_index = 1, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %a_cb = ttl.attach_cb %a, %cb0 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %init_cb = ttl.attach_cb %init, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %out_view = ttl.cb_reserve %cb1 : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %res = ttl.compute
      ins(%a_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      outs(%init_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} {
    ^bb0(%a_tile: !ttcore.tile<32x32, f32>, %out_tile: !ttcore.tile<32x32, f32>):
      %i = ttl.iter_index 0 : index
      %j = ttl.iter_index 1 : index
      %bc = ttl.tile_bcast %a_tile, %out_tile 2 : i32 into dst[%c0] : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
      ttl.tile_store %bc, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.yield
  } -> tensor<1x1x!ttcore.tile<32x32, f32>>
  return %res : tensor<1x1x!ttcore.tile<32x32, f32>>
}
```

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
```diff
@@ -321,40 +321,43 @@ func.func @all_fpu_ops(%a: tensor<2x2x!ttcore.tile<32x32, f32>>,
...
 // CHECK:           ttl.tile_store %[[EXP]], %{{.*}}[%[[I0]], %[[I1]]]
 // CHECK-NEXT:      ttl.yield
 //
-// SFPU path: add also needs copy_tile for both operands
+// SFPU path: add also needs copy_tile for both operands. Both block args for A
+// are SFPU-only consumers, so no conflict diagnostic fires.
 // SFPU-LABEL: func.func @block_arg_fpu_and_sfpu
 // SFPU-DAG:       %[[C0:.*]] = arith.constant 0 : index
 // SFPU-DAG:       %[[C1:.*]] = arith.constant 1 : index
 // SFPU:           ttl.compute
```
brnorris03: [low] The `SFPU` prefix dropped its `// SFPU-SAME: ttl.unroll_factor = 4` assertion. I ran the SFPU pipeline on the new IR and the factor is still 4, so the check should be restored rather than deleted (otherwise the SFPU run no longer asserts the unroll factor at all):
```suggestion
// SFPU:           ttl.compute
// SFPU-SAME:      ttl.unroll_factor = 4
```

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -177,18 +227,46 @@ struct TTLSetComputeKernelConfigPass
...
-      if (needsUnpackToDestFp32(computeOp)) {
-        needsUnpackFp32 = true;
-        return WalkResult::interrupt();
+      const F32InputCBUsage usage = collectF32InputCBUsage(computeOp);
+      kernelSFPUCBs.insert_range(usage.sfpuCBs);
+      for (auto [cb, consumer] : usage.fpuCBConsumers) {
+        // Keep the first FPU consumer for stable diagnostics when multiple
+        // compute regions consume the same CB through the FPU path.
+        kernelFPUCBConsumers.insert({cb, consumer});
       }
```
brnorris03: [low] Nondeterministic consumer kept for the diagnostic. `usage.fpuCBConsumers` is a `SmallDenseMap`, whose iteration order is unspecified, so when the same conflicting CB has FPU consumers in more than one compute region the op the error attaches to is nondeterministic. The message text is identical today, so it does not flake, but a future multi-consumer `--verify-diagnostics` test could. Consider a deterministic tiebreak (e.g. keep the consumer that `isBeforeInBlock`, or iterate computes/ops in IR order and only `insert` the first seen).

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -217,10 +217,29 @@ static void emitTileStores(PatternRewriter &rewriter, Location loc,
...
+/// op's tensor result so dtype-changing ops such as `ttl.typecast` produce
+/// correctly-typed intermediates inside a fused chain.
 static Value emitTileOpFor(OpBuilder &builder, Location loc,
-                           Operation *sourceOp, ValueRange tileOperands,
-                           Type tileType) {
+                           Operation *sourceOp, ValueRange tileOperands) {
+  Type tileType = getFusedResultTileType(sourceOp);
+  if (!tileType) {
+    return nullptr;
+  }
```
brnorris03: [low] This `!tileType` guard is dead. The sole caller already checks `op->getNumResults() != 1` and `isa<RankedTensorType>(op->getResult(0).getType())` before calling `emitTileOpFor`, so `getFusedResultTileType` here can never return null, and it is recomputed after the caller already validated. Either drop this guard, or drop the caller's pre-checks and keep this as the single guard.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
brnorris03, line 109: Incorrect compilation when an f32 CB consumed by both an SFPU DST op and `tile_bcast`/`tile_transpose` is configured with `UnpackToDestFp32` and no diagnostic is emitted.

`getF32FPUSrcCBIndex` records only reduce/matmul/FPU-binary consumers as requiring `Default` unpack mode, so the conflict loop in `runOnOperation` never compares the SFPU set against `tile_bcast`/`tile_transpose`. When one f32 CB feeds both `tile_exp` (SFPU, added to `sfpuCBs`) and `tile_bcast` (recorded nowhere), the pass sets `ttl.unpack_to_dest_fp32` for that CB. Per tt-llk #1338 `unary_bcast`/`transpose_dest` produce incorrect results under that mode, so the bcast/transpose miscompiles — the failure the `wantsUnpackToDestFp32` comment above documents for the request side but not the conflict side.

Reproduced with `ttl-set-compute-kernel-config`: one f32 CB read by `tile_exp` and `tile_bcast` yields `ttl.unpack_to_dest_fp32 = array<i32: 1>` and no error.

Proposed fix:
```cpp
/// Return the CB index when `value` is an f32 input block argument of
/// `computeOp` consumed by a tile op that must keep its source CB in `Default`
/// unpack mode. FPU-style ops (reduce, matmul, FPU-eligible add/sub/mul) route
/// the operand through SRCA/SRCB; `tile_bcast`/`tile_transpose` lower to
/// `unary_bcast`/`transpose_dest`, which produce incorrect results under
/// `UnpackToDestFp32` on their source CB (tt-llk #1338). Both are incompatible
/// with the mode.
static std::optional<int64_t>
getF32DefaultUnpackCBIndex(Operation *op, Value operand, ComputeOp computeOp) {
  if (!isa<TileReduceOp, TileMatmulBlockOp, TileBcastOp, TileTransposeOp>(op) &&
      !isFPUEligibleBinaryOp(op)) {
    return std::nullopt;
  }
  return getF32InputCBIndexForBlockArg(operand, computeOp);
}
```

Struct field `fpuCBConsumers` -> `defaultUnpackCBConsumers`; the call site in `collectF32InputCBUsage` then uses `getF32DefaultUnpackCBIndex` and `usage.defaultUnpackCBConsumers`. In `runOnOperation`, rename `kernelFPUCBConsumers` -> `kernelDefaultUnpackCBConsumers` and `fpuConsumer` -> `defaultConsumer`, and replace the diagnostic with:
```cpp
      defaultConsumer->emitOpError()
          << "f32 input from CB " << cb
          << " is consumed both by a tile op that requires default unpack "
             "mode (FPU reduce/matmul/add/sub/mul, or tile_bcast/"
             "tile_transpose) and by an SFPU op that needs UnpackToDestFp32; "
             "the two modes are mutually exclusive on a given CB. Split the "
             "source into separate CBs (one per strategy) so the SFPU consumer "
             "keeps full f32 precision";
```

Both `expected-error` strings in `unpack_to_dest_fp32_conflict_invalid.mlir` then update to `f32 input from CB 1 is consumed both by a tile op that requires default unpack mode`.

Regression tests (verified with above fix):

Append to `unpack_to_dest_fp32_conflict_invalid.mlir`:
```mlir
// -----

#map = affine_map<(d0, d1) -> (d0, d1)>

// Purpose: tile_bcast / tile_transpose must keep their source CB in Default
// unpack mode (tt-llk #1338). When an f32 CB feeds both an SFPU DST op
// (tile_exp) and a tile_bcast, the conflict is diagnosed rather than silently
// configuring UnpackToDestFp32 and corrupting the bcast.
func.func @f32_cb1_used_by_bcast_and_sfpu(
    %a: tensor<1x1x!ttcore.tile<32x32, f32>>,
    %b: tensor<1x1x!ttcore.tile<32x32, f32>>)
    -> tensor<1x1x!ttcore.tile<32x32, f32>>
    attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
  %init = tensor.empty() : tensor<1x1x!ttcore.tile<32x32, f32>>
  %cb0 = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb1 = ttl.bind_cb {cb_index = 1, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb2 = ttl.bind_cb {cb_index = 2, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %a_cb = ttl.attach_cb %a, %cb0 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %b_cb_sfpu = ttl.attach_cb %b, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %b_cb_bcast = ttl.attach_cb %b, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %init_cb = ttl.attach_cb %init, %cb2 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %out_view = ttl.cb_reserve %cb2 : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %res = ttl.compute
      ins(%a_cb, %b_cb_sfpu, %b_cb_bcast : tensor<1x1x!ttcore.tile<32x32, f32>>, tensor<1x1x!ttcore.tile<32x32, f32>>, tensor<1x1x!ttcore.tile<32x32, f32>>)
      outs(%init_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      {indexing_maps = [#map, #map, #map, #map], iterator_types = ["parallel", "parallel"]} {
    ^bb0(%a_tile: !ttcore.tile<32x32, f32>, %b_sfpu: !ttcore.tile<32x32, f32>, %b_bcast: !ttcore.tile<32x32, f32>, %out_tile: !ttcore.tile<32x32, f32>):
      %i = ttl.iter_index 0 : index
      %j = ttl.iter_index 1 : index
      %ex = ttl.tile_exp %b_sfpu into dst[%c0] : !ttcore.tile<32x32, f32> -> !ttcore.tile<32x32, f32>
      // expected-error @below {{f32 input from CB 1 is consumed both by a tile op that requires default unpack mode}}
      %bc = ttl.tile_bcast %b_bcast, %out_tile 2 : i32 into dst[%c0] : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
      ttl.tile_store %ex, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.tile_store %bc, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.yield
  } -> tensor<1x1x!ttcore.tile<32x32, f32>>
  return %res : tensor<1x1x!ttcore.tile<32x32, f32>>
}
```

Append to `unpack_to_dest_fp32_negative.mlir`:
```mlir
// -----

#map = affine_map<(d0, d1) -> (d0, d1)>

// Purpose: tile_bcast consuming an f32 CB must NOT enable UnpackToDestFp32
// (tt-llk #1338); with no SFPU consumer on that CB there is also no conflict.
// CHECK-LABEL: func.func @bcast_only_f32_does_not_set_unpack
func.func @bcast_only_f32_does_not_set_unpack(
    %a: tensor<1x1x!ttcore.tile<32x32, f32>>)
    -> tensor<1x1x!ttcore.tile<32x32, f32>>
    attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
  %init = tensor.empty() : tensor<1x1x!ttcore.tile<32x32, f32>>
  %cb0 = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %cb1 = ttl.bind_cb {cb_index = 1, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
  %a_cb = ttl.attach_cb %a, %cb0 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %init_cb = ttl.attach_cb %init, %cb1 : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %out_view = ttl.cb_reserve %cb1 : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32, f32>>
  %res = ttl.compute
      ins(%a_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      outs(%init_cb : tensor<1x1x!ttcore.tile<32x32, f32>>)
      {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} {
    ^bb0(%a_tile: !ttcore.tile<32x32, f32>, %out_tile: !ttcore.tile<32x32, f32>):
      %i = ttl.iter_index 0 : index
      %j = ttl.iter_index 1 : index
      %bc = ttl.tile_bcast %a_tile, %out_tile 2 : i32 into dst[%c0] : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
      ttl.tile_store %bc, %out_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, f32>, tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.yield
  } -> tensor<1x1x!ttcore.tile<32x32, f32>>
  return %res : tensor<1x1x!ttcore.tile<32x32, f32>>
}
```

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
brnorris03, line 355: The `SFPU` prefix dropped its `// SFPU-SAME: ttl.unroll_factor = 4` assertion. I ran the SFPU pipeline on the new IR and the factor is still 4.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
brnorris03, line 245: `usage.fpuCBConsumers` is a `SmallDenseMap`, whose iteration order is unspecified, so when the same conflicting CB has FPU consumers in more than one compute region the op the error attaches to is nondeterministic. `MapVector` is a deterministic alternative.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
brnorris03: nit: I think this `!tileType` guard is dead code.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -659,15 +711,20 @@ static LogicalResult buildFusedCompute(Operation *sinkOp,
...
-      // tile_matmul_block so the elementwise path can resolve them.
+      // tile_matmul_block so the elementwise path can resolve them. Each
+      // emitted matmul keeps the dtype of its own tensor result (from the
+      // map key) so that intermediates do not get the final sink dtype.
       if (!tileResult) {
         for (Value operand : getElementwiseOperands(op)) {
           auto dfIt = deferredMatmul.find(operand);
           if (dfIt != deferredMatmul.end()) {
             auto [mmLhs, mmRhs] = dfIt->second;
+            auto matmulTensor = cast<RankedTensorType>(operand.getType());
```
brnorris03: nit: somewhat inconsistent casting -- `cast<RankedTensorType>(operand.getType())` asserts on failure, while the new `getFusedResultTileType` uses `dyn_cast` + null-return for the same query.

brnorris03 (approved):
Just one more significant issue with a proposed fix and extra tests, otherwise minor things, mostly optional. Thank you for all the improvements!

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
vtsilytskyiTT: Yes, reverted

test/ttlang/Dialect/TTL/Transforms/AssignDST/dst_fpu_binary.mlir
vtsilytskyiTT: Yes, reverted

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: Thanks for catching the problem and providing solution. This again feels like patching. I have to remember all OPs, that should be listed in `isa<>`. Error-prone and not scalable :(

brnorris03: Yeah, the fact that it was missed is proof of that. Should look for better long-term solution. But our target "ISA" is extremely messy now, so as you see it's a challenge to classify ops in a more scalable way (not impossible I'm sure).

brnorris03: We should brainstorm at a higher level of how to encode at a high level (in the dialect) the differences in f32 support for different kinds of ops, I don't think a simple trait would be sufficient.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Yes. I'll remove it, though assertion is also an option.

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
vtsilytskyiTT: Agree, order matters for error messages.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Actually, I can reuse `getFusedResultTileType` function here.

PR #603, opened 2026-05-19
element read/write ttl ops and ast

arichinsTT:
### Problem description

Data movement threads have no way to read or write individual scalar
elements within a DFB slot. The existing DFB interface operates on whole
tiles/blocks, but use cases like KV-cache updates, top-K, and
element-level data manipulation in DM threads require per-element L1
access.

### What's changed

**New `ttl.raw_element_read` / `ttl.raw_element_write` ops** -- ODS
definitions in `TTLOps.td` with variadic index coordinates, supporting
any block rank (1D, 2D, 3D, ...) and f32/bf16 dtypes. A shared C++
verifier in `TTLOps.cpp` enforces: noc-thread-only usage, coordinate
count == block rank, scalar type matches block element dtype, and f32/bf16
restriction. Python DSL wrappers in `operators.py` auto-coerce int
literals to `arith.constant` index values and dispatch to the generated
bindings.

**Design documentation** in `docs/development/DFBManagement.md` covering
motivation, Python DSL interface, MLIR op specs, verifier rules,
coordinate semantics, DFB lifecycle interaction, and planned lowering
approach.

### Checklist


docs/development/DFBManagement.md
```diff
@@ -520,3 +520,166 @@ Index reuse is restricted to compiler-allocated DFBs. User-declared DFBs retain
...
+### Test coverage
+
+- `test/ttlang/Dialect/TTL/IR/raw_element_ops.mlir` -- verifier acceptance
+  tests for f32/bf16 read and write on 2D and 3D tiled blocks.
+- `test/ttlang/Dialect/TTL/IR/raw_element_ops_invalid.mlir` -- verifier
+  rejection tests covering: missing kernel thread attribute, compute thread
+  usage, coordinate count mismatch, and scalar dtype mismatch.
+- `test/python/simple_raw_element_access.py` -- compile-only Python DSL
+  test verifying the frontend lowers `ttl.raw_element_read` and
+  `ttl.raw_element_write` into the expected TTL ops.
```
brnorris03: I wouldn't put test enumeration in a design doc. Also could abbreviate all of the above into a single section

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1345,6 +1345,74 @@ def TTL_StoreOp : TTL_Op<"store", [MemoryEffects<[MemWrite]>]> {
...
+    ```mlir
+    %val = ttl.raw_element_read %block[%i, %j]
+           : tensor<1x1x!ttcore.tile<32x32, f32>> -> f32
+    ```
+  }];
+  let arguments = (ins
+    AnyRankedTensor:$block,
+    Variadic<Index>:$coords
+  );
+  let results = (outs AnyType:$result);
```
brnorris03: Could tighten `$result` (and `$value` on the write op) to `AnyTypeOf<[F32, BF16]>` (that eliminates the need for dtype checks in the verifier)

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -1315,6 +1315,79 @@ mlir::LogicalResult mlir::tt::ttl::CreatePipeOp::verify() {
...
+    return op->emitOpError()
+           << "scalar type (" << scalarTy
+           << ") must match block element dtype (" << expectedScalarTy << ")";
+  }
+
+  // 4. Only f32 and bf16 are supported.
+  if (!scalarTy.isF32() && !scalarTy.isBF16()) {
+    return op->emitOpError()
+           << "only f32 and bf16 element types are supported, got " << scalarTy;
+  }
```
brnorris03: For non-f32/bf16 tile dtypes, `dataTypeToElementType` returns a TileType, so check #3 reports a confusing scalar-vs-tile mismatch before check #4 -- perhaps move #4 before #3?

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1345,6 +1345,74 @@ def TTL_StoreOp : TTL_Op<"store", [MemoryEffects<[MemWrite]>]> {
...
+//===----------------------------------------------------------------------===//
+// Raw element access operations (data movement threads only)
+//===----------------------------------------------------------------------===//
+
+def TTL_RawElementReadOp : TTL_Op<"raw_element_read",
+    [MemoryEffects<[MemRead]>, TTL_DataMovementOpTrait]> {
+  let summary = "Read a scalar element from a block at flat coordinates";
+  let description = [{
+    `ttl.raw_element_read` extracts a single scalar element from a block
+    (a tensor view acquired from `ttl.cb_wait` or `ttl.cb_reserve`) at
```
brnorris03, line 1390: I don't think the verifier actually checks that this is a "a tensor view acquired from `ttl.cb_wait` or `ttl.cb_reserve`". Perhaps add to `verifyRawElementOp`? Can use `getAttachedCB` from `TTLOpsUtils.h`

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -1315,6 +1315,79 @@ mlir::LogicalResult mlir::tt::ttl::CreatePipeOp::verify() {
...
+/// Shared verification for raw_element_read and raw_element_write. Checks:
+///   1. Enclosing function is a data movement (noc) kernel thread.
+///   2. Coordinate count matches block tensor rank.
+///   3. Scalar type matches block's underlying element dtype.
+///   4. Only f32 and bf16 are supported.
+static mlir::LogicalResult verifyRawElementOp(mlir::Operation *op,
+                                              mlir::RankedTensorType blockTy,
+                                              mlir::ValueRange coords,
+                                              mlir::Type scalarTy) {
+  using namespace mlir::tt;
```
brnorris03: nit: I think it's best to avoid adding `using namespace` it's not saving much typing and it is clearer to have the full namespaces visible (also more consistent, since there are mlir:: namespaced things in this function).

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -180,8 +180,9 @@ LogicalResult lowerCBToPipe(CopyOp op, Value srcCB, Value pipe,
         rewriter, loc, getSenderSemIdx(pipeType));
     auto senderSemAddr =
         ttk::GetSemaphoreOp::create(rewriter, loc, senderSemIdx);
-    auto senderSemPtr =
-        ttk::CastToL1PtrOp::create(rewriter, loc, senderSemAddr);
+    auto senderSemPtr = ttk::CastToL1PtrOp::create(
+        rewriter, loc, ttk::L1AddrPtrType::get(rewriter.getContext(), 32),
+        senderSemAddr);
     auto expectedVal = arith::ConstantOp::create(
```
brnorris03, line 186: I guess this is the new tt-mlir change? I was a bit confused initially why it's touching PipeLowering.cpp for unrelated ops. Anyway `elementWidth` defaults to 32 in `TTKernel_L1AddrPtr, so `L1AddrPtrType::get(rewriter.getContext())` would be equivalent (and a bit shorter).

python/ttl/operators.py
```diff
@@ -742,6 +742,109 @@ def typecast(input: TensorBlock, dtype) -> TensorBlock:
...
+    Example:
+        ttl.raw_element_write(block, row, col, val)
+    """
+    from ttl.dialects import arith
+
+    if len(args) < 2:
+        raise ValueError(
+            "raw_element_write requires at least one coordinate and a value"
+        )
+    coord_args = args[:-1]
```
brnorris03, line 850: Why this particular constraint? The MLIR op would accept rank-0 now. If it shouldn't, I think this check is better to have in the op verifier, not here.

Also, there is no equivalent check in raw_element_read (which is not a problem if the MLIR op verifier does the checking).

python/ttl/operators.py
```diff
@@ -742,6 +742,109 @@ def typecast(input: TensorBlock, dtype) -> TensorBlock:
...
+    Only supported in data movement (noc) threads.
+
+    Args:
+        block: Block tensor (from cb.reserve() or cb.wait())
+        *coords: Index values matching the block tensor rank
+
+    Returns:
+        Scalar value matching the block's element dtype
+    """
+    from ttl.dialects import arith
```
brnorris03: remove, already imported at module scope

python/ttl/operators.py
```diff
@@ -742,6 +742,109 @@ def typecast(input: TensorBlock, dtype) -> TensorBlock:
...
+    Only supported in data movement (noc) threads.
+
+    Args:
+        block: Block tensor (from cb.reserve() or cb.wait())
+        *args: N index values followed by the scalar value to write.
+
+    Example:
+        ttl.raw_element_write(block, row, col, val)
+    """
+    from ttl.dialects import arith
```
brnorris03: remove

test/ttlang/Dialect/TTL/IR/raw_element_ops.mlir
```diff
@@ -0,0 +1,75 @@
...
+// CHECK: %[[VAL:.*]] = ttl.raw_element_read %{{.*}}[%{{.*}}, %{{.*}}, %{{.*}}] : tensor<2x3x4x!ttcore.tile<32x32, f32>> -> f32
+func.func @raw_element_read_3d(
+    %block: tensor<2x3x4x!ttcore.tile<32x32, f32>>)
+    attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
+  %c0 = arith.constant 0 : index
+  %c1 = arith.constant 1 : index
+  %c2 = arith.constant 2 : index
+  %val = ttl.raw_element_read %block[%c0, %c1, %c2] : tensor<2x3x4x!ttcore.tile<32x32, f32>> -> f32
+  func.return
+}
```
brnorris03, line 83: Suggest adding these cases that are currently allowed by the verifier:
- a row-major (non-tile) block case
- a rank-1 case

test/ttlang/Dialect/TTL/IR/raw_element_ops_invalid.mlir
```diff
@@ -0,0 +1,96 @@
...
+
+// raw_element_write with scalar type mismatch (f32 value into bf16 block).
+func.func @write_dtype_mismatch(
+    %block: tensor<1x1x!ttcore.tile<32x32, bf16>>, %val: f32)
+    attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
+  %c0 = arith.constant 0 : index
+  // expected-error @below {{scalar type ('f32') must match block element dtype ('bf16')}}
+  ttl.raw_element_write %block[%c0, %c0], %val : tensor<1x1x!ttcore.tile<32x32, bf16>>, f32
+  func.return
+}
```
brnorris03, line 124: Suggest adding a test with a tile with an unsupported dtype (e.g., i32) -- right now I don't think that verifier check is covered.

test/python/simple_raw_element_access.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
brnorris03: I would postpone python lit tests (which are typically runtime tests with full-pipeline lit checks) until the next PR that has the lowering.

brnorris03 (commented):
Looks very good overall, a few minor things here and there, some test suggestions. I would not include the e2e test in this PR (simple_raw_element_access.py).

lib/Dialect/TTL/IR/TTLOps.cpp
```diff
@@ -1341,6 +1341,87 @@ mlir::LogicalResult mlir::tt::ttl::CreatePipeOp::verify() {
...
+      mlir::tt::ttl::kKernelThreadAttrName);
+  if (!threadAttr ||
+      threadAttr.getValue() != mlir::tt::ttkernel::ThreadType::Noc) {
+    return op->emitOpError()
+           << "is only allowed in data movement (noc) threads";
+  }
+
+  // 2. Block must originate from a circular buffer (ttl.cb_wait / cb_reserve).
+  if (!mlir::tt::ttl::getAttachedCB(block)) {
+    return op->emitOpError() << "block must be a tensor view acquired from "
```
brnorris03, line 1376: This is a bit looser than the documented constraint (would also accept results of `ttl.attach_cb`, `tensor.extract`, and `tensor.extract_slice`), so may be good to add some more checks. Couuld create a separate stricter helper that accepts only `CBWaitOp`/`CBReserveOp` as the source, plus an invalid test for `ttl.attach_cb` (I may have suggested getAttachedCB previously but realize now it's not quite right for this situation.)

python/ttl/operators.py
```diff
@@ -8,8 +8,9 @@
 
 from typing import List, Tuple, Union
 
-from ttl.dialects import arith
-from ttl.ir import RankedTensorType, Type, FloatAttr, F32Type
+
+from ttl.dialects import arith, arith
```
brnorris03: Remove the duplicate `arith` import.

python/ttl/operators.py
```diff
@@ -755,6 +756,112 @@ def typecast(input: TensorBlock, dtype) -> TensorBlock:
...
+        return elem_type
+    if elem_type == BF16Type.get(block_type.context):
+        return elem_type
+    raise ValueError(
+        f"raw element access only supports f32 and bf16, got element type {elem_type}"
+    )
+
+
+@syntax("raw_element_read")
+def raw_element_read(block, *coords):
```
brnorris03, line 794: Up to you, but since we don't have lowering yet, I wouldn't add these yet to python -- someone could try using them in an actual program. You could have this change in the following PR that does the lowering.

brnorris03 (approved):
A couple minor comments, generally good to go (the verifier is the only more significant one that would be good to get exactly right).

python/ttl/operators.py
arichinsTT: I think I want to have them there to show how the ops work in the language.  The lowering will come out soon

PR #604, opened 2026-05-19
Simulator: Remove string IDs

FConstantinos:
Remove core ids that were based on strings

### DIFF
--- a/python/sim/greenlet_scheduler.py
+++ b/python/sim/greenlet_scheduler.py
@@ -9,6 +9,7 @@ yield transformations. Each compute or datamovement kernel runs in its own green
 and blocking operations (wait/reserve) switch back to the scheduler.
 """
 
+from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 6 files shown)


arichinsTT (approved):
LGTM

python/sim/greenlet_scheduler.py
```diff
@@ -20,11 +21,38 @@
...
+
+    ``linear_core`` is the linear core index used by ``program`` when registering
+    threads. ``suffix`` is the kernel template name (e.g. ``compute``, ``dm0``).
+
+    User-visible strings match the historical format ``core{linear_core}-{suffix}``;
+    use :func:`kernel_thread_display_name` for traces and diagnostics.
+    """
+
+    linear_core: int
+    suffix: str
```
phizalev-TT, line 40: Should suffix be an enumeration if there is specific set of valid values?

FConstantinos: Addressed here:

https://github.com/tenstorrent/tt-lang/pull/609

PR #610, opened 2026-05-20
Simulator: Implement spec v0.17 API changes 

FConstantinos:
- Add ttl.block namespace (python/sim/block.py) with broadcast, fill, mask, mask_posinf, where, squeeze, unsqueeze, transpose moved from ttl.math
- ttl.math retains only math/reduction functions (reduce_sum, reduce_max, unary ops, etc.)
- squeeze/unsqueeze now accept only dims: List[int] (drop single dim: int form)
- broadcast/reduce/fill signatures updated: explicit dims and shape args, output_hint and scaler params removed
- Add examples/eltwise_broadcast_reduce.py (new spec example)
- Update matmul.py to block-sized loops with unsqueeze+broadcast pattern
- Update all examples and simulator tests to new API


python/sim/block.py
```diff
@@ -0,0 +1,408 @@
...
+    # indexing below treats the last two dims of ``shape`` as the within-tile
+    # axes, so this function only handles blocks with at least two grid
+    # dimensions.  1-D blocks should be unsqueezed by the caller via
+    # ``ttl.block.unsqueeze`` before being broadcast.
+    elem = block._buf.to_torch()  # type: ignore[attr-defined]
+
+    batch = block_shape[:-2]
+    TM_s, TK_s = block_shape[-2], block_shape[-1]
+    tile_h = elem.shape[-2] // TM_s if TM_s > 0 else 1
+    tile_w = elem.shape[-1] // TK_s if TK_s > 0 else 1
```
brnorris03: the comment says "1-D blocks should be unsqueezed" but nothing here verifies that they were? Maybe check for `if ndim < 2` and raise an error?

python/sim/block.py
```diff
@@ -0,0 +1,408 @@
...
+    if block.layout == ROW_MAJOR_LAYOUT:
+        raise ValueError("broadcast is not supported for Row-Major layout blocks")
+
+    if len(shape) != ndim:
+        raise ValueError(
+            f"broadcast shape {shape} has {len(shape)} dimensions but block has {ndim}"
+        )
+
+    # Normalize dims and validate that block has grid size 1 in each broadcast dim.
+    internal_dims = {d % ndim for d in dims}
```
brnorris03, line 145: sim accepts emtpy dims, but compiler produces an error -- I think it should be an error, right?

python/sim/block.py
```diff
@@ -0,0 +1,408 @@
...
+def _apply_binary_op(
+    a: Block, b: Block, op: Callable[[torch.Tensor, torch.Tensor], torch.Tensor]
+) -> Block:
+    a_shape = a._shape  # type: ignore[attr-defined]
+    b_shape = b._shape  # type: ignore[attr-defined]
+    if a_shape != b_shape:
+        raise ValueError(
+            f"Shape mismatch in binary op: a has shape {a_shape}, b has shape {b_shape}"
+        )
+    layout = a.layout
```
brnorris03, line 38: what about b (and c for ternary below) layouts? What if they don't match?

python/sim/block.py
FConstantinos: done

python/sim/block.py
FConstantinos: yes, fixed

python/sim/block.py
FConstantinos: I have added layout checks throughout all operations

PR #614, opened 2026-05-21
[ttl] Pipe verification; implement recv post -> send post -> send wait -> recv wait protocol (#608)

### LINKED ISSUE #608: [bug] PipeLowering sender-lockstep reserve_back on pipe dst CB hangs under iteration
Under Python `for`-loop iteration, a unicast pipe whose `if_src` and `if_dst` are role-gated by `node_x` in the same `@ttl.datamovement` thread hangs on hardware after a couple iterations. Tested on v1.1.1 / blackhole.

## Repro

`N_ITERS=20`, `block_count=2`. Hangs at dispatch on real HW.

```python

"""Minimal repro: under Python for-loop iteration, a unicast Pipe
whose if_src and if_dst live in the same @ttl.datamovement thread
(role-gated by node_x) hangs after a couple of iterations on hardware.

Shape: 2x1 grid. Sender at (1,0) stages one tile in send_cb and
forwards via the pipe; receiver at (0,0) reserves in_cb in if_dst
and the compute thread copies in_cb -> out_cb. Both if_src and
if_dst are gated by node_x in the same NCRISC thread.

block_count=2 means the sender exhausts its local in_cb counter on
the third iteration.
"""

import torch
import ttnn
import ttl


N_ITERS = 20


@ttl.operation(grid=(2, 1))
def unicast_loop(inp, out):
    net = ttl.PipeNet([ttl.Pipe(src=(1, 0), dst=(0, 0))])
    send_cb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=2)
    in_cb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=2)

    @ttl.compute()
    def compute():
        x, _ = ttl.node(dims=2)
        if x == 0:
            for _ in range(N_ITERS):
                with in_cb.wait() as t, out_cb.reserve() as o:
                    o.store(t)

    @ttl.datamovement()
    def dm():
        x, _ = ttl.node(dims=2)
        for _ in range(N_ITERS):
            if x == 1:
                with send_cb.reserve() as blk:
                    ttl.copy(inp[0, 0], blk).wait()
                with send_cb.wait() as blk:
                    def send(pipe):
                        ttl.copy(blk, pipe).wait()
                    net.if_src(send)
            if x == 0:
                def recv(pipe):
                    with in_cb.reserve() as blk:
                        ttl.copy(pipe, blk).wait()
                net.if_dst(recv)
                with out_cb.wait() as blk:
                    ttl.copy(blk, out[0, 0]).wait()

    @ttl.datamovement()
    def dm_brisc():
        pass


if __name__ == "__main__":
    try:
        inp = ttnn.from_torch(
            torch.randn(32, 32, dtype=torch.bfloat16),
            device=device, memory_config=ttnn.DRAM_MEMORY_CONFIG,
        )
        out = ttnn.from_torch(
            torch.zeros(32, 32, dtype=torch.bfloat16),
            device=device, memory_config=ttnn.DRAM_MEMORY_CONFIG,
        )
        unicast_loop(inp, out)
        print(f"N_ITERS={N_ITERS}  DONE", flush=True)
    finally:
        ttnn.close_device(device)
```

## Where the offending ops are added

`lib/Dialect/TTL/Transforms/PipeLowering.cpp` (v1.1.1) `lowerCBToPipe` injects sender-side `cb_reserve_back` / `cb_push_back` on the destination CB to keep the sender's local `fifo_wr_ptr` in lockstep with receivers:

- `PipeLowering.cpp:218-220` — `ttk::CBReserveBackOp::create(rewriter, loc, senderRecvCB, numTilesI32)`
- `PipeLowering.cpp:369-371` — matching `ttk::CBPushBackOp`

On the sender's core, nothing consumes `senderRecvCB`, so there is no `cb_pop_front` / `cb_wait_front` to advance `pages_acked`. Each loop iter increments `pages_received` by 1; iteration `block_count + 1` blocks forever on `cb_reserve_back`.

Introduced in #578 / 224813fe (\"Fix write pointer tracking on pipe sender across iterations\").

The `ttl-verify-dfb-spsc` verifier from the same commit doesn't catch this — it walks `ttl.cb_reserve` / `ttl.cb_wait` before `convert-ttl-to-ttkernel` injects the offending ops.

## Lowered IR snippet (sender's dm body, n=20 loop)

The relevant `cb_ctarg_1` (`in_cb`) pair lives in the sender's pipe-active gate with no matching wait/pop:

```
emitc.for ... step 1 (20 iters) {
  if (node_x == 1) {                       // sender role
    cb_ctarg_0.reserve_back(1);             // send_cb (user)
    noc_async_read_tile(...);
    cb_ctarg_0.push_back(1);
    cb_ctarg_0.wait_front(1);
    if (node_y == 0) {                      // pipe-src match
      cb_ctarg_1.reserve_back(1);           // <-- injected
      noc_async_write(... cb_ctarg_1.get_write_ptr() ...);
      noc_semaphore_inc(...);
      cb_ctarg_1.push_back(1);              // <-- injected, no pop ever
    }
    cb_ctarg_0.pop_front(1);
  }
  // ... receiver branch (node_x == 0) is balanced ...
}
```

brnorris03:
### Problem description

The PipeNet spec currently describes pipe transfer as if data is held in a logical in-transit buffer between the sender and receiver. The runtime protocol does not have such a separate buffer: the sender must write directly into a receiver-owned DFB slot. That means the receiver has to reserve a destination slot and publish its address before the sender can perform the transfer.

The previous lowering did not make that synchronization contract explicit. It treated pipe receives as a sender-side implementation detail, which let the compiler accept syntactically valid same-thread schedules that could deadlock at runtime. In addition, the previous implementation also regressed true unicast loops by adding synthetic DFB producer operations that were not balanced with the user's reserve/wait structure; this breaks the tt-lang contract with users that users are completely responsible for the data movement associated with pipes.

### What's changed

- Pipe receives were previously modeled as a sender-side implementation detail. They now use the user's reserved destination DFB slot: `ttl.copy(pipe, dst)` lowers to the receive post/address publication, and waiting on that copy waits for transfer completion.
- Pipe sends previously assumed the transfer target could be derived without an explicit receiver post. They now respect the receiver-posted protocol for both unicast and multicast, including repeated transfers in loops.
- Pipe synchronization state was previously shared too broadly across pipes and data-movement threads. It is now scoped so independent pipes and concurrent data-movement threads do not interfere with each other. The detailed lowering protocol is documented in `docs/development/PipeNets.md`.
- Receive completion previously used state that did not advance correctly for repeated receives. It now uses per-PipeNet runtime counters, so repeated unicast and multicast receives in loops advance correctly across iterations.
- The PipeNet verifier previously accepted same-thread pipe schedules that could deadlock. It now rejects those schedules and emits detailed diagnostics for common mistakes:
  - waiting for a receive before the send that completes it;
  - sending before the receiver has posted `ttl.copy(pipe, dst)`.
- Pipe receive waits were previously omitted from the wait-for graph when their guard could not be analyzed. They now have the same static guard requirement as the receive post, so `ttl.wait` on a pipe receive handle under an unanalyzable coordinate-dependent guard is rejected.
- Internal receive lowering and verification previously allowed invalid pipe receive IR to fail later with less specific diagnostics. They now reject invalid receive IR earlier with clearer diagnostics.
- Per-destination multicast receive addresses are tracked as future work in issue #617.
- Pipe tests previously encoded schedules that did not consistently publish receiver-owned destinations before sends. They now express valid receiver-posted schedules. Existing loopback multicast cases that did not need loopback were changed to exclude the source from the destination range.
- Test coverage previously missed invalid same-thread schedules, unanalyzable receive-wait guards, internal pipe receive op verifier errors, true unicast loops, multicast loops, loopback multicast, unicast forwarding-chain patterns, and mesh-SPMD pipe execution. The PR adds negative pytest/lit coverage for the invalid cases and positive coverage for the valid cases, including the issue #608 reproducer.

### Tests

```
+--------------------------------------+----------------------+------------------+-----------------------------+
| Pattern / test area                  | Pipe type            | Size shape       | Control-flow covered        |
+--------------------------------------+----------------------+------------------+-----------------------------+
| true unicast split loop              | unicast 1 -> 1       | 1 tile           | 20-iter loop, same DM thread|
| gather                               | N unicast -> 1 dst   | 1 tile           | per-core if, receiver sum   |
| gather multiblock                    | N unicast -> 1 dst   | 1 x 2 tiles      | multi-tile DFB slots        |
| forward ring                         | unicast ring         | 1 tile           | separate DM send/recv       |
| row rings, grid=full                 | many unicast rings   | 1 tile           | full-grid guards            |
| pipe conv chain                      | unicast chain        | 1 x 2 tiles      | conditional send, pipeline  |
| row/col forwarding chains            | unicast chains       | 1 x 2 tiles      | 5-iter loops, row+col chain |
| scatter                              | multicast 1 -> N     | 1 tile           | non-loopback multicast      |
| broadcast_2d                         | multicast 1 -> 2D    | 1 tile           | loopback, same DM ordered   |
| all-to-all 1D/2D                     | multicast N -> N     | 1 tile           | loopback, reductions        |
| overlapping multicast                | multicast overlap    | 1 tile           | multiple senders per recv   |
| partial overlap                      | multicast overlap    | 1 x 2 tiles      | asymmetric recv counts      |
| gather + broadcast loop              | unicast + multicast  | multi-row block  | stripe loops, two PipeNets  |
| rendezvous state                     | unicast/multicast    | 1 tile           | same receiver/source cases  |
| cross-DFB loopback multicast         | multicast loopback   | 1 tile           | 2-iter loop, src/dst DFBs   |
| mcast matmul                         | row/col multicast    | 8 x 8 blocks     | K loops, large grid         |
| minimal matmul mirror                | row/col multicast    | 8 x 8 blocks     | K loops, tt-metal shape     |
| mesh SPMD unicast                    | unicast 1 -> 1       | per-device shard | multi-device mesh tensors   |
| captured/module/mixed-scope nets     | mixed PipeNet usage  | 1 tile           | Python scope variants       |
| invalid schedule tests               | unicast/multicast    | 1 tile           | wait-before-send errors     |
| invalid receive-wait guard           | unicast              | 1 tile           | unknown coord guard reject  |
+--------------------------------------+----------------------+------------------+-----------------------------+
```

Fixes #608.


zoecarver (approved):
Happy to see the compiler-inserted reserves going away :)

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -824,7 +903,40 @@ struct CopyLowering : OpConversionPattern<CopyOp> {
...
+struct PipeRecvPostLowering : OpConversionPattern<PipeRecvPostOp> {
+  PipeRecvPostLowering(const TypeConverter &typeConverter, MLIRContext *context,
+                       const PipeRuntimeLayout *pipeRuntimeLayout)
+      : OpConversionPattern(typeConverter, context),
+        pipeRuntimeLayout(pipeRuntimeLayout) {}
+
+  LogicalResult
+  matchAndRewrite(PipeRecvPostOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    return lowerPipeRecvPost(op, adaptor.getPipe(), op.getDst(),
```
zoecarver, line 921: Should this be adaptor.getDst?

lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
```diff
@@ -187,6 +192,28 @@ struct WaitUse {
   int64_t cbIndex;
 };
 
+CopyOp findDefiningCopy(Value value) {
```
zoecarver: Filter pip copies like `findPipeReceiveCopy` does?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03: Good catch! Updated both PipeRecvPostLowering and PipeRecvWaitLowering.

brnorris03: Actually, we do need op here -- adaptor no longer has the attributes which I still need -- `ttl.cb_reserve`, `ttl.attach_cb`, and slice offset.

lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
brnorris03: Agreed! Extracted the filter and reused it, also made the helper return optional.

FConstantinos (approved):

Issue #625, opened 2026-05-24
PipeNet dual delivery-path reproducer deadlocks when both routes are enabled


brnorris03:
### Problem

Łukasz Galas submitted a PipeNet reproducer for a deadlock in a K-split
communication pattern.

Original report summary:

> Edge cores broadcast pieces of a matrix along two different delivery paths via
> PipeNets. If just one of these paths is enabled it works; if both are enabled,
> it deadlocks.

This issue tracks the original `both` case: `REPRO_MODE_ID = 2`, where the
main-grid and helper-column delivery routes are enabled together for even K
tiles.

### Expected behavior

The kernel completes when both delivery routes are enabled.

### Actual behavior

The original schedule deadlocks when both routes are enabled.

### Complete reproducer

```python
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
# SPDX-License-Identifier: Apache-2.0
#
# Minimal reproducer: main K-split PipeNets + helper-column PipeNets on
# (GRID_DIM + 1, GRID_DIM).
#
# This is a small hardware test where edge cores broadcast pieces of a matrix
# along two different on-chip delivery paths (into a main tile block and an
# optional extra column), to check whether turning both paths on at once finishes
# or deadlocks.
#
# Grid layout (GRID_DIM=2 -> 3x2 cores, coordinates (core_x, core_y)):
#
#     core_x:   0      1      2 (=GRID_DIM helper column)
#   core_y=0  (0,0)  (1,0)  (2,0)
#   core_y=1  (0,1)  (1,1)  (2,1)
#
# K-tile routing (k_tiles=2 -> half_k=1 odd + 1 even step per buffer):
#   odd  k_idx -> row_upper_net / col_upper_net  (main grid only)
#   even k_idx -> row_lower_net / col_lower_net  (main grid)
#              -> helper_row_even_net / helper_col_even_net
#
# REPRO_MODE_ID:
#   0 = main_only    1 = helper_only    2 = both (main + helper on even K)

from __future__ import annotations

import torch
import ttnn
import ttl

TILE = 32
GRID_DIM = 2  # 3x2 grid; k_tiles=2, half_k=1

REPRO_MODE_ID = 2
REPRO_MODE_NAMES = ("main_only", "helper_only", "both")


@ttl.operation(grid=(GRID_DIM + 1, GRID_DIM), fp32_dest_acc_en=True)
def k_split_dual_pipes_repro(x, out):
[...]

PR #624, opened 2026-05-24
[ttl] Introduce pipe transfer IR

brnorris03:
### Problem description

PR #622 makes PipeNet lowering use explicit physical resources, but the pre-TTKernel IR still infers transfer structure from `ttl.create_pipe`, `ttl.copy`, and `ttl.wait`. That keeps the logical pipe transfer contract implicit, even though later follow-ups need to reason about transfer identity, receiver-authored destination publication, ready counting, sender transfer, and receiver completion as separate concepts.

### What's changed

This PR introduces an internal `ttl.pipe_transfer` IR layer without changing the user-facing `ttl.create_pipe`, `ttl.copy`, and `ttl.wait` behavior.

- Adds `!ttl.pipe_transfer`, `!ttl.pipe_token<net N>`, and `#ttl.pipe_transfer_kind<point_to_point|collective>`.
- Adds `ttl.pipe_transfer.create`, `ttl.pipe_transfer.post`, `ttl.pipe_transfer.send`, and `ttl.pipe_transfer.wait`.
- Converts existing public pipe ops to the new internal transfer IR before TTKernel lowering.
- Updates PipeGraph and PipeLowering to consume transfer ops instead of the old internal receive-post/receive-wait helpers.
- Keeps the current PR #622 physical lowering behavior intact.

This PR does not add liveness allocation, phased lowering, DFB batching, or transfer grouping.

### Tests

- `cmake --build build --target ttlang-opt`
- `source build/env/activate && llvm-lit build/test -v --filter='(pipes|convert_pipe_ops)'`
- `git diff --check`

### Stacked PR sequence

Pipe compilation work required to match the optimized kernels' scalable communication model: GlobalSemaphore-backed counting, explicit L1 address/state, DFB lifetime allocation, and batching where storage is limiting.

| # | PR | Status | Main problem solved |
| --- | --- | --- | --- |
| 1 | #614 | Base PR | Makes PipeNet transfers correct by switching to receiver-posted destination DFB addresses. This removes sender inference of receiver DFB write-pointer state, but not semaphore pressure. |
| 2 | #622 | Base PR | Removes collective receiver-count and source-local pipe-count semaphore growth by using receiver-authored SRAM address tables, compiler-emitted pipe resource plans, and GlobalSemaphore-backed ready counters. This eliminates the local semaphore-id scaling limit for large uniform collective transfers. |
[...]

Issue #623, opened 2026-05-23
PipeNet phased channel lowering for repeated sends on one logical channel


brnorris03:
Parent: #619
Related: #620, #622

PR #622 removes local semaphore-id pressure for large uniform multicast PipeNets by moving address publication into SRAM tables and using GlobalSemaphore-backed ready counters. The remaining depth-1 protocol still treats a logical pipe channel as having one live address publication and one ready/completion exchange at a time.

This blocks programs that reuse the same logical PipeNet for multiple send/receive rounds in one kernel body, including receive-ahead or pipelined loops. The current verifier correctly rejects these schedules as wait-for cycles because sender-ready counters are reset per send and address-table entries have depth 1.

Required work:

- Represent channel phase explicitly in IR, likely after `ttl.pipe_channel` exists.
- Allocate address-table entries by channel phase/lifetime rather than by logical pipe only.
- Lower ready counting with monotonic phase counts or separate per-phase counters so senders do not infer receiver state.
- Add deterministic diagnostics when required phase depth cannot be allocated.
- Re-enable the strict xfail coverage in `test/python/pipe/test_pipenet_sync.py` for repeated GlobalSemaphore-backed fanout rounds.

PR #627, opened 2026-05-24
[ttl] Allocate pipe resources by transfer liveness

brnorris03:
### Problem description

PR #624 introduces explicit Pipe Transfer IR, but physical synchronization resources are still allocated mostly by logical pipe or PipeNet identity. That means two same-source transfers consume distinct sender-ready counters and address-table slots even when their transfer lifetimes do not overlap.

In this PR, a transfer is live from the receive post (`ttl.copy` from pipe to destination DFB) that publishes the receiver-owned destination address until the send (`ttl.copy` from source DFB to pipe) that consumes that address and the corresponding sender-ready count. After that send, the source-core address-table slot and sender-ready counter can be reused even if the public transfer handle remains live until `ttl.wait`. Allocating by logical pipe keeps resource use tied to the number of logical pipes instead of the number of concurrently live transfers, and prevents the generated resource plan from expressing reuse when same-source transfers are serialized.

### What's changed

This PR uses explicit Pipe Transfer IR lifetimes to allocate the source-core resources whose values are needed only between receive post (`ttl.copy` from pipe to destination DFB) and send (`ttl.copy` from source DFB to pipe). For example, this reduces the issue [#625](https://github.com/tenstorrent/tt-lang/issues/625) reproducer compiler-managed pipe SRAM scratch from 64 bytes to 32 bytes by reusing non-overlapping address-table entries.

- Allocates sender-ready counters from transfer live intervals instead of from logical pipe count.
- Allocates receiver-authored source-core address-table slots from the same transfer live intervals.
- Extracts shared live-interval allocation utilities so future resource allocators can reuse the same deterministic interval coloring logic.
- Keeps receiver-completion counters per PipeNet in this PR.
- Keeps user-facing `ttl.create_pipe`, `ttl.copy`, and `ttl.wait` behavior unchanged.
- Preserves the PR #622 aggregate-ready protocol and the PR #624 internal Pipe Transfer IR contract.
- Adds a deterministic diagnostic for the same-block case where a later receive post would require more than one live receive post for the same logical pipe.
- Clarifies that TT-Metal NoC multicast supports one destination L1 (SRAM) address for all receivers, so non-uniform collective destination addresses are rejected rather than tracked as future per-receiver multicast support.

This PR does not add phased lowering, receive-ahead transfer posting, DFB batching, or transfer grouping (future planned PRs, see below).

### Tests

New tests cover same-source overlap and reuse in MLIR lowering, the queue-depth diagnostic for a second receive post before an intervening send, the updated non-uniform multicast destination diagnostic, and runtime Python lit regression coverage based on the dual-route ksplit PipeNet schedule in #625.

### Stacked PR sequence

Pipe compilation work required to match the optimized kernels' scalable communication model: GlobalSemaphore-backed counting, explicit L1 address/state, live-interval allocation, and batching where storage is limiting.

| # | PR | Status | Main problem solved |
| --- | --- | --- | --- |
| 1 | #614 | Base PR | Makes PipeNet transfers correct by switching to receiver-posted destination DFB addresses. This removes sender inference of receiver DFB write-pointer state, but not semaphore pressure. |
| 2 | #622 | Base PR | Removes collective receiver-count and source-local pipe-count semaphore growth by using receiver-authored SRAM address tables, compiler-emitted pipe resource plans, and GlobalSemaphore-backed ready counters. This eliminates the local semaphore-id scaling limit for large uniform collective transfers. |
| 3 | #624 | Base PR | Represents transfer phase, receiver-authored address publication, ready counting, send, and receive-token wait explicitly before control-flow-general lowering. This removes the need to infer transfer phase and queue depth from `ttl.copy` placement. |
| 4 | #627 | Current PR | Assigns source-core address-table slots and sender-ready counters from explicit pipe transfer lifetimes. This reduces those resources from logical pipe count to concurrently live same-source transfer count while keeping receiver completion per PipeNet. |
| 5 | TBD | Planned follow-up | Supports phased pipe transfer lowering for receive-ahead and pipelined loops with monotonic ready counts, finite address-table depths, and deterministic diagnostics when safe pipe transfer state cannot be allocated. This extends the current one-live-post protocol to programs where later receive posts may be live before earlier sends complete. Issue [#623](https://github.com/tenstorrent/tt-lang/issues/623). |
| 6 | TBD | Planned follow-up | Lowers large all-to-all into communication batches when receiver DFB capacity is smaller than the number of incoming pipes. This eliminates the simultaneous receiver DFB slot scaling limit when the program can consume batches incrementally. |
| 7 | TBD | Planned follow-up | Groups post/wait state when pipe tokens prove identical lifetime and completion behavior. This reduces duplicate pipe synchronization resources after liveness allocation exists. |


lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -196,17 +195,20 @@ getFirstPipeGlobalSemaphoreArgOffset(const PipeResourcePlan &info) {
...
+enum class ReadyCounterAddressStorage {
+  LocalSemaphore,
+  GlobalSemaphoreRuntimeArg,
 };
 
-struct GlobalReadyCounterAddressInfo {
-  int64_t runtimeCommonArgIndex;
+/// Sender-ready counters can live either in local semaphore space or in
+/// GlobalSemaphore-backed SRAM. The storage kind disambiguates the index value.
+struct ReadyCounterAddressInfo {
```
vtsilytskyiTT: This looks like a typical OOP task. There are 2 classes: `LocalSemaphore` and `GlobalSemaphoreRuntimeArg`, both contain a single value `int64_t index;` and both implement the same interface (or override a method from base class. The all `std::get_if<PipeGlobalReadyCounterInfo>` below should go away.

lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
```diff
@@ -124,41 +126,47 @@ static void reuseDFBIndices(func::FuncOp funcOp, ArrayRef<BindCBOp> dfbOps) {
...
-    llvm::sort(intervals, [](const Interval &lhs, const Interval &rhs) {
-      return lhs.start < rhs.start;
-    });
+  for (auto &entry : typeToIntervals) {
+    SmallVector<ValueLiveInterval> &intervals = entry.second;
+    SmallVector<unsigned> intervalIndices;
+    intervalIndices.reserve(intervals.size());
+    for (unsigned index = 0, size = intervals.size(); index < size; ++index) {
+      intervalIndices.push_back(index);
+    }
```
vtsilytskyiTT: C++ has std::iota. I'm not sure whether it will work with SmallVector.

lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
```diff
@@ -124,41 +126,47 @@ static void reuseDFBIndices(func::FuncOp funcOp, ArrayRef<BindCBOp> dfbOps) {
...
+              const ValueLiveInterval &lhs = intervals[lhsIndex];
+              const ValueLiveInterval &rhs = intervals[rhsIndex];
+              if (lhs.start != rhs.start) {
+                return lhs.start < rhs.start;
+              }
+              if (lhs.end != rhs.end) {
+                return lhs.end < rhs.end;
+              }
+              return lhsIndex < rhsIndex;
+            },
```
vtsilytskyiTT: This should be the `ValueLiveInterval`'s method. You can override `bool operator<` and then use `std::less` everywhere.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -962,95 +997,379 @@ void buildPipeNetIndex(ModuleOp mod, PipeNetIndex &index) {
...
+  Post,
+  Send,
+};
+
+/// One ordered post/send operation used to validate bounded rendezvous depth.
+struct PipeTransferRendezvousEvent {
+  /// Pipe transfer post or send operation.
+  Operation *op;
+  /// Whether the operation creates or consumes one posted rendezvous phase.
+  PipeTransferRendezvousEventKind kind;
```
vtsilytskyiTT, line 1031: Same pattern. You are using enum + structure where OOP applies better.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -962,95 +997,379 @@ void buildPipeNetIndex(ModuleOp mod, PipeNetIndex &index) {
...
+                                   const PipeTransferAllocationUnit &rhs) {
+  return std::make_tuple(lhs.interval.startOrdinal, lhs.pipe.srcX,
+                         lhs.pipe.srcY, lhs.pipe.pipeNetId, lhs.pipe.dstStartX,
+                         lhs.pipe.dstStartY, lhs.pipe.dstEndX, lhs.pipe.dstEndY,
+                         lhs.ordinal) <
+         std::make_tuple(rhs.interval.startOrdinal, rhs.pipe.srcX,
+                         rhs.pipe.srcY, rhs.pipe.pipeNetId, rhs.pipe.dstStartX,
+                         rhs.pipe.dstStartY, rhs.pipe.dstEndX, rhs.pipe.dstEndY,
+                         rhs.ordinal);
+}
```
vtsilytskyiTT: Again this may me the `PipeTransferAllocationUnit`'s method `operator<`.

lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
brnorris03: It probably will, but incurs an extra copy if I recall correctly. Probably not a big deal in terms of compiler performance, so will change.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: I would argue, though, whether heavy (in terms of LOC, code complexity and runtime dispatch) design for something with exactly 2 kinds should be an actual class hierarchy. I personally think that's over-engineering for a local helper struct. Why do I need the full power of an OO hierarchy here?

lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
brnorris03: Agreed, updated.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: Agreed, updated.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: Agreed -- replaced ready-counter variant/get_if logic with an abstract `PipeReadyCounterInfo` and concrete local/global counter classes.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: I did move the behavior onto `PipeTransferRendezvousEvent` simplifying the callsites, but keeping the simple enum class for the two event kinds.

vtsilytskyiTT: OOP reduces number of `if` clauses in the code. They become hidden in vtable. This is the only real benefit in our case.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -1010,12 +1044,31 @@ enum class PipeTransferRendezvousEventKind {
   Send,
 };
 
+struct PipeTransferAllocationUnit;
```
vtsilytskyiTT, line 1016: I see, circular dependency between structs. Didn't think of this :(

vtsilytskyiTT (approved):
LGTM

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: In general, yes, but for something that will never have more than two constant values, it's overkill in definition complexity, just to eliminate one switch over clear enum values.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: It's not a problem IMO, was a good suggestion.

include/ttlang/Dialect/TTL/Transforms/LiveIntervalUtils.h
```diff
@@ -0,0 +1,213 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//===----------------------------------------------------------------------===//
+// TTL Live Interval Utilities
```
zoecarver, line 6: Drive by thought (not something to change now): I wonder if we could combine this with CB re-use logic...

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -962,165 +1032,477 @@ void buildPipeNetIndex(ModuleOp mod, PipeNetIndex &index) {
...
+
+static LogicalResult
+validateMaxLivePostsPerLinearBlock(const PipeTransferAllocationUnit &unit,
+                                   int64_t maxLivePosts) {
+  if (unit.rendezvousEvents.size() <= static_cast<size_t>(maxLivePosts + 1)) {
+    return success();
+  }
+
+  llvm::MapVector<Block *, SmallVector<PipeTransferRendezvousEvent>>
+      eventsByBlock;
```
zoecarver, line 1150: If two posts are not in the same block, no block ever sees two posts, and validation passes, right? I think `convert_pipe_ops_future.mlir` has this case, but how important do you think it is to handle? Is it something we should prioritize? Maybe a safe fallback for now, like if a unit has >= 2 posts and any pair isn't same-block ordered, error?

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -962,165 +1032,477 @@ void buildPipeNetIndex(ModuleOp mod, PipeNetIndex &index) {
...
+    }
+    return success();
+  }
+  llvm_unreachable("unknown pipe transfer rendezvous event kind");
+}
+
+static LogicalResult
+validateMaxLivePostsPerLinearBlock(const PipeTransferAllocationUnit &unit,
+                                   int64_t maxLivePosts) {
+  if (unit.rendezvousEvents.size() <= static_cast<size_t>(maxLivePosts + 1)) {
```
zoecarver: Maybe I'm being pedantic but could a case like two posts trigger this shortcut despite being two events? Maybe instead we could tighten to just <= (no +1)?

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -188,52 +188,92 @@ getFirstPipeGlobalSemaphoreArgOffset(const PipeResourcePlan &info) {
...
+  void observe(PipeReadyCounterObserver &observer) const override {
+    observer.observeGlobalSemaphore(index);
+  }
+
+private:
+  int64_t index;
 };
 
-using ReadyCounterAddressInfo =
-    std::variant<LocalReadyCounterAddressInfo, GlobalReadyCounterAddressInfo>;
```
zoecarver, line 200: Sorry if this is pedantic, but I'd prefer the value type with two cases over shared_ptr + virtual info class + observer, which seems a bit heavy handed for what it's doing.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -188,52 +188,92 @@ getFirstPipeGlobalSemaphoreArgOffset(const PipeResourcePlan &info) {
...
+/// Ready counter backed by a TTKernel local semaphore id.
+class LocalSemaphoreReadyCounterInfo final : public PipeReadyCounterInfo {
+public:
+  explicit LocalSemaphoreReadyCounterInfo(int64_t index) : index(index) {}
+
+  ReadyCounterAddressInfo
+  getAddressInfo(Operation *op,
+                 const PipeResourcePlan &pipeResourcePlan) const override {
+    (void)op;
+    (void)pipeResourcePlan;
```
zoecarver: nit throughout: can we just remove the parameter/name?

lib/Dialect/TTL/Transforms/PipeLowering.cpp
vtsilytskyiTT: It was my request. You'll have observer anyway. Virtual class removes "if (type1) { ... } else { ... }" constructs from the code. Matter of flavor, of course. For me less `if` conditions is nicer.

include/ttlang/Dialect/TTL/Transforms/LiveIntervalUtils.h
brnorris03: We do reuse the generic coloring helper. The interval construction is simple enough that given the completely different endpoints for this vs dfb reuse, implementing it concisely here seemed preferable than generalizing, which would introduce more complexity. An abstract implementation would need a lot of specialization (not just start and end ops), e.g., grouping, enpoint projection, missing endpoint policy, etc. I think that if we have to do this for a third type of liveness, then perhaps it is worth considering a more general framework and refactor this at that point.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: I kind of lean on the side of not overengineering that, so went back to the switch (which I think is clear enough). If we ever find the need to extend this, I could refactor it to use OO.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: Good catch, leftover from an early version where it was considering the post + send event, but since it doesn't actually check the kinds of events, it would apply to two posts. Fixed.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: The old validation needed a fix, thanks. This PR supports queue depth 1 per logical pipe: only one receive post may be live before the matching send consumes the posted address/readiness state. That's a (known) current lowering limitation; depth > 1 will be added separately, and at that point the validation will be updated (again) accordingly.

zoecarver (approved):

Issue #630, opened 2026-05-26
PipeNet schedule verifier adds program-order edges across mutually exclusive scf.if branches


brnorris03:
### Problem

`TTLVerifyPipeNetGuards.cpp` can report a false PipeNet wait-for cycle for pipe events in mutually exclusive `scf.if` branches.

The diagnostic is also misleading. It says a send is ordered after a receive wait in the same data-movement thread, but in the reproducer below those operations are in sibling `scf.if` branches and cannot both execute in one runtime trace.

### Reproducer

Save this as `/tmp/pipenet_if_branch_repro.mlir`:

```mlir
module attributes {ttl.launch_grid = [1 : i64, 1 : i64]} {
  func.func @coord_invariant_if_false_positive_cycle(%runtime_flag: i1)
      attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
    %pipe = ttl.create_pipe src(0, 0) dst(0, 0) to(0, 0) net 0
        {pipeNetName = "net"}
        : !ttl.pipe<src(0, 0) dst(0, 0) to(0, 0) net 0>
    %send_cb = ttl.bind_cb {cb_index = 0, block_count = 2}
        : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
    %recv_cb = ttl.bind_cb {cb_index = 1, block_count = 2}
        : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>

    scf.if %runtime_flag {
      %recv_reserve = ttl.cb_reserve %recv_cb
          : <[1, 1], !ttcore.tile<32x32, bf16>, 2>
          -> tensor<1x1x!ttcore.tile<32x32, bf16>>
      %recv_view = ttl.attach_cb %recv_reserve, %recv_cb
          : (tensor<1x1x!ttcore.tile<32x32, bf16>>,
             !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>)
          -> tensor<1x1x!ttcore.tile<32x32, bf16>>
      %recv = ttl.copy %pipe, %recv_view
          : (!ttl.pipe<src(0, 0) dst(0, 0) to(0, 0) net 0>,
             tensor<1x1x!ttcore.tile<32x32, bf16>>)
          -> !ttl.transfer_handle
      %send = ttl.copy %send_cb, %pipe
          : (!ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>,
[...]

PR #632, opened 2026-05-27
ttlverifypipenetguards.cpp: fix false-positive cycle on mutually exclusive scf.if branches

patelkrish5090:
### Problem description

`verifyPipeScheduleCycles` used a flat `module.walk` to build `ProgramOrder`
edges between pipe events. The frontier key `(funcOp, coord.x, coord.y)` had
no knowledge of control flow, so the last event in the then-branch of an
`scf.if` became the predecessor of the first event in the else-branch. That
spurious edge created a false cycle on loopback pipes where both branches
contained a locally valid schedule.

Reproducer: a 1x1 grid, single loopback pipe, both branches of `scf.if` do
post-receive / send / wait-send / wait-receive. Each branch is valid on its
own, but the verifier reported "receive wait occurs before the send that
completes it" because it incorrectly ordered the then-branch's wait before
the else-branch's send.

### What's changed

Split `verifyPipeScheduleCycles` into two passes:

1. Flat `module.walk` to collect `PipeScheduleNode` entries (unchanged
   behavior).
2. CFG-aware traversal (`buildProgramOrderEdgesInRegion`) that tracks a
   frontier map of `(funcOp, coord) -> last node id`. On `scf.if`, each
   branch receives an independent copy of the frontier so no `ProgramOrder`
   edge is ever added between sibling branches. The post-if frontier is the
   union of both branch exits.

Added `verify_pipenet_schedule_if_branch.mlir` with two positive tests
(the exact reproducer and a nested-if variant) using `CHECK-NOT` to confirm
no false-positive diagnostic is emitted.

### Checklist

THis will fix the issue #630

### DIFF
--- a/lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLVerifyPipeNetGuards.cpp
@@ -1431,6 +1431,74 @@ void emitPipeScheduleCycleDiagnostic(ArrayRef<PipeScheduleNode> nodes,
   state.sawError = true;
 }
 
+// Maps (funcOp, coord.x, coord.y) to the last pipe schedule node id seen
+// along the current execution path.
+using ProgramOrderFrontier = llvm::DenseMap<ProgramPointIdentity, unsigned>;
+
+static void buildProgramOrderEdgesInOps(
+    Block &block, SmallVectorImpl<PipeScheduleNode> &nodes,
+    const llvm::DenseMap<PipeNodeIdentity, unsigned> &nodeIds,
+    ProgramOrderFrontier &frontier, func::FuncOp funcOp);
+
+static void buildProgramOrderEdgesInRegion(
+    Region &region, SmallVectorImpl<PipeScheduleNode> &nodes,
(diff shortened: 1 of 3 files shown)


brnorris03:
Thank you for the contribution -- however there is a stack of 3 more pre-existing pipe PRs (all ready for review), the last being #627, so it would be best if this PR is rebased after they merge (this PR will conflict with all of them and rebasing a whole stack is extremely unpleasant). Also, it would be good to first check that the issue is still present in PR 627.

brnorris03:
I guess if you don't mind waiting in line -- you can stack this on top of 627 (I have been using Graphite's free gt client to manage stacking, makes it a bit easier to propagate changes)? Also can you try adding a pytest?

patelkrish5090:
I have checked, and the issue is definitely still present in PR #627. The root cause is the flat module.walk used to build ProgramOrder edges in verifyPipeScheduleCycles. Since the #627 stack doesn't rewrite that specific traversal logic in TTLVerifyPipeNetGuards.cpp, this false-positive bug still occurs there.

I have added pytest in new commit, you can review it. Also I dont mind waiting for PR #627 to be merged, if you want i can stack mine on top. Totally your call.

PR #634, opened 2026-05-28
dfb: add DataflowBuffer.peek() for non-popping reads to enable multi-pass tile reuse without DRAM reloads or intermediate buffers

patelkrish5090:
### Problem description

Kernels that need to iterate over the same tiles more than once (e.g., rmsnorm backward: reduce_sum first, then element-wise
multiply each tile by the reciprocal) currently have two options: re-read the block from DRAM a second time, or allocate an intermediate buffer to hold the data. Both are wasteful.

On TT-Metal hardware, cb_wait_front(n) is idempotent -- calling it a second time without an intervening cb_pop_front simply returns immediately if the tiles are still present. This property can be exposed safely at the ttl level.

### What's changed

Added DataflowBuffer.peek(). Like wait(), it emits ttl.cb_wait. Unlike wait(), the with-block exit does not auto-insert ttl.cb_pop.
The slot is held until the caller issues a subsequent wait() (which does pop) or an explicit pop.

The typical pattern for a two-pass kernel:

    with inp_cb.peek() as blk:       # first pass, slot held
        scalar = ttl.reduce(blk, ...) 
    with inp_cb.wait() as blk:       # second pass, auto pop on exit
        o.store(ttl.mul(blk, scalar))

Changes:
- python/ttl/dataflow_buffer.py: add DataflowBuffer.peek()
- python/ttl/_src/ttl_ast.py: extend with-block handler to accept
  "peek" and skip the cb_pop release entry

### Checklist

### DIFF
--- a/python/ttl/_src/ttl_ast.py
+++ b/python/ttl/_src/ttl_ast.py
@@ -1282,10 +1282,10 @@ class TTLGenericCompiler(TTCompilerBase):
             method_name = context_expr.func.attr
             cb_node = context_expr.func.value
 
-            if method_name not in ("reserve", "wait"):
+            if method_name not in ("reserve", "wait", "peek"):
                 self._raise_error(
                     context_expr,
-                    f"'with' only supports 'reserve()' or 'wait()', got '{method_name}'",
+                    f"'with' only supports 'reserve()', 'wait()', or 'peek()', got '{method_name}'",
                 )
 
             if not isinstance(cb_node, ast.Name):
@@ -1299,7 +1299,7 @@ class TTLGenericCompiler(TTCompilerBase):
                 self._raise_error(cb_node, f"'{cb_node.id}' not found in scope")
             cb_val = cb_table[cb_node.id]
 
-            # Get tensor type from CB for reserve/wait result
+            # Get tensor type from CB for reserve/wait/peek result
             tensor_type = self._get_cb_tensor_type(cb_val, node=context_expr)
             if method_name == "reserve":
                 tensor = self._emit_op_signposts(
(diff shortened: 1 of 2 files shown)


brnorris03:
I think this can be expressed with the current DFB lifetime model. The issue is that the example splits the input lifetime across two `with inp_cb...` scopes. Instead, keep one acquired input block live across both consumers and release it after the second use:

```python
with inp_cb.wait() as inp_blk:
    reduced = ttl.math.reduce_sum(inp_blk, dims=[...])
    scale = ttl.block.broadcast(reduced, dims=[...], shape=(...))
    with out_cb.reserve() as out_blk:
        out_blk.store(inp_blk * scale)
```

or use the explicit form when the lifetime should not match a `with` scope:

```python
inp_blk = inp_cb.wait()
...
inp_blk.pop()
```

Can you try that? If stuck, I can also help with converting.

The spec already defines this explicit form: acquisition without `with` requires an explicit `pop` for `wait`. Adding `peek()` would be a language/API change, not only a compiler convenience, and would need a spec update proposal (and associated discussion before adopting) plus simulator support before the compiler accepts it. You are welcome to contribute such a proposal if we determine that it is indeed a new capability that is missing and needed!

patelkrish5090:
Ohh yeah!! my bad i missed that inp_blk can simply be used twice inside the same with inp_cb.wait() scope and the slot remains live until the block exits. The proposed peek() is not needed for the stated use case. 

Regarding the `with` approach, nested one will prolly work for two pass use case

```
with inp_cb.wait() as inp_blk:
    # reduce over the block
    reduced = ttl.math.reduce_sum(inp_blk, dims=[-1])
    scale = ttl.block.broadcast(reduced, dims=[1], shape=(rows, cols))
    # element-wise scale, reusing inp_blk from the same acquisition
    with out_cb.reserve() as out_blk:
        out_blk.store(inp_blk * scale)
    # inp_cb.pop() fires here automatically
```

inp_blk is a tensor view that stays live for the full with body, so both passes read from the same acquired slot with no extra DRAM reads or intermediate buffers. Check this if its correct then i will close the PR.

PR #654, opened 2026-06-01
[ttl] Add loop carried tensor state [acc-2]

brnorris03:
### Problem description
TT-Lang tensor variables can be reassigned across loop iterations, producing loop-carried tensor state before accumulation-specific optimizations are introduced. That state must lower correctly from Python through SCF and into TTL compute lowering without depending on `ttl.accumulation_scope` formation or a hardware accumulation strategy. This is PR 2 of several split out from the full accumulation scope redesign in PR #651.

### What's changed
This PR adds the loop-carried tensor-state foundation:

- Adds frontend support for reassigned tensor values in `for` loops by emitting `scf.for` iter_args for loop-carried SSA values.
- Adds support for reassigned values across `if` expressions so structured control flow can return updated values.
- Preserves DFB-attached block `+=` lowering through `__iadd__` instead of treating the block value as an SSA iter_arg.
- Adds compiler-allocated DFB materialization helpers shared by loop-state materialization and intermediate DFB insertion.
- Adds `ttl-materialize-loop-state` to replace ranked-tensor `scf.for` iter_args with explicit compiler-managed DFB state.
- Inserts `ttl-materialize-loop-state` into the `ttlang-opt`, E2E, and ME2E pipelines before intermediate DFB insertion.
- Updates the DFB and accumulation-lowering docs to describe general loop-carried tensor-state materialization.
- Adds MLIR lit coverage for tensor loop-state materialization, including zero-trip behavior and conditional recurrence IR.
- Adds runtime coverage for tensor recurrences, multi-accumulator loops, multi-tile loop state, augmented assignment rewriting, and DFB-attached block `+=` behavior.

### Ticket
None.

### Checklist

### DIFF
--- a/python/pykernel/_src/kernel_ast.py
+++ b/python/pykernel/_src/kernel_ast.py
@@ -16,6 +16,202 @@ from .kernel_types import ClassRegistry
 from .utils import _cast, _get_type_str
 
 
+def _extract_target_names(target):
+    """Names bound by a single assignment target, supporting nested tuples
+    and starred unpacking. Subscript/Attribute targets bind storage, not
+    variables, and are skipped."""
+    if isinstance(target, ast.Name):
+        yield target.id
+    elif isinstance(target, (ast.Tuple, ast.List)):
+        for elt in target.elts:
+            yield from _extract_target_names(elt)
+    elif isinstance(target, ast.Starred):
+        yield from _extract_target_names(target.value)
+
+
+class _ScopedCollector(ast.NodeVisitor):
+    """Base for collectors that walk a statement body without descending
+    into nested function or lambda definitions."""
+
+    def visit_FunctionDef(self, node):
+        return
+
+    def visit_AsyncFunctionDef(self, node):
+        return
+
+    def visit_Lambda(self, node):
+        return
+
+    def visit_For(self, node):
+        for stmt in node.body:
+            self.visit(stmt)
+
+    def visit_With(self, node):
+        for stmt in node.body:
+            self.visit(stmt)
+
+    def visit_If(self, node):
+        for stmt in node.body:
+            self.visit(stmt)
+        for stmt in node.orelse:
+            self.visit(stmt)
+
+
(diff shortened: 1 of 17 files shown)


python/pykernel/_src/kernel_ast.py
```diff
@@ -16,6 +16,202 @@
...
+
+    def visit_AugAssign(self, node):
+        # `target op= value` reads `target` implicitly. For block targets
+        # __iadd__ lowers them in place, so AugAssign-only entries
+        # are filtered out by `_get_loop_carried_var_names`.
+        for name in _extract_target_names(node.target):
+            self._add(name, from_augassign=True)
+
+
+class _AssignedVariableCollector(_ScopedCollector):
```
zoecarver: > we have two paths now which are slightly different, can they be unified? _get_if_carried_var_names uses _AssignedVariableCollector and for-loop path uses _LoopCarriedVarCollector. Beyond unifying, I want to make sure they are sufficiently robust. It seems like this change over-indexes on the specific #527 bug, but this is a class of bugs, many of which I don’t think will be caught with the current changes. For example, 
tmp = acc; acc = tmp + x
new = acc + d; acc = new
 a = b = acc + 1Then there’s a tail of things that we don’t really support at the language level, but we should make sure to error for, like the walrus operator, ternary, switch statements, etc. Similarly, for if there are any other loops that we don’t want to handle like while loops. I don’t want to scope creep here, but I also think this seems like a good place to formalize the edges of the language.

zoecarver (commented):
We had an offline discussion in which raised a number of high level/architectural comments about this full PR stack. Including a more concrete one from what we discussed here as it pertains to this PR and isn't as high level.

python/pykernel/_src/kernel_ast.py
brnorris03: Good point, refactored, please take a look

zoecarver (approved):
LGTM

PR #656, opened 2026-06-01
Fix bugs in elemwise tutorial and batched matmul example in the spec

rguoTT:
### Problem description

The batched matmul example in the documentation incorrectly divides the `i` dimension with `TILE_SIZE` as only the 2 innermost dimension are "tiled". Otherwise the `a_dfb` blocks will have incorrect size which is less than the number of matrices in `a` due to the division. [examples/eltwise_add_3d.py](https://github.com/tenstorrent/tt-lang/blob/main/examples/eltwise_add_3d.py) performs a similar batched computation which is correct.

The tutorial example [elementwise-tutorial/step_3_multinode.py](https://github.com/tenstorrent/tt-lang/compare/rguo/spec_and_example_bugfix?expand=1#diff-67519c17c2e03f9d2a54ecba0a6e8c25a3bcaade670d5ae32ae9303cc2fe9d86) calculates `cols_per_node` using `grid_rows` but it should be `grid_cols`.

### What's changed

Removed the division by `TILE_SIZE` for the `i` dimension and use `I` directly in later steps for the matmul example in the spec.

Changed `grid_rows` to `grid_cols` in [elementwise-tutorial/step_3_multinode.py](https://github.com/tenstorrent/tt-lang/compare/rguo/spec_and_example_bugfix?expand=1#diff-67519c17c2e03f9d2a54ecba0a6e8c25a3bcaade670d5ae32ae9303cc2fe9d86).

### DIFF
--- a/docs/sphinx/specs/TTLangSpecification.md
+++ b/docs/sphinx/specs/TTLangSpecification.md
@@ -455,14 +455,14 @@ def elwise_write():
 # ---------------------
 # Batched matrix multiplication with bias:
 #
-# y[i, m, n] = ∑(a[i, m, k] * b[k, n]) + c[m, n]
+# y[l, m, n] = ∑(a[l, m, k] * b[k, n]) + c[m, n]
 #              k
 #
 # Tensor   Torch shape   Note
-# a        I, M, K       Batched a matrix (e.g. input activations)
+# a        L, M, K       Batched a matrix (e.g. input activations)
 # b        K, N          Non-batched b matrix (e.g. weights)
 # c        M, N          Non-batched bias matrix c (e.g. weights)
-# y        I, M, N       Batched y matrix (e.g. output activations)
+# y        L, M, N       Batched y matrix (e.g. output activations)
 #
 # All tensors have tiled layout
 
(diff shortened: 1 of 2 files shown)


docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -466,15 +466,14 @@ def elwise_write():
 #
 # All tensors have tiled layout
 
-# Shape in tiles (I, M, N and K are evenly divisible by TILE_SIZE)
-I_TILES = I // TILE_SIZE
+# Shape in tiles (M, N and K are evenly divisible by TILE_SIZE)
```
brnorris03, line 470: Your fixes look great, this is more of a style comment. I know this was already in the spec (so not insisting on changing here), but I don't like `I` being used (generally letters that look like numbers depending on the font are to be avoided when naming single-letter vars). So since you are touching this part of the spec anyway, maybe makes sense to improve naming, too? I think `B` is more suitable here (instead of `I`), WDYT?

phizalev-TT: I agree. Let's go with B instead of I.

phizalev-TT (approved):
Let's make the change suggested by @brnorris03 and then LGTM!

docs/sphinx/specs/TTLangSpecification.md
rguoTT: I agree that `I` is not good here but if we change to `B` then later in the code we will have `b_block`, `B_BLOCK_SIZE`, etc. converted from `i_block`, `I_BLOCK_SIZE`, respectively. I think the user might get confused since we also have the two input tensors as `a` `b` and we have the CB `b_blk` which is similar in name but a different thing.

phizalev-TT: What if we use L?

brnorris03 (approved):
LGTM!

PR #657, opened 2026-06-02
Proof of concept: `ttl.atom` inlinable, thread-unified kernels that compose

zoecarver:
Introduces a `ttl.operation` analogous called `ttl.atom` which can take pipenets and dfbs as arguments and be composed with other atoms. These atoms can be inlined and can do just compute, just dm, or a mix. 

To support inlining and composition well, atoms must be "thread unified." If atoms had three threads, how would they be merged together? It would get messy and require calling each kernel in the appropriate thread. If atoms were at the thread level, then we couldn't represent an mcast that needs its sender to be send/receive noc split in a single unit. So we need to both introduce the idea of unified threads (aka single threaded mode) and the ability to inline/compose kernels. 

This has been on the tt-lang roadmap for a while, but usually a far out thing. This PR introduces the first implementation of it.  This PR is not particularly large or complicated, the actual logic is under 1000 loc, mostly just new tests. The entire compilation pipeline is unchanged, this is just some preprocessing that happens at the python frontend level.


python/ttl/ttl_api.py
arichinsTT: I think these can be removed now

python/ttl/dataflow_buffer.py
```diff
@@ -69,13 +70,19 @@ def __init__(
         self.tensor = tensor
         self.shape = shape
         self.block_count = block_count
+        self._dtype = dtype
```
arichinsTT, line 81: should we allow users to set a dtype that doesn't match the tensor dtype when it exists?

python/ttl/atom.py
```diff
@@ -0,0 +1,554 @@
...
+
+    dfbs: Dict[str, DataflowBuffer] = {}
+    nets: Dict[str, PipeNet] = {}
+    kept: List[ast.stmt] = []
+    for stmt in fn_def.body:
+        name = _setup_assign_target(stmt)
+        if name is None:
+            kept.append(stmt)
+            continue
+        value = eval(ast.unparse(stmt.value), ns)  # noqa: S307
```
arichinsTT, line 469: I get that this has already succesfully parsed, but is there another option for getting values from the ast without using eval?

arichinsTT (commented):
looks good overall, would like to see some examples of invalid atom usage and correct error handling

python/ttl/dataflow_buffer.py
zoecarver: Hmm. Is there a way to make this initializer private in python and just rely on the factory functions?

python/ttl/atom.py
zoecarver: We could implement a constant evaluator/ast parser, but that seems heavy handed when this works fine. Do you have a suggested way to do it?

python/ttl/dataflow_buffer.py
zoecarver: I can just raise an error I guess.

python/ttl/atom.py
arichinsTT: no this is fine for now, but we should keep an eye on it

arichinsTT (approved):
LGTM excited to see future work on this feature

brnorris03 (commented):
I really like this direction. On naming, I think calling this `@ttl.operation` may be better if others agree (not a fan of atom, but could get used to it).

Since it's a language change/addition, it would be good to explicitly write up the new semantics (some concrete suggestions below) -- this way it will be clear how the implementation covers it. As is, it's not explicitly stated what is legal or not, so we have to infer from whatever the implementation does right now, which may or may not be what the final intent is. I think changing the actual spec requires more discussion, but a short document to go with this PR would be great. I especially want to see a discussion on how atoms compose and what correctness guarantees composition has.

Specific observations/suggestions:

1. Splitting. The splitter should fail before lowering when a statement has anchors from more than one concrete thread (e.g., `ttl.copy(ttl.exp(x), y)`), except for explicit control constructs that the splitter handles recursively. Inferring from the implementation, a valid atom should have a clear per-statement thread owner or be duplicated only when it is side-effect-free compile-time control.

The validation should also prove that every DFB producer has exactly one thread owner and every `push()` / `pop()` is co-located with the matching producer or consumer side. It's not great all this has to happen in Python, but I think postponing this is not viable: if these issues are only detected after synthetic thread generation, the compiler errors would be significantly harder for users to interpret.

2. Scalar arguments and caching. It looks like scalar atom parameters are captured as compile-time values and emitted into the generated IR, but the compile cache key is still the tensor-oriented one from `@ttl.operation`. That means two calls with the same tensor metadata but different scalar values can reuse the first compiled kernel. For example, if an atom computes `out_blk.store(in_blk * scale)`, then `atom(x, out, scale=1.0)` followed by `atom(x, out, scale=2.0)` can reuse the kernel compiled with `scale=1.0`. I think either scalar parameters should be included in the atom cache key, or scalar parameters should be rejected until there is an explicit runtime scalar ABI (adding TODO with issue is fine).

3. Function signatures. Default parameter values are syntactically accepted, but the top-level bind does not apply omitted defaults, and inline atom calls reject omitted parameters rather than substituting default expressions. I think this should be specified one way or the other: either reject default parameter values in atom signatures, or call `apply_defaults()` and teach the inliner to substitute default values.

4. Atom parameter ABI. It would help to write down what kinds of parameters an atom may accept and whether they are runtime values or compile-time values. My current understanding is: tensors become runtime tensor accessors, scalar values are compile-time constants, `ttl.DFB` parameters are inline-only, and PipeNet/DFB objects should not appear as top-level runtime parameters. If that is right, the decorator should validate unsupported signatures early instead of letting them fail later while compiling a generated thread body.

5. DFB ownership and thread inference. The current implementation already has some good checks for DFB producers that would need to be split onto both NCRISC and BRISC, but I think the same idea should be generalized: each DFB acquire should have one inferred thread owner, and multiple consumer/producer roles should be rejected with an atom-specific diagnostic before the generated MLIR reaches the lower-level SPSC verifier. That would make the source-level error much easier to understand.

6. Duplicated language knowledge. The splitter has its own registries for TTL ops, namespaces, PipeNet methods, DFB producer methods, and block methods. This seems likely to drift from `TTLGenericCompiler` as new ops are added. It would be more robust if thread ownership came from shared operation metadata or from the same syntax registration mechanism used by the compiler frontend. I think now is the best time to do this rather than doing a later refactoring.

7. Source handling and diagnostics. Because `@ttl.atom` reads the decorated function through `inspect.getsourcelines()`, atom functions must come from a Python source file that `inspect` can recover; functions defined through stdin or an interactive shell will not work. That may be an acceptable restriction, but it should be documented and reported with an atom-specific diagnostic. Also, setup evaluation through `eval(ast.unparse(...))` loses useful source locations for things like `PipeNet` construction, which means later diagnostics can point at generated `"<string>"` locations instead of user code. It would be better to preserve the original source locations through setup lifting. I don't expect perfect diagnostics in the initial PR, adding TODOs would be fine.

8. Tests. I think this needs more negative and IR-level coverage before being merged as a language feature. In particular: a mixed compute/data-movement statement should fail; different scalar argument values should either recompile or be rejected; defaults/keyword calls should have explicit coverage; unsupported atom body constructs should produce atom-specific errors; and there should be FileCheck coverage showing that each generated thread contains only the intended work.

Composability is one of the main new capabilities being enabled here -- so test coverage needs to expand beyond the current simple helper-inlining case. It would be good to cover nested helper chains, repeated calls to the same helper, helper calls under control flow, keyword/default argument behavior, PipeNet parameters, and at least one helper used inside a larger multi-PipeNet atom. Again, deferring with TODOs is fine (with issues), as long as there is a clear path to implementing whatever isn't yet supported here.

9. Diagnostics. Many failures currently look like they would produce generic unknown-variable or synthetic-thread errors. It would be much easier to use if the atom layer reported rules directly, e.g. "statement mixes compute and data movement work", "default parameters are not supported in `@ttl.atom`", "top-level atom parameters must be tensors or compile-time scalars", or "inline-only DFB parameter used in a top-level atom call". Again, this can be marked as TODO with issues.

10. Sim support. What is the plan for eventual `tt-lang-sim` support? I am a bit uncomfortable adding significant new language features without Kostas around, and I think there should at least be a rough sketch of how this can be supported in the simulator without replicating significant portions of frontend logic. Since `tt-lang-sim` shadows the compiler `ttl` module with the simulator namespace, it will need simulator exports for `atom`, `DFB`, and `make_dfb`, but the atom source parsing/inlining/splitting logic should ideally be shared and then consumed by separate compiler and simulator backends.

zoecarver:
> On naming, I think calling this @ttl.operation may be better if others agree (not a fan of atom, but could get used to it).

I think this is a good idea, and I'd like to do it, but I want to build consensus and let this feature be tested and iterated on as needed for a bit first. Maybe we can re-visit next week after folks have time to play with it? I also want to land the 4-5 PRs open for ttl.atom before modifying the top-level, publicly documented API. 

> Since it's a language change/addition, it would be good to explicitly write up the new semantics 

Sure happy to do this. And this document could evolve into a spec change proposal. 

> Scalar arguments and caching. It looks like scalar atom parameters are captured as compile-time values and emitted into the generated IR, but the compile cache key is still the tensor-oriented one from @ttl.operation.

I will add a test for this, we should cache the whole source, so I think it should be OK today, but again will test. If it's broken, I suspect it's also broken for traditional ttl.operations which would deserve a GH issue and maybe a follow up fix rather than blocking this. 

> Default parameter values are syntactically accepted

Good find. 

> Atom parameter ABI. It would help to write down what kinds of parameters an atom may accept and whether they are runtime values or compile-time values. My current understanding is: tensors become runtime tensor accessors, scalar values are compile-time constants, ttl.DFB parameters are inline-only, and PipeNet/DFB objects should not appear as top-level runtime parameters. If that is right, the decorator should validate unsupported signatures early instead of letting them fail later while compiling a generated thread body.

I am not sure ABI is the correct word here: atoms are purely syntactic, almost like macros in C. There is no binary interface as they are not separately compiled. 

The correct mental model, I think, is that everything is inlined and programs behave as traditional ttl.operations regarding what is compile time vs runtime. I think this gets at a place where we are underspecified today, regarding what is compile time vs runtime. 

The idea that the top level caller must not have DFB or pipenet parameters is sort of the implicit model right now, but you're right it's not enforced anywhere. How do you suggest we enforce this? If we want to vend ttl.atoms as a library, I don't see any way to do it other than at the call site. 

But stepping back further, why should they be forbidden? Our compiler can just inspect those python objects at runtime and determine what it needs to. If it doesn't work today, it should be a small lift, and I can't think of a reason to disallow it (I'd like your thoughts though). 

> each DFB acquire should have one inferred thread owner, and multiple consumer/producer roles should be rejected with an atom-specific diagnostic before the generated MLIR reaches the lower-level SPSC verifier. 

I disagree. In a similar vein to some of your other comments, I think we should try to share this logic with the mlir side of the compiler. MLIR also seems like a better place for this analysis. I don't see any benefit to attempting it earlier. 

> That would make the source-level error much easier to understand.

Why?

> ... rest of your comments ...

These are more concrete suggestions for the PR, which I will try to apply where appropriate, and push to TODOs an issues otherwise.

brnorris03:
>>each DFB acquire should have one inferred thread owner, and multiple consumer/producer roles should be rejected with an atom-specific diagnostic before the generated MLIR reaches the lower-level SPSC verifier.

>I disagree. In a similar vein to some of your other comments, I think we should try to share this logic with the mlir side of the compiler. MLIR also seems like a better place for this analysis. I don't see any benefit to attempting it earlier.

>>That would make the source-level error much easier to understand.

Why?

Because, based on past user complaints, having cryptic (to users) compiler errors is a negative experience. Of course I agree that MLIR analysis is better than python AST analysis. I just don't immediately see how this would translate into what happens when users write incorrect code now. If you already observe concrete, clear, actionable errors being produced in those cases (and no hangs), then there is no concern.

zoecarver:
> Because, based on past user complaints, having cryptic (to users) compiler errors is a negative experience. Of course I agree that MLIR analysis is better than python AST analysis. I just don't immediately see how this would translate into what happens when users write incorrect code now.

I think we should instead audit and update our mlir errors and make sure they are raised with appropriate source locations.

brnorris03:
> > Because, based on past user complaints, having cryptic (to users) compiler errors is a negative experience. Of course I agree that MLIR analysis is better than python AST analysis. I just don't immediately see how this would translate into what happens when users write incorrect code now.
> 
> I think we should instead audit and update our mlir errors and make sure they are raised with appropriate source locations.

I don't think that's an issue on the MLIR side -- the diagnostics line propagation is done by the frontend and it is fine for the usual case where we have the three user kernels, but I don't know how it applies here. I didn't see anything specific to ensure that line numbers are correctly attributed to pre-split user code, so I am guessing that errors will be referring to split code, which the user did not write -- I could be wrong.

python/ttl/atom.py
```diff
@@ -0,0 +1,554 @@
...
+        target_arch = _device_target_arch(args)
+
+        cache_key = _make_cache_key(
+            args,
+            fp32_dest_acc_en=opts["fp32_dest_acc_en"],
+            dst_full_sync_en=opts["dst_full_sync_en"],
+            target_arch=target_arch,
+            compiler_options=compiler_options,
+        )
+
```
brnorris03: Are scalar atom parameters intended to be supported as compile-time parameters? The docstring says atoms may take compile-time scalars, and `_compile_atom` captures `int`/`float` values into the generated thread bodies. If that is supported, the atom cache key needs to include scalar args, too. If not, can we detect them and reject them explicitly and add a negative test?

python/ttl/_src/atom_inline.py
```diff
@@ -0,0 +1,305 @@
...
+                raise ValueError(
+                    f"@ttl.atom: inlining {self.callee_name!r} into "
+                    f"{self.caller_name!r}: cannot assign to or delete "
+                    f"parameter {node.id!r} inside the callee body."
+                )
+            return ast.copy_location(copy.deepcopy(self.bindings[node.id]), node)
+        if node.id in self.rename_map:
+            new = ast.Name(id=self.rename_map[node.id], ctx=node.ctx)
+            return ast.copy_location(new, node)
+        return node
```
brnorris03: The current inlining parameter renaming is a bit too aggressive and not sufficiently scope-aware, for example renaming variables in nested function bodies to caller vars without considering the nested function's parameters.

```python
@ttl.atom()
def helper(net: ttl.PipeNet, block: ttl.DFB, pipe):
    def send(pipe):
        ttl.copy(block, pipe)
    net.if_src(send)
```

I think inlining `helper(outer_net, outer_block, outer_pipe)` currently would rewrite the copied callee body to:

```python
def send(pipe):
    ttl.copy(outer_block, outer_pipe)
outer_net.if_src(send)
```

The callback parameter `pipe` should still refer to the PipeNet callback argument, but it is replaced with the caller expression for the helper's `pipe` parameter.

A possible approach is to implement the inliner as a structured AST clone rather than direct `Name` replacement: collect callee-local definitions, assign fresh names where needed, update references through a scoped remapping table, and remove names declared by nested `FunctionDef`/`Lambda` nodes from the active parameter remapping before visiting their bodies. That should still allow non-shadowed outer helper parameters like `block` to remap, while preserving callback-local names. An AST-only unit test using `ast.unparse()` should be enough to cover this case.

python/ttl/_src/atom_split.py
```diff
@@ -0,0 +1,800 @@
...
+        """
+        threads: Set[str] = set()
+        for node in _iter_skip_nested_fns(stmt):
+            if isinstance(node, ast.Call):
+                t = self._classify_call(node)
+                if t in THREADS:
+                    threads.add(t)
+            elif isinstance(node, ast.BinOp) and isinstance(node.op, ast.MatMult):
+                threads.add("trisc")
+        if threads:
```
brnorris03, line 468: Should we reject ordinary statements that have more than one concrete thread owner before generating the per-thread bodies? (Or alternatively, could have a preprocessing "splitting" phase to manage that automatically when there is mixed-core statements.)

For example:

```python
def k(x, out):
    dst = out_cb.reserve()
    ttl.copy(ttl.exp(x), dst)
```

`ttl.exp(x)` is compute work and `ttl.copy(...)` is data movement. `_stmt_anchor_threads` returns both `trisc` and `ncrisc`, so the full statement is cloned into both generated bodies:

```python
# trisc
ttl.copy(ttl.exp(x), dst)

# ncrisc
dst = out_cb.reserve()
ttl.copy(ttl.exp(x), dst)
```

That leaves the TRISC body with a data-movement op and a use of `dst` whose reserve only exists on NCRISC. I think this should be an atom-level error before lowering (ordinary statements must have exactly concrete thread owner), at least as an initial approach, can be handled better later. If you agree, then also need a negative test -- or if this is supposed to be supported, a positive test.

python/ttl/_src/atom_split.py
```diff
@@ -0,0 +1,800 @@
...
+                    if bt and len(bt) == 1:
+                        return next(iter(bt))
+        return None
+
+    # --- post-check: forbid duplicate DM-side reserves ---
+
+    def _check_duplicate_producers(self) -> None:
+        for name, stmts in self._producers.items():
+            threads = self.block_threads.get(name, frozenset())
+            if "ncrisc" in threads and "brisc" in threads:
```
brnorris03: Possible bug -- this guard only rejects NCRISC+BRISC. A single `.reserve()`/`.wait()` whose block is written on both TRISC (`store`) and a DM thread (`ttl.copy`) is not caught -> two `cb_reserve_back` against one CB, resulting in runtime pointer corruption. 

Repro (and suggested regression if you fix): a block used by `blk.store(...)` and `ttl.copy(a, blk)` emits `shared_cb.reserve()` on both trisc and ncrisc bodies, no error. Fine to postpone with a TODO and issue.

python/ttl/_src/atom_split.py
```diff
@@ -0,0 +1,800 @@
...
+
+# ----- public API -----------------------------------------------------------
+
+
+@dataclass
+class SplitResult:
+    trisc_body: List[ast.stmt]
+    ncrisc_body: List[ast.stmt]
+    brisc_body: List[ast.stmt]
+    trisc_param_uses: Set[str] = field(default_factory=set)
```
brnorris03: Dead code?  `param_uses_for`, `_params_referenced`, and the `all_param_names` arg.

python/ttl/atom.py
```diff
@@ -0,0 +1,554 @@
...
+    ns = dict(scope)
+    ns.setdefault("make_dfb", make_dfb)
+    ns.setdefault("make_dataflow_buffer_like", make_dataflow_buffer_like)
+    ns.setdefault("ttl", _ttl)
+
+    dfbs: Dict[str, DataflowBuffer] = {}
+    nets: Dict[str, PipeNet] = {}
+    kept: List[ast.stmt] = []
+    for stmt in fn_def.body:
```
brnorris03, line 464: Possible bug -- lifting only walks top-level `fn_def.body`. A `make_dataflow_buffer_like` / `PipeNet` / `Pipe` declared inside a top-level `if`/`for`/`with` is never lifted -> not in `local_dfb_names`, not in the pipe graph -> its `.wait()`/`.reserve()` isn't recognized as a producer and duplicates onto all three threads.

Could address by doing something similar to `_validate_callee_for_inline`: scan the top-level body and raise if a setup-factory call appears below the top-level statement list. Possible repro: `if COMPILE_FLAG: scratch_cb = ttl.make_dataflow_buffer_like(...)`.

python/ttl/dataflow_buffer.py
```diff
@@ -60,22 +60,37 @@ def __init__(
...
         block_count: int,
+        dtype: Any = None,
     ):
         if len(shape) < 2:
             raise ValueError(f"DFB shape must have at least 2 dimensions, got {shape}")
         if block_count < 1 or block_count > 32:
             raise ValueError(f"block_count must be in range [1, 32], got {block_count}")
+        # A buffer's dtype has one source: a backing tensor or an explicit
+        # dtype. Supplying both is only valid when they resolve to the same type.
+        if dtype is not None and getattr(tensor, "dtype", None) is not None:
```
brnorris03, line 71: Minor -- unreachable through the public factories: `make_dfb` passes `tensor=None`, `make_dataflow_buffer_like` passes `dtype=None`, only reachable through direct constructor calls `DataflowBuffer(...)`, but no existing test for that? I'd add a test or remove the check.

python/pykernel/_src/kernel_ast.py
```diff
@@ -548,7 +548,12 @@ def _match_bool_op(lhs, rhs):
 
     def visit_BinOp(self, node):
         def materialize(value):
-            if not value:
+            # A host int operand (e.g. a shape/grid-derived bound) becomes an
+            # index constant so it can combine with SSA values. Matches how
+            # captured int constants are emitted at function entry.
+            if isinstance(value, int) and not isinstance(value, bool):
```
brnorris03, line 990: Thanks! Also fixes the old `if not value` bug where an int operand of `0` raised before.

brnorris03 (approved):

phizalev-TT (approved):

python/ttl/atom.py
zoecarver: Removed; added invalid test.

python/ttl/_src/atom_inline.py
zoecarver: fixed with deep copy + disallowing shadow naming.

python/ttl/_src/atom_split.py
zoecarver: Added error + invalid test.

python/ttl/_src/atom_split.py
zoecarver: `_check_producer_threads` now rejects any acquire used by more than one thread, not only NCRISC+BRISC. The regression at test/python/atom/test_atom_split.py:99 covers .reserve(), .store(), and ttl.copy() spanning TRISC/NCRISC. The scoped with reserve() form is also covered.

python/ttl/atom.py
zoecarver: This now just errors.

zoecarver: In the future maybe an opportunity to do tighter re-use.

python/ttl/dataflow_buffer.py
zoecarver: added a test

python/ttl/_src/atom_split.py
zoecarver: done

zoecarver:
Merging but happy to address anything else post commit.

PR #660, opened 2026-06-03
[`ttl.atom`] Introducing op library: composable atoms for common operations 

zoecarver:
These ops will 
* provide more comprehensive tests for real world patterns
* serve as examples for people to reference 
* eventually be high performance operations that can be dropped into models
* provide a foundation for our benchmark infrastructure 

The current ops already have some amount of composition with shared utilities for common patterns, but I'm hoping to eventually extend that (eg provide generic tree reductions) and formalize the API inference (eg to always provide dfb-level variants).

### DIFF
--- a/python/pykernel/_src/kernel_ast.py
+++ b/python/pykernel/_src/kernel_ast.py
@@ -548,7 +548,12 @@ class TTCompilerBase(PyKernelAstBase):
 
     def visit_BinOp(self, node):
         def materialize(value):
-            if not value:
+            # A host int operand (e.g. a shape/grid-derived bound) becomes an
+            # index constant so it can combine with SSA values. Matches how
+            # captured int constants are emitted at function entry.
+            if isinstance(value, int) and not isinstance(value, bool):
+                return arith.ConstantOp(IndexType.get(self.ctx), value).result
+            if value is None:
                 raise ValueError("Binary operands not found")
             if isinstance(value, OpView):
                 value = value.result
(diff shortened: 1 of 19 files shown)


zoecarver:
This is a nice example of the composition and simplicity these ops afford: https://github.com/tenstorrent/tt-lang/pull/660/changes#diff-df47da681fce358657a3151560fb11a53c3091f64af1333742612d26e4ddc89b

python/pykernel/_src/kernel_ast.py
```diff
@@ -548,7 +548,12 @@ def _match_bool_op(lhs, rhs):
 
     def visit_BinOp(self, node):
         def materialize(value):
-            if not value:
+            # A host int operand (e.g. a shape/grid-derived bound) becomes an
+            # index constant so it can combine with SSA values. Matches how
+            # captured int constants are emitted at function entry.
+            if isinstance(value, int) and not isinstance(value, bool):
```
arichinsTT, line 554: this whole materialize function feels a hacky, but host int operands being created here feels wrong. might be fine if it was shared with all numerical expressions nodes (unary and comparison), but why can't we instantiate the value when we visit it's node and we are just checking for null or grabbing the result here?

zoecarver: I feel like this is pretty standard compiler logic for "parser found a literal, turn it into an SSA constant," but I agree it should be eventually expanded to cover more AST nodes.

arichinsTT: I agree the logic is simple, but it should be handled at the literal parsing not during the parsing of the arguments of the binary operand

PR #661, opened 2026-06-04
Overhaul benchmarks

zoecarver:
Reworks `benchmarks/` from a single ad-hoc matmul sweep (with its own copy of the matmul kernels) into a small, composable suite with one shared harness, two benchmark kinds, and a unified driver.
  
The benchmark-local kernel copies are gone: benchmarks now measure the ttl.ops.* library directly, so the op library is the single source of truth (tests guard correctness, benchmarks measure those exact ops).

**Two kinds of benchmark**

  - e2e: wall-clock comparison of an op against a ttnn reference across a shape sweep, reported as a ttlang / reference ratio. Ops covered: matmul, rmsnorm, topk, flash_mla.
  - cycles: single-core Tracy / device-profiler cycle counts against a low-level tt-metal primitive, reported as a ttl / metal ratio. One run, no averaging. Currently just flash_shard (`ttl.ops.flash_mla` shard vs metal `compute_sdpa_chunk`).

  **Shared harness (`common/`)**

  A benchmark declares what to measure; the harness owns how. An e2e benchmark is a BenchSpec (name, CSV fields, cases, a run_case, a label fn, a device opener) and gets the device lifecycle, the timing loop, failure handling, CSV, ratio plotting, and a CLI for free, so the same spec drives both a standalone run and the driver. `common/` also provides best-of-N timing (enqueue back-to-back, sync once), a pcc helper, single + stacked ratio plots, and the Tracy device-profiler parsing for cycles benches.

```
  benchmarks/
    common/        shared, op-agnostic harness (timing, CSV, plot, sweep/CLI)
    e2e/           end-to-end benchmarks: each op vs a ttnn reference
    cycles/        single-core device-cycle A/Bs vs a metal primitive
      flash_shard/ ttl.ops.flash_mla shard vs metal compute_sdpa_chunk
    driver.py      runs every registered benchmark
```

  Every benchmark is runnable on its own (one shape selectable with --filter), and the driver sweeps them all into one stacked ratio PNG:

```
  python -m benchmarks.e2e.matmul --filter 8k        # one op, one shape
  TT_METAL_DEVICE_PROFILER=1 python -m benchmarks.driver --plot   # everything
```


brnorris03:
Why did this close? Is there a new replacement PR?

Issue #380, opened 2026-03-10
[external bug report][from claude] Variable assigned in if/else branches lost during tracing (KeyError in datamovement)


zoecarver:
## Summary

When a variable is assigned inside both branches of an `if/else` statement in a `@ttl.datamovement()` function, the tracer/compiler loses track of it. Accessing the variable after the if/else block raises `KeyError`.

This works correctly in the simulator (which evaluates Python directly) and in `@ttl.compute()` functions. The bug is specific to if/else in datamovement on HW.

## Reproducer

```python
import torch
import ttnn
import ttl

@ttl.kernel(grid=(2, 1))
def if_dm_kernel(inp, out):
    inp_dfb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), buffer_factor=2)
    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), buffer_factor=2)

    @ttl.compute()
    def compute_fn():
        x, y = ttl.core(dims=2)
        with inp_dfb.wait() as a, out_dfb.reserve() as o:
            if x == 0:
                o.store(ttl.math.exp(a))
            else:
                o.store(ttl.math.tanh(a))

    @ttl.datamovement()
    def dm_read():
        x, y = ttl.core(dims=2)
        with inp_dfb.reserve() as blk:
            if x == 0:
                tx = ttl.copy(inp[0, 0], blk)
            else:
                tx = ttl.copy(inp[0, 1], blk)
            tx.wait()  # <-- KeyError: 'tx'

    @ttl.datamovement()
    def dm_write():
        x, y = ttl.core(dims=2)
        with out_dfb.wait() as blk:
            if x == 0:
                tx = ttl.copy(blk, out[0, 0])
            else:
                tx = ttl.copy(blk, out[0, 1])
            tx.wait()  # <-- KeyError: 'tx'

if __name__ == "__main__":
    device = ttnn.open_device(device_id=0)
    inp = ttnn.from_torch(torch.full((32, 64), 0.5, dtype=torch.bfloat16),
                          dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
                          device=device, memory_config=ttnn.DRAM_MEMORY_CONFIG)
    out = ttnn.from_torch(torch.zeros((32, 64), dtype=torch.bfloat16),
                          dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
[...]

PR #662, opened 2026-06-04
raw element read/write lowering and fp numeric

### LINKED ISSUE #572: [ttl] element_read scan kernel compares bf16 values as raw i32 bit patterns

arichinsTT:
### Problem description
                                                                                                                                                                                                                             
`ttl.raw_element_read` and `ttl.raw_element_write` had no lowering through `convert-ttl-to-ttkernel`. Float point comparisons are lowered to soft float bit wise operates greater than and eq/neq                                   
                                                                                                                                                                                                                             
### What's changed
                                                                                                                                                                                                                             
#### Raw element access lowering
                                                                                                                                                                                                                             
`RawElementReadLowering` and `RawElementWriteLowering` are added to the `convert-ttl-to-ttkernel` pass. Both lower to a CB pointer acquisition, an L1 typed pointer cast, a flat element offset, and `ttkernel.load_from_l1` / `ttkernel.store_to_l1`.                                                    
                                                                                                                                                                                                                             
```mlir
// TTL
%val = ttl.raw_element_read %block[%r, %c] : tensor<1x1x!ttcore.tile<32x32, f32>> -> f32
                                                                                                                                                                                                                             
// TTKernel
%ptr  = ttkernel.get_read_ptr(%cb)
%l1   = ttkernel.reinterpret_cast<tt_l1_ptr uint32_t*>(%ptr) : (i32) -> !ttkernel.l1_addr_ptr
%bits = ttkernel.load_from_l1(%l1, %offset) : (!ttkernel.l1_addr_ptr, i32) -> i32
// Temporary bridge: keeps the FP type visible to downstream arith ops while
// the integer storage bits are preserved for ttl-lower-scalar-cmpf.
%val  = builtin.unrealized_conversion_cast %bits : i32 to f32                                                                                                                                                                                    
                                                                                                                                                                                                                             
// TTL
ttl.raw_element_write %block[%r, %c], %val : tensor<1x1x!ttcore.tile<32x32, f32>>, f32

// TTKernel
%bits = ...  // integer storage bits materialized from %val
%ptr  = ttkernel.get_write_ptr(%cb)
%l1   = ttkernel.reinterpret_cast<tt_l1_ptr uint32_t*>(%ptr) : (i32) -> !ttkernel.l1_addr_ptr
ttkernel.store_to_l1(%bits, %l1, %offset) : (i32, !ttkernel.l1_addr_ptr, i32) -> ()
```

The read/write pointer (get_read_ptr vs get_write_ptr) is resolved by tracing the block operand through ttl.attach_cb and unrealized_conversion_cast wrappers to its originating ttl.cb_wait or ttl.cb_reserve.                                                                                                                                                                                                                                  

Tiled element offset. For blocks with a !ttcore.tile element type the flat offset uses the hardware face layout: a 32×32 tile is stored as four 16×16 faces in row-major face order.                                                                                                                                                                                                             

face 0  rows [0–15],  cols [0–15]  → base offset    0                                                                                                                                                                                            
face 1  rows [0–15],  cols [16–31] → base offset  256                                                                                                                                                                                            
face 2  rows [16–31], cols [0–15]  → base offset  512
face 3  rows [16–31], cols [16–31] → base offset  768                                                                                                                                                                                            
                                                                                                                                                                                                                             
For row-major blocks (scalar element type) the offset is a standard row-major linearization. Both layouts support rank-1 and ND blocks.                                                                                                                                                                                        
                                                                                                                                                                                                                             
Scalar float comparison lowering (ttl-lower-scalar-cmpf)
                                                                                                                                                                                                                             
A new pass, ttl-lower-scalar-cmpf, is inserted after convert-ttl-to-ttkernel and before ttkernel-insert-inits. It consumes the unrealized_conversion_cast bridge values produced by RawElementReadLowering and lowers arith.cmpf to the matching TTKernel soft-float comparison op, keeping integer storage bits as the payload throughout.
```mlir                                                                                                                                                                                                                             
// After convert-ttl-to-ttkernel
%a    = builtin.unrealized_conversion_cast %bits_a : i32 to f32
%b    = builtin.unrealized_conversion_cast %bits_b : i32 to f32
%cond = arith.cmpf ogt, %a, %b : f32
                                                                                                                                                                                                                             
// After ttl-lower-scalar-cmpf
%cond = ttkernel.float32_greater(%bits_a, %bits_b) : (i32, i32) -> i1                                                                                                                                                                            
```
bf16 follows the same pattern via ttkernel.bfloat16_greater with i16 storage bits. arith.select on the comparison result is supported and passes through unmodified; arith.addf and other float arithmetic on raw-read values are not yet supported and will fail at lowering time.                                                                                                                                                                                            
                                                                                                                                                                                                                             
Python DSL                                                                                                                                                                                                                                       
                                                                                                                                                                                                                             
\> and \< comparisons on ttl.raw_element_read results are now supported in @ttl.datamovement kernels. The AST visitor emits arith.cmpf ogt/olt; ttl-lower-scalar-cmpf lowers to the TTKernel soft-float op. An implicit f32 -> bf16 truncation is inserted in operators.py when a raw-read f32 value flows into a bf16 context.                                                                                                                                                                                                                 
                                                                                                                                                                                                                             
### Submodule bump  
                                                                                                                                                                                                                             
third-party/tt-mlir is bumped to pick up ttkernel.reinterpret_caste bit widths, ttkernel.float32_greater and ttkernel.bfloat16_greater.
                                                                                                                                                                                                                             
### Tests
                                                                                                                                                                                                                             
MLIR FileCheck tests cover tiled and row-major layouts (f32 and bf16), all four tile-face corners and boundaries, static and dynamic coordinates, rank-1 and rank-3 blocks, the read-write chain (verifying integer bits flow directly                                                                                                                                                                    
without arith.bitcast), and arith.select on comparison results. Invalid cases cover unsupported predicates and non-resolable operands.                                                                                                                                                                                   
                                                                                                                                                                                                                             
Python pytest coverage exercises 7 patterns including negative and mixed-sign values to validate float32_greater sign semantics at the hardware level. Compile-only tests cover argmax, topk, and kv-cache patterns that motivated the feature.                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                             
Fixes #572.


test/python/raw_element_argmax_bf16.py
```diff
@@ -0,0 +1,116 @@
...
+Mirrors raw_element_argmax.py but with bf16 input/output tensors. The
+lowering must produce 16-bit L1 pointer access (uint16_t) and
+comparisons must route through bfloat16_greater().
+
+Compile-only test.
+"""
+
+import os
+
+os.environ["TTLANG_COMPILE_ONLY"] = "1"
```
zoecarver: Can we run some of these and compare against the ttnn variant?

zoecarver: Oops, I see `test_raw_element.py` now, but I'd like to see those tests compare against a ttnn api call.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -1084,6 +1084,278 @@ struct FuncKernelFinalize : OpRewritePattern<FuncOp> {
...
+/// Determine whether the original (pre-conversion) block came from a
+/// CBWaitOp (reader, use get_read_ptr) or CBReserveOp (writer, use
+/// get_write_ptr).
+static bool isBlockFromCBWait(Value block) {
+  block = traceUnrealizedCasts(block);
+  if (auto viewLike = llvm::dyn_cast_if_present<ViewLikeOpInterface>(
+          block.getDefiningOp())) {
+    return llvm::isa<CBWaitOp>(viewLike.getOperation());
+  }
+  return false;
```
zoecarver: I think this needs to be stricter. What if there's an attach cb call in between? Maybe this can fail if it can't find either?

test/python/test_raw_element.py
```diff
@@ -0,0 +1,835 @@
...
+
+    Reads the reference zero value from row 1 col 0 and the sentinel
+    from row 1 col 1 to avoid f32/bf16 type mismatch with constants.
+    """
+    inp_dfb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=2)
+    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+    @ttl.compute()
+    def compute():
+        pass
```
zoecarver: Two more ideas for test: 
* one where compute is not `pass`, so maybe do some math, store to output cb, then element_read that. 
* maybe one where we `copy` in the reader thread, then element_write, then do some compute on it.

zoecarver:
Overall looks great! Can't wait to use it!

test/python/raw_element_argmax_bf16.py
arichinsTT: for performance?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
arichinsTT: I agree, I'm not as familiar with the variation in patterns, but how do we feel about having a general ttl utility function for this?

test/python/test_raw_element.py
arichinsTT: I can def whip up an example, just realized the tile packing example didn't make it in, would that be a good example?

arichinsTT: also if you have a kernel in mind, even it's obscure, feel free to share

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -1084,6 +1084,278 @@ struct FuncKernelFinalize : OpRewritePattern<FuncOp> {
...
 };
 
+//===----------------------------------------------------------------------===//
+// Raw Element Access Lowering
+//===----------------------------------------------------------------------===//
+
+/// Determine whether the original (pre-conversion) block came from a
+/// CBWaitOp (reader, use get_read_ptr) or CBReserveOp (writer, use
+/// get_write_ptr).
+static bool isBlockFromCBWait(Value block) {
```
brnorris03: Suggest tracing through `ttl.attach_cb` here before checking for `CBWaitOp`. The Python DFB API returns `attach_cb(cb_wait(...))`, and `isCBAcquireView` was updated for that form, but this helper still returns false for it. That means `raw_element_read(rblk, ...)` from Python lowers through `get_write_ptr`. Please add an `attach_cb` case and a TTLToTTKernel test that covers `attach_cb` around `cb_wait`.

lib/Dialect/TTL/Transforms/LowerScalarCmpF.cpp
```diff
@@ -0,0 +1,140 @@
...
+        if (bitWidth == 32) {
+          result = ttk::Float32GreaterOp::create(
+              builder, loc, builder.getI1Type(), rhsInt, lhsInt);
+        } else {
+          result = ttk::Bfloat16GreaterOp::create(
+              builder, loc, builder.getI1Type(), rhsInt, lhsInt);
+        }
+        break;
+      }
+      case arith::CmpFPredicate::OEQ: {
```
brnorris03: Since the frontend lowers the Python `==` / `!=` to `arith.cmpf oeq/one`, I think the lowering must preserve `arith.cmpf` semantics or reject the operation. Right now it is not correct for signed 0 or NaNs. I think that it should not just reinterpret a normal Python float comparison as exact-bit comparison. This may require a spec change, perhaps, but I would really like to avoid going down the slippery slope of implicitly redefining both top-level language (Python) and existing dialect semantics (arith) just to be able to reuse some part of a dialect.

If exact-bit comparison is the intended semantics, then I think it should be a separate API/op; otherwise this needs numeric equality or should be rejected like `>=` / `<=`. @phizalev-TT what is your take on this?

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -150,14 +150,17 @@ inline std::optional<mlir::Type> getTileElementType(mlir::Type type) {
   return std::nullopt;
 }
 
-/// Return true when `tensor` was directly acquired from a CB via
-/// ttl.cb_wait or ttl.cb_reserve (the only two ViewLikeOpInterface
-/// implementations whose view source is a CircularBufferType).
-/// Unlike getAttachedCB, this rejects ttl.attach_cb, tensor.extract, and
-/// tensor.extract_slice chains -- it only traces through unrealized
-/// conversion casts.
+/// Return true when `tensor` was acquired from a CB via ttl.cb_wait or
```
brnorris03, line 192: The raw-element read/write verifier should be more specific. Reads require `cb_wait` and writes require `cb_reserve`, but this accepts either for both the raw element variants. For example, `raw_element_write` on a wait block would store through the read side of the DFB, which violates DFB producer/consumer semantics.

lib/Dialect/TTL/Transforms/LowerScalarCmpF.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace {
+
+namespace ttk = mlir::tt::ttkernel;
+
+/// Resolve a float-typed SSA value to its underlying signless integer bit
+/// pattern. Two sources are handled:
+///   1. unrealized_conversion_cast(iN -> fN) from RawElementReadLowering.
+///   2. arith.constant <float> -- materializes the bit pattern as an integer.
+/// Returns a null Value on failure.
```
brnorris03: Would be good to share this bit-materialization logic with `materializeIntBits` in `ConvertTTLToTTKernel.cpp` and return `FailureOr<Value>` instead of a null `Value`. Both helpers now encode the same cast/constant bit extraction rules, and the failure contract is implicit.

test/ttlang/Conversion/TTLToTTKernel/scalar_cmpf_invalid.mlir
```diff
@@ -0,0 +1,78 @@
+// Negative tests for ttl-lower-scalar-cmpf: unsupported predicates emit errors.
+// RUN: not ttlang-opt --ttl-lower-scalar-cmpf --split-input-file %s 2>&1 | FileCheck %s
```
brnorris03: This should use `--verify-diagnostics` with `expected-error @below` instead of piping stderr through FileCheck -- see all other negative tests.

test/python/raw_element_argmax_bf16.py
brnorris03: I think mainly for correctness -- most of the python lit tests are runtime tests.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03: Maybe worth reusing or extending the existing DFB-view tracing helpers here rather than adding another local trace. `getCBFromView()` already handles `ttl.attach_cb`, but it returns only the CB and loses whether the acquire was `CBWaitOp` or `CBReserveOp` (shouldn't be too hard to add more information to the return). This code needs that acquire kind to choose `get_read_ptr` vs `get_write_ptr`. A shared helper that traces through `unrealized_conversion_cast` / `ttl.attach_cb` and returns the underlying acquire op would keep this consistent with the verifier and avoid another slightly different tracing implementation.

lib/Dialect/TTL/Transforms/LowerScalarCmpF.cpp
arichinsTT: dropping operations for now

test/python/raw_element_argmax_bf16.py
zoecarver: yeah just for correctness and as a reference for how it should be used, but benchmarks later would be great!

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
zoecarver: Sure, a utility is fine, but I think we should make sure we classify as wait or reserve and if we can't do that, return a failure case.

test/python/test_raw_element.py
zoecarver: I think the two I mentioned are sufficient. If I have more, I'll send them to you.

test/python/raw_element_argmax_bf16.py
arichinsTT: done in the e2e pytests, currently xfailed due to ssa scalar value not escaping if scope, but it is there

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
arichinsTT: I think this is fixed, tho I kept the old api and just used a nullptr check

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1003,6 +1003,82 @@ def writer():
...
+### Comparison Operators
+
+Values from `raw_element_read` support scalar float comparisons:
+
+| Python operator | Supported |
+| :---- | :---- |
+| `>` | Yes |
+| `<` | Yes |
+| `==` | Yes |
+| `!=` | Yes |
```
brnorris03: I think this is deferred now, so should they be marked as not yet supported?

test/ttlang/Dialect/TTL/IR/raw_element_ops.mlir
brnorris03, line 1: a couple more tests to consider (haven't actually checked the CHECK lines so may need adjusting):
```
  // Read f32 from an attach_cb-wrapped cb_wait block (the form the Python
  // frontend emits). Must trace through attach_cb to get_read_ptr.
  // CHECK-LABEL: func.func @read_attach_cb_wait
  // CHECK: %[[CB:.*]] = ttkernel.get_compile_time_arg_val(0)
  // CHECK: %[[PTR:.*]] = ttkernel.get_read_ptr(%[[CB]])
  // CHECK: %[[L1:.*]] = ttkernel.reinterpret_cast<tt_l1_ptr uint32_t*>(%[[PTR]]) : (i32) ->
  !ttkernel.l1_addr_ptr
  // CHECK: ttkernel.load_from_l1(%[[L1]], {{.*}}) : (!ttkernel.l1_addr_ptr, i32) -> i32
  module {
    func.func @read_attach_cb_wait()
        attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
      %cb = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
      %wait = ttl.cb_wait %cb : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32,
  f32>>
      %block = ttl.attach_cb %wait, %cb
          : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>)
            -> tensor<1x1x!ttcore.tile<32x32, f32>>
      %c5 = arith.constant 5 : index
      %val = ttl.raw_element_read %block[%c0, %c5] : tensor<1x1x!ttcore.tile<32x32, f32>> -> f32
      func.return
    }
  }

  // -----

  // Write f32 to an attach_cb-wrapped cb_reserve block. Must trace through
  // attach_cb to get_write_ptr.
  // CHECK-LABEL: func.func @write_attach_cb_reserve
  // CHECK: %[[CB:.*]] = ttkernel.get_compile_time_arg_val(0)
  // CHECK: %[[PTR:.*]] = ttkernel.get_write_ptr(%[[CB]])
  // CHECK: %[[L1:.*]] = ttkernel.reinterpret_cast<tt_l1_ptr uint32_t*>(%[[PTR]]) : (i32) ->
  !ttkernel.l1_addr_ptr
  // CHECK: ttkernel.store_to_l1({{.*}}, %[[L1]], {{.*}}) : (i32, !ttkernel.l1_addr_ptr, i32) -> ()
  module {
    func.func @write_attach_cb_reserve(%val: f32)
        attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
      %cb = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>
      %res = ttl.cb_reserve %cb : <[1, 1], !ttcore.tile<32x32, f32>, 2> -> tensor<1x1x!ttcore.tile<32x32,
  f32>>
      %block = ttl.attach_cb %res, %cb
          : (tensor<1x1x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 1], !ttcore.tile<32x32, f32>, 2>)
            -> tensor<1x1x!ttcore.tile<32x32, f32>>
      ttl.raw_element_write %block[%c0, %c0], %val : tensor<1x1x!ttcore.tile<32x32, f32>>, f32
      func.return
    }
  }
```

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1003,6 +1003,82 @@ def writer():
...
+```python
+@ttl.datamovement()
+def dm_write():
+    with inp_dfb.wait() as rblk:
+        with out_dfb.reserve() as wblk:
+            max_val = ttl.raw_element_read(rblk, 0, 0)
+            for c in range(1, 32):
+                val = ttl.raw_element_read(rblk, 0, c)
+                if val > max_val:
+                    max_val = val
```
brnorris03, line 1062: This is actually not yet supported by the frontend (issue #380), so may be worth mentioning somewhere in the example section that this only works with the simulator (if it does) at the moment, with future compiler support anticipated. Or consider adding a smaller simpler example that does work today.

brnorris03 (approved):
Looks great, a couple of small suggestions to consider, otherwise ready to merge IMO.

Issue #663, opened 2026-06-05
Make ttl-verify-dfb-spsc domain-aware for disjoint DFB consumers


brnorris03:
## Problem

`ttl-verify-dfb-spsc` currently enforces SPSC by grouping every
`ttl.cb_reserve` and `ttl.cb_wait` for a `cb_index` by enclosing
`ttl.kernel_thread` function. That is conservative when multiple thread
functions touch the same DFB but their execution domains over `ttl.launch_grid`
are disjoint.

The runtime contract is per physical CB at a launch coordinate: at any
coordinate, a `cb_index` must have at most one active producer thread and at
most one active consumer thread. The current structural check rejects some IR
that still satisfies that contract.

This is distinct from #581. Auto-splitting is the right repair for true fan-out
where multiple consumers can overlap on the same coordinate. In the disjoint
domain case, no additional DFB is required if the verifier can prove the
consumer domains do not overlap.

## Minimal reproducer

Main commit used for the source reproducer:

```text
8eb21608 [ttl] Use receiver-authored SRAM address tables for scalable PipeNet collectives(#620) (#622)
```

This reproducer uses no PipeNets. One producer runs on every launched core.
Two consumer functions wait on the same `cb_index`, but one is guarded by
`core_x == 0` and the other by `core_x == 1`.

```mlir
module attributes {ttl.launch_grid = [2 : i64, 1 : i64]} {
  func.func @producer() attributes {ttl.kernel_thread = #ttkernel.thread<noc>} {
    %cb = ttl.bind_cb {cb_index = 0, block_count = 2}
        : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
    %slot = ttl.cb_reserve %cb
        : <[1, 1], !ttcore.tile<32x32, bf16>, 2>
        -> tensor<1x1x!ttcore.tile<32x32, bf16>>
    ttl.cb_push %cb : <[1, 1], !ttcore.tile<32x32, bf16>, 2>
    func.return
  }

  func.func @consumer_x0() attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
    %cb = ttl.bind_cb {cb_index = 0, block_count = 2}
        : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
    %core_x = ttl.core_x : index
    %zero = arith.constant 0 : index
    %is_x0 = arith.cmpi eq, %core_x, %zero : index
    scf.if %is_x0 {
[...]

PR #664, opened 2026-06-05
[ttl] Verify DFB SPSC per launch node, not per thread (#663)

brnorris03:
### Problem description
DFB SPSC verification was thread-global, so programs with one DFB used by multiple threads on disjoint launch nodes were rejected even though each launched node still had one producer and one consumer. This problem required significant overuse of DFBs to get around the verifier errors.

### What's changed
The verifier now uses a shared launch-node domain analysis that propagates finite `ttl.launch_grid` domains through structured predicates and PipeNet scopes. (The bulk of the code changes are extracting the domain analysis into a shared header/library, so that it can be used by both the pipes and dfb verifiers.) DFB participants are checked for producer or consumer overlap per launched node, unknown domains remain conservative, and malformed domain metadata forces pass failure.

### Checklist


include/ttlang/Dialect/TTL/Transforms/LaunchNodeDomainAnalysis.h
brnorris03, line 1: This and the corresponding cpp file were extracted from `TTLVerifyPipeNetGuards.cpp` so that the domain utilities can be reused in `TTLVerifyDFBSPSC.cpp`. It is not new functionality, but it was cleaned up substantially after being extracted.

lib/Dialect/TTL/Transforms/LaunchNodeDomainAnalysis.cpp
```diff
@@ -0,0 +1,688 @@
...
+    src = it->second;
+  }
+  if (auto it = netDestinationDomains.find(netId);
+      it != netDestinationDomains.end()) {
+    dst = it->second;
+  }
+  return src.unionWith(dst);
+}
+
+LogicalResult LaunchNodeDomainState::initialize(ModuleOp module) {
```
zoecarver: nit: now always returns success

zoecarver (approved):

phizalev-TT (approved):
LGTM!

lib/Dialect/TTL/Transforms/LaunchNodeDomainAnalysis.cpp
brnorris03: Changed to void.

PR #670, opened 2026-06-06
[ttl] atom inlining + fused flash-MLA, with sound DFB index reuse

zoecarver:
* Allows `ttl.atom`s to define dfbs locally
* Provides a flash mla that is fully fused (4 atoms called in one top level caller to produce a single program with DFBs to send data between atoms)
* Improves re-use:

To fit one L1, DFB index reuse runs in ttl-finalize-dfb-indices as a module pass: two DFBs share a CB index only with the same producer thread, same consumer thread, same element type, and disjoint pop-bounded lifetimes in both threads' program order. Capacity is a >= check (slot sized to the largest member), user and compiler DFBs share slots, pipe-attached DFBs never reuse (destination blocks are sender-slot addressed). Moves reach the runtime via ttl.dfb_index_map. Compiler intermediates drop to one block (compute-local). 

Tests: atom inline/split pytests, flash sweep over both chunk sizes, lit for reuse classes/lifetimes/compaction.

Perf report: 
```
  ┌─────────┬───────────────────────┬────────────┬────────────┬─────────────┐
  │ context │ baseline (chain sk=2) │ fused sk=1 │ fused sk=2 │   verdict   │
  ├─────────┼───────────────────────┼────────────┼────────────┼─────────────┤
  │ ctx-512 │ 12×                   │ 5.85×      │ 5.34×      │ 2.2× better │
  ├─────────┼───────────────────────┼────────────┼────────────┼─────────────┤
  │ ctx-1k  │ 10×                   │ 6.07×      │ 5.15×      │ 1.9× better │
  ├─────────┼───────────────────────┼────────────┼────────────┼─────────────┤
  │ ctx-2k  │ 7×                    │ 6.97×      │ 5.72×      │ 1.2× better │
  ├─────────┼───────────────────────┼────────────┼────────────┼─────────────┤
  │ ctx-8k  │ 8×                    │ 12.87×     │ 8.35×      │ ~par        │
  ├─────────┼───────────────────────┼────────────┼────────────┼─────────────┤
  │ ctx-32k │ 13×                   │ 21.13×     │ 13.25×     │ ~par        │
  └─────────┴───────────────────────┴────────────┴────────────┴─────────────┘
```

### DIFF
--- a/benchmarks/common/harness.py
+++ b/benchmarks/common/harness.py
@@ -74,7 +74,8 @@ def sweep(spec: BenchSpec, *, filter: Optional[str] = None, warmup=3, runs=5) ->
             rows.append(row)
             print(_line(spec, row), flush=True)
     finally:
-        ttnn.close_device(device)
+        if device is not None:
+            ttnn.close_device(device)
     return rows
 
 
--- a/benchmarks/common/report.py
+++ b/benchmarks/common/report.py
@@ -11,9 +11,20 @@ import torch
 
 
 def pcc(result, ref) -> float:
-    """Pearson correlation between a result tensor and its reference."""
+    """Pearson correlation between a result tensor and its reference.
+
+    Raises instead of returning nan so a kernel that never wrote its output
+    (constant tensor) or produced nan/inf is reported as such, not as a
+    silent nan ratio.
+    """
     a = result.flatten().float()
     b = ref.flatten().float()
+    if not torch.isfinite(a).all():
+        raise ValueError("pcc: result contains nan/inf")
+    if a.max() == a.min():
+        raise ValueError(
+            f"pcc: result is constant ({a.min().item()}); output likely never written"
+        )
     return torch.corrcoef(torch.stack([a, b]))[0, 1].item()
 
 
--- a/benchmarks/e2e/flash_mla.py
+++ b/benchmarks/e2e/flash_mla.py
@@ -9,9 +9,10 @@ deepseek_v3 MLA demo. Cases:
 
   - ``ctx-*``: single-user decode (1 user, 64 heads) at a fixed 32k KV cache,
     sweeping the decode position so the attended context grows 512 -> 32k. The
-    ttl chain (shard -> tree_reduce -> normalize) processes exactly the attended
-    length, is PCC-checked against a torch MLA golden, and the ratio is taken
-    against the ttnn paged decode reading the same position via ``cur_pos``.
+    fused single-kernel ttl op (shard + tree_reduce + normalize inlined into one
+    launch) processes exactly the attended length, is PCC-checked against a torch
+    MLA golden, and the ratio is taken against the ttnn paged decode reading the
+    same position via ``cur_pos``.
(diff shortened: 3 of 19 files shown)


brnorris03:
I am getting near 0 and nan PCC values for flash-mla on qb (pr661's PCC is fine), also some performance degradation vs pr661 in a few cases. Summary is [here](https://gist.github.com/brnorris03/9a5f1ca852729c37abd5a07c4f3ec824) -- what am I doing wrong?

brnorris03:
I don't think this optimization is safe in general; for example, in the test below, two logical DFBs are incorrectly aliased (use the same underlying physical DFB). This is because `_reuse_inlined_dfb_indices` computes liveness from direct DFB variable names only, not by following `wait()` results to later consumers (unlike `ttl-insert-cb-sync`). This can result in reusing a physical DFB before the logical value is popped. I think this name-base local liveness analysis could only be always correct for atom-local DFBs, not any used across threads. Also normally I would implement such analysis in MLIR, is there a specific reason for doing it in the frontend? 

```python
#

"""Runtime regressions for @ttl.atom scratch DFB reuse.

The helper atom declares scratch DFBs whose source spans do not capture their
runtime ownership after atom splitting. Reusing those DFB indices causes a
consumer to read the wrong tile from the shared queue.
"""

import pytest
import torch

import ttl

ttnn = pytest.importorskip("ttnn", exc_type=ImportError)

from ttlang_test_utils import assert_allclose, to_l1


@ttl.atom()
def _forward_then_add(in_tensor, out_tensor):
    """Move a tile through DM-produced and compute-produced scratch DFBs."""
    dm_to_compute = ttl.make_dfb("bf16", shape=(1, 1), block_count=2)
    compute_to_dm = ttl.make_dfb("bf16", shape=(1, 1), block_count=2)

    dm_block = dm_to_compute.reserve()
    ttl.copy(in_tensor[0:1, 0:1], dm_block)

    input_block = dm_to_compute.wait()
    output_block = compute_to_dm.reserve()
    one_block = ttl.block.fill(1.0, shape=input_block.shape)
    output_block.store(ttl.add(input_block, one_block))

    result_block = compute_to_dm.wait()
    ttl.copy(result_block, out_tensor[0:1, 0:1])


@ttl.atom(grid=(1, 1), options="--ttl-relax-dfb-spsc")
def atom_one_inline_site(in_tensor, out_tensor):
    _forward_then_add(in_tensor, out_tensor)


@pytest.mark.parametrize("input_value", [2.0, 7.0])
def test_inlined_scratch_reuse_keeps_runtime_queues_distinct(device, input_value):
    tile = ttnn.TILE_SIZE
    in_t = torch.full((tile, tile), input_value, dtype=torch.bfloat16)
    expected = in_t.float() + 1.0

    in_tensor = to_l1(in_t, device)
    out_tensor = to_l1(torch.zeros(tile, tile, dtype=torch.bfloat16), device)

    atom_one_inline_site(in_tensor, out_tensor)

    got = ttnn.to_torch(out_tensor).reshape(tile, tile).float()

    assert_allclose(got, expected, rtol=0, atol=0)
```

PR #676, opened 2026-06-09
[ttl] Support scaled matmul recurrence lowering

### LINKED ISSUE #665: [FEATURE] Support scaled additive recurrence with a matmul addend (out = scale*acc + a@b)
## Feature Description

Support accumulation of the form `out = scale * acc + (a @ b)`, where a matmul
result is added to a *scaled* prior value. The matmul-accumulator lowering
currently requires the accumulator to trace to a CB-attached compute input, so
the scale factor on `acc` makes the fused accumulator an SSA value and the
lowering asserts.

## Use Case

The flash-attention online-softmax output recurrence is exactly this shape:

    o_state = alpha * o_old + (exp_scores @ v)

where `alpha = exp(m_old - m_new)` rescales the running output each K-chunk.
The l-state update `l = alpha * l_old + chunk_sum` is the same scaled-additive
shape without the matmul. These are the core of any flash/streaming-softmax
kernel; today each requires a hand-allocated scratch DFB to stage the scaled
term so the matmul stays accumulator-free.

## Example Reproducer

`out = scale * c + (a @ b)` in a single statement (grid 1x1, 1x2 @ 2x1 tiles):

```python
@ttl.operation(grid=(1, 1))
def scaled_matmul_acc(a, b, c, scale, out):
    a_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 2), block_count=2)
    b_cb = ttl.make_dataflow_buffer_like(b, shape=(2, 1), block_count=2)
    c_cb = ttl.make_dataflow_buffer_like(c, shape=(1, 1), block_count=2)
    scale_cb = ttl.make_dataflow_buffer_like(scale, shape=(1, 1), block_count=2)

    @ttl.compute()
    def compute():
        a_blk = a_cb.wait(); b_blk = b_cb.wait()
        c_blk = c_cb.wait(); s_blk = scale_cb.wait()
        ow = out_cb.reserve()
        ow.store(ttl.add(ttl.mul(s_blk, c_blk), a_blk @ b_blk))   # out = scale*c + a@b

    @ttl.datamovement()
    def dm():
        ab = a_cb.reserve(); ttl.copy(a[0:1, 0:2], ab)
        bb = b_cb.reserve(); ttl.copy(b[0:2, 0:1], bb)
        cb = c_cb.reserve(); ttl.copy(c[0:1, 0:1], cb)
        sb = scale_cb.reserve(); ttl.copy(scale[0:1, 0:1], sb)
        ob = out_cb.wait(); ttl.copy(ob, out[0:1, 0:1])
```

## Current Behavior

Compilation aborts:

```
LowerMatmulCompute.cpp:91: generateMatmulCompute(...):
Assertion `accIdx && *accIdx < op.getInputs().size() &&
"accumulator must trace to a compute input"' failed.
```

`add(mul(scale, c), a@b)` is fused into a matmul-with-accumulator whose
accumulator is `mul(scale, c)` -- an SSA value, not a compute input.

## Workaround

Stage the scaled term in its own DFB so the add reads a CB-attached value and
the matmul carries no accumulator:

```python
corr = corr_cb.reserve(); corr.store(ttl.mul(s_blk, c_blk))
ow.store(ttl.add(corr_cb.wait(), a_blk @ b_blk))
```

This costs one extra DFB per scaled-matmul recurrence.

## Proposed Solution

Recognize `out = scale * acc + (a @ b)` (and the matmul-free `scale * acc + x`)
as a scaled additive recurrence. Either:

1. lower the scale as a pre-step into the matmul accumulator slot so the
   accumulator traces to a CB-attached value, or
2. when the fused accumulator does not trace to a compute input, fall back to
   materializing it through compiler-managed DFB state instead of asserting.

## Additional Context

Relates to the accumulation redesign (parent #638): #646 (non-additive
combiners -- the sibling `m = max(m_old, chunk_max)` state), #650 (synthesize
explicit DFB state fallback). The assert is in
`lib/Dialect/TTL/Transforms/LowerMatmulCompute.cpp`.

brnorris03:
### Problem description

`LowerMatmulCompute` could not lower compute bodies that store a scaled accumulator expression plus a matmul result, such as:

```python
out = scale * acc + (a @ b)
```

This blocked the flash-style recurrence:

```python
o_state = alpha * o_old + (exp_scores @ v)
```

The previous workaround staged the scaled accumulator through a dataflow buffer before adding it to the matmul result, incurring an extra DFB.

### What's changed

- Adds DST footprint metadata through a TTL DST access interface.
- Adds `ttl.dst_index` so a single tile can be selected from multi-tile DST producers.
- Updates `LowerMatmulCompute` to precompute SSA accumulator expressions into the matmul output DST slots, then use `matmul_block` to accumulate the matmul addend.
- Preserves `matmul_block` accumulation for dataflow-buffer-backed accumulators and multi-tile outputs.
- Preserves in-place DST reuse for fused post-ops so existing post-op matmuls do not consume an extra DST slot.
- Moves expanded DST capacity validation before rewrite mutation so errors are reported once and failed lowering leaves IR unchanged.
- Updates TTKernel conversion and TTL scheduling to use DST footprints for indexed reads, range writes, and DST hazards.
- Updates DST allocation/utilization docs for the new interface and lowering invariant.
- Adds MLIR coverage and a Python device test with a TTNN golden for the scaled recurrence and flash-like broadcast recurrence.

### Checklist

- [x] New/Existing tests provide coverage for changes.

Closes #665

### DIFF
--- a/docs/development/DST_Allocation.md
+++ b/docs/development/DST_Allocation.md
@@ -79,13 +79,15 @@ CBs, and accumulates the reduction result into a fresh DST slot:
   called after the last `reduce_tile` to clear this mask before any
   non-reduce operation.
 
-**(TTL) Matmul** (`ttl.tile_matmul_block`) reads A and B tiles from CBs and
-accumulates the matrix product into a fresh DST slot:
-- Both inputs stay in CBs (`TTLCBInputTileOpTrait`)
-- Output gets a fresh DST slot (not in-place)
+**(TTL) Matmul** (`ttl.tile_matmul_block`) reads A and B tiles from dataflow
+buffers and accumulates the matrix product into DST:
+- Both inputs stay in dataflow buffers (`TTLCBInputTileOpTrait`)
+- A tile-level matmul result occupies one DST slot before block expansion
+- A block matmul result occupies a contiguous `M*N` DST footprint after block
+  expansion, starting at the op's `dst_index`
 - `inputs_footprint = 0`
 - Accumulation: Multiple `matmul_tiles` calls to the same DST index
-  accumulate C += A * B across K-dimension tiles. The DST slot must
+  accumulate C += A * B across K-dimension tiles. The DST footprint must
   remain live across the entire K loop. `tile_regs_acquire` implicitly
   zeros DST before the loop; all K iterations accumulate before
   `tile_regs_commit`:
@@ -108,6 +110,10 @@ accumulates the matrix product into a fresh DST slot:
   results are spilled to an intermediate CB and reloaded with
   `copy_tile` in subsequent blocks (see
   `bmm_large_block_zm.cpp` in tt-metal for this pattern).
+- A scaled additive recurrence such as `scale * acc + (a @ b)` precomputes
+  `scale * acc` into the output DST footprint, then emits `matmul_block` over
+  the same footprint. The matmul op has no accumulator operand in this form;
+  hardware accumulation into prefilled DST preserves the final stored value.
 - Init: `mm_init_short` configures UNPACK + MATH (not PACK). Can
   be called multiple times per kernel to re-enter matmul mode after
   other operations.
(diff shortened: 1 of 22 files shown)


zoecarver (approved):
LGTM

phizalev-TT (approved):
LGTM!

PR #682, opened 2026-06-12
[ci] Build S3 light wheels on manylinux 2.34

brnorris03:
### Problem description
S3 `tt-lang-light` wheels need to be installable on glibc 2.34 systems for CPython 3.10 and CPython 3.12, with wheel tags compatible with the corresponding `ttnn` wheels. The publish flow also needs to preserve existing bundled-wheel behavior while adding a reproducible manylinux light-wheel build that can be exercised both locally and in CI. 

### What's changed
This PR adds a `manylinux_2_34` wheel-builder container flow that prebuilds the TT-Lang toolchain for each supported Python ABI, then uses those images to build `cp310` and `cp312` `tt-lang==...+light` wheels plus the `tt-lang-light` metapackage.

The S3 publish workflow now treats light-wheel publishing as additive: scheduled publishes continue producing the bundled wheel and also build the manylinux light wheels. The normal bundled/public wheel flow stays separate from the manylinux light-wheel flow, and the publish helpers verify exact wheel names and versions before upload.

The simulator remains importable on Python 3.10 by deferring `sys.monitoring` (requires >= 3.12) access until automatic copy-wait hook installation is actually required. Python 3.10 users can import `ttl.sim` and run paths that do not require those hooks; automatic copy-wait injection reports a Python 3.12+ requirement when used.

### Validation
- `cp310` and `cp312` local `tt-lang-light` wheel builds passed with the rebuilt manylinux images.
- Repaired wheels import successfully and pass `smoke-test-wheel.py`.
- Repaired wheel contents contain no vendored `libzstd`, no `_mlirRegisterEverything`, and no `tt_lang.libs` payload.
- A Python 3.10 manylinux container without `/opt/ttlang-toolchain` installed the local `tt-lang-light` wheel, used an external tt-metal install, and successfully ran `tutorials/elementwise/step_1_single_node_single_tile_block.py` on hardware.
- Example CI workflow (fails here because can only publish from `main`, but showing all non-publish steps passing): https://github.com/tenstorrent/tt-lang/actions/runs/27477670892

### Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,7 @@
 /third-party/llvm-project/build*/
 /third-party/tt-metal/build*/
 /third-party/tt-mlir/build*/
+/third-party/*/.cpmcache/
 /ird-toolchain/
 /dist*/
 /*.whl
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,8 +2,14 @@
 /build*/
 /.build/
 /*.build/
+/third-party/llvm-project/build*/
+/third-party/tt-metal/build*/
+/third-party/tt-mlir/build*/
+/ird-toolchain/
 /dist*/
 /*.whl
+/docker*.log
+/.docker-image-*
 
 # Python cache
 __pycache__/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,9 @@
 # Build directories
-build/
-.build/
-*.build/
+/build*/
+/.build/
+/*.build/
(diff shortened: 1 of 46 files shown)


zoecarver (approved):
Wow! Lots of change. Thank you for doing this! I skimmed but am less familiar with the build infra. LGTM!

phizalev-TT (approved):

PR #688, opened 2026-06-14
[ttl] DFB index reuse: module-level finalize pass (compiler half of #670)

zoecarver:
File-level squash of the non-atom half of the dfb-reuse work onto atom. Intended to rebase onto main independently of the ttl.atom changes.

- TTLFinalizeDFBIndices: module-level index reuse keyed on producer/consumer thread identity + element type with disjoint per-thread lifetimes; pipe-attached DFBs excluded; emits the ttl.dfb_index_map module attr (consumed by the atom-side runtime in the companion commit).
- TTLInsertIntermediateDFBs: single-block (blockCount 1) compiler intermediates.
- TTLValidateCBBudget: TTL_CB_BUDGET_DEBUG per-CB byte breakdown.
- TTL.h: kDFBIndexMapAttrName constant.
- lit tests (finalize_dfb_indices, insert_intermediate_dfbs*) + DFBManagement.md.

### DIFF
--- a/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
@@ -6,11 +6,26 @@
 // TTL Finalize DFB Indices
 //===----------------------------------------------------------------------===//
 //
-// Module-level pass that runs after all DFB-creating passes. Reuses
-// compiler-allocated DFB indices when lifetimes do not overlap, then
-// computes the true DFB count, updates ttl.base_cta_index on every
-// function, and collects compiler-allocated DFBs into the
-// ttl.compiler_allocated_dfbs module attribute for the Python runtime.
+// Module-level pass that runs after all DFB-creating passes. Reuses DFB
+// indices when lifetimes do not overlap, then computes the true DFB count,
+// updates ttl.base_cta_index on every function, and publishes the final
+// index assignment for the Python runtime (ttl.dfb_index_map for
+// user-declared DFBs, ttl.compiler_allocated_dfbs for compiler-allocated
+// ones).
+//
+// A logical DFB is one cb_index, bound by one bind_cb per kernel thread that
+// touches it. The same physical CB index has shared pages_received /
+// pages_acked counters and per-RISC read/write pointers, so two logical DFBs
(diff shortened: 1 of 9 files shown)


lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
```diff
@@ -40,254 +51,404 @@ namespace mlir::tt::ttl {
...
+        auto pb = b->intervals.find(b->producer.getOperation())->second;
+        return std::tie(pa.start, pa.end, a->origIndex) <
+               std::tie(pb.start, pb.end, b->origIndex);
+      });
+      size_t classBase = slots.size();
+      for (LogicalDFB *dfb : members) {
+        size_t placed = slots.size();
+        for (size_t s = classBase; s < slots.size(); ++s) {
+          if (llvm::none_of(slots[s], [&](LogicalDFB *member) {
+                return conflicts(*member, *dfb);
```
brnorris03, line 324: This only checks per-thread lifetime overlap, no? So not necessarily that the producer's push order matches the consumer's pop order. 

Repro: reader pushes `cb0,cb1`, compute waits `cb1,cb0` and merges `cb1,cb0`.

```suggestion
                return conflicts(*member, *dfb) ||
                       !fifoOrderAgrees(*member, *dfb);
```

Needs a simple `fifoOrderAgrees(a, b)` helper that returns `true` when producer == consumer, else `(producerA.start < producerB.start) == (consumerA.start < consumerB.start)`.

lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
```diff
@@ -40,254 +51,404 @@ namespace mlir::tt::ttl {
...
+          for (Value v : worklist) {
+            for (Operation *user : v.getUsers()) {
+              if (isPipeOp(user)) {
+                dfb.eligible = false;
+                return;
+              }
+              for (Value result : user->getResults()) {
+                next.push_back(result);
+              }
+            }
```
brnorris03, line 216: Can you explain a bit more the reasoning behind this? Why 3? Is it correct for arbitrary (valid) chains of ops?

PR #691, opened 2026-06-16
Same type typecast is no-op.

vtsilytskyiTT:
### Problem description
Typecast for the same type is treated as an error.

### What's changed
Now it is no-op.

### Ticket
None

### Checklist


zoecarver (approved):
LGTM! Thank you!!

PR #697, opened 2026-06-18
fix(test): drop undefined `sc` arg in layernorm minimal-DFB reduce_sum

ezietlowTT:
### Problem

`make_layernorm_kernel_minimal_dfbs` in `test/python/test_layernorm.py` computes the
variance reduction as:

```python
var_blk += ttl.math.reduce_sum(diff * diff, sc, dims=[1])
```

but `sc` is **never defined** anywhere in the module — it's a stray, undefined name. The
factory is exercised (`_run_layernorm(make_layernorm_kernel_minimal_dfbs, ...)`), so tracing
this kernel raises `NameError: name 'sc' is not defined`.

### Fix

Drop the stray `sc`, matching every other `reduce_sum` in the file (e.g. the mean pass uses
the bare `reduce_sum(xj, dims=[1])`), and consistent with how `mean_scale` (`ms`) carries the
1/N scaling separately rather than via a reduce operand:

```python
var_blk += ttl.math.reduce_sum(diff * diff, dims=[1])
```

One-token change; only the minimal-DFB layernorm path is affected (the other two factories
already use the bare form).

### Validation

- Now identical to the two sibling layernorm factories that already pass.
- Local: `python -m pytest test/sim` green (95/95); file is `black`-clean.
- **Draft** pending the on-device `test/python/test_layernorm.py` PCC run via CI (`assert_pcc`),
  which is the gate I can't reproduce locally without the full device build.


brnorris03 (approved):
Thank you!

PR #699, opened 2026-06-22
[ttl] Expose exp kernel arguments

vtsilytskyiTT:
### Problem description
Exp kernel supports useful flags, like scale, which are not available in tt-lang.

### What's changed
Use custom Exp implementation instead of UnaryElementwise.
Add user-facing `scale` attribute and SFPU-facing clamping, approximation.

### Depends on

1. https://github.com/tenstorrent/tt-mlir/pull/8796 (implementation)
2. https://github.com/tenstorrent/tt-mlir/pull/8830 (fixup)
3. https://github.com/tenstorrent/tt-mlir/pull/8883 (runtime arguments)

### Checklist

### DIFF
--- a/lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
+++ b/lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
@@ -162,6 +162,18 @@ static llvm::DenseMap<mlir::TypeID, InitOpInfo> buildComputeToInitMap() {
                                         typecastOp.getOutDtypeAttr());
       }};
 
+  // ExpTile: exp_tile_init configures the SFPU per exp flags. It takes approx,
+  // scale (the fp32 scale factor template), and input_clamping read off the
+  // exp_tile op. scale_en / iterations are compute-only and not part of the
+  // init. (exp is excluded from the generic unary macro above for this reason.)
+  map[mlir::TypeID::get<ttk::ExpTileOp>()] = {
+      [](OpBuilder &b, Location l, Operation *computeOp) {
+        auto expOp = cast<ttk::ExpTileOp>(computeOp);
+        ttk::ExpTileInitOp::create(b, l, expOp.getApproxAttr(),
+                                   expOp.getScaleAttr(),
+                                   expOp.getInputClampingAttr());
+      }};
+
   // Transpose: resolves output CB from annotated attribute.
   map[mlir::TypeID::get<ttk::TransposeTileOp>()] = {
       [](OpBuilder &b, Location l, Operation *computeOp) {
@@ -228,6 +240,27 @@ static InitKey computeInitKey(Operation *op) {
   if (auto typecast = dyn_cast<ttk::TypecastTileOp>(op)) {
     int64_t disc = (static_cast<int64_t>(typecast.getInDtype()) << 16) |
                    static_cast<int64_t>(typecast.getOutDtype());
+    return {typeId, {}, disc};
+  }
+
+  // For exp: distinct flag combinations configure exp_tile_init differently
+  // and must not share an init. The init depends on approx, input_clamping,
+  // and the fp32 scale template, so encode all three in the discriminator.
+  // scale_en / iterations are compute-only and do not affect the init.
+  if (auto exp = dyn_cast<ttk::ExpTileOp>(op)) {
+    uint32_t scaleBits = 0x3F800000u; // default 1.0f for exp_tile_init.
+    if (auto scaleAttr = exp.getScaleAttr()) {
+      scaleBits = static_cast<uint32_t>(scaleAttr.getInt());
+    }
+    BoolAttr approxAttr = exp.getApproxAttr();
+    bool approx = approxAttr && approxAttr.getValue();
+    int64_t inputClamping =
+        static_cast<int64_t>(ttk::InputClamping::ClampToNegative);
+    if (auto inputClampingAttr = exp.getInputClampingAttr()) {
+      inputClamping = static_cast<int64_t>(inputClampingAttr.getValue());
+    }
+    int64_t disc = (static_cast<int64_t>(scaleBits) << 8) |
+                   (static_cast<int64_t>(approx) << 1) | inputClamping;
     return {typeId, {}, disc};
   }
 
--- a/lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
+++ b/lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
@@ -42,6 +42,8 @@ namespace ttk = mlir::tt::ttkernel;
 
 namespace {
 
+constexpr int64_t defaultExpIterations = 8;
+
 /// Materializes a float attribute as an i32 carrying its IEEE 754 bits. Scalar
 /// SFPU tile APIs take i32 params even for float scalars.
 static Value floatAttrToI32Bits(OpBuilder &rewriter, Location loc,
@@ -49,6 +51,14 @@ static Value floatAttrToI32Bits(OpBuilder &rewriter, Location loc,
   auto f32Val = arith::ConstantOp::create(
       rewriter, loc, rewriter.getF32FloatAttr(attr.getValueAsDouble()));
   return arith::BitcastOp::create(rewriter, loc, rewriter.getI32Type(), f32Val);
+}
+
+/// Materializes a float attribute as an i32 attribute carrying its IEEE 754
+/// bits.
+static IntegerAttr floatAttrToI32BitsAttr(OpBuilder &builder, FloatAttr attr) {
+  uint32_t bits =
+      static_cast<uint32_t>(attr.getValue().bitcastToAPInt().getZExtValue());
+  return builder.getI32IntegerAttr(bits);
 }
 
 /// Look up a CB for a copy_tile source.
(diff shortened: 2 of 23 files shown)


test/python/test_exp_flags.py
```diff
@@ -0,0 +1,155 @@
...
+@ttl.operation(grid=(1, 1))
+def exp_scale_kernel(in_t, out_t):
+    a_cb = ttl.make_dataflow_buffer_like(in_t, shape=(1, 1), block_count=2)
+    out_cb = ttl.make_dataflow_buffer_like(out_t, shape=(1, 1), block_count=2)
+
+    @ttl.compute()
+    def compute_fn():
+        x = a_cb.wait()
+        r = out_cb.reserve()
+        r.store(ttl.exp(x, scale=2.0))
```
phizalev-TT: Can this be `r.store(ttl.exp(x * 2.0))`?

brnorris03: +1 -- I think that would be more consistent with previous scaling implementations.

vtsilytskyiTT: Do you mean rewrite multiplication as the scale parameter? This is a great user experience improvement, let me try to do that.

phizalev-TT:
Please document new arguments in the spec: https://github.com/tenstorrent/tt-lang/blob/main/docs/sphinx/specs/TTLangSpecification.md#basic-unary-math-functions

python/ttl/operators.py
```diff
@@ -997,6 +1068,7 @@ def raw_element_write(block, *args):
     "signpost",
     "fill",
     "typecast",
+    "exp",
```
phizalev-TT, line 1172: `exp` should be under `ttl.math`, e.g. `ttl.math.exp()`.

zoecarver (approved):

lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
```diff
@@ -237,6 +257,28 @@ static InitKey computeInitKey(Operation *op) {
...
+      scaleBits = static_cast<uint32_t>(scaleAttr.getInt());
+    }
+    BoolAttr approxAttr = exp->getAttrOfType<BoolAttr>("approx");
+    bool approx = approxAttr && approxAttr.getValue();
+    int64_t inputClamping =
+        static_cast<int64_t>(InputClamping::ClampToNegative);
+    if (auto inputClampingAttr =
+            exp->getAttrOfType<IntegerAttr>("input_clamping")) {
+      inputClamping = inputClampingAttr.getInt();
+    }
```
brnorris03: A few more instances of the hardcoded strings that are candidates for named const (and saw more below).

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
```diff
@@ -1051,6 +1100,8 @@ void populateTTLTileOpsToTTKernelPatterns(TypeConverter *typeConverter,
   patterns.add<TTLTileFillToTTKernel>(ctx);
   patterns.add<TTLTileMulUnaryConstToTTKernel>(ctx);
   patterns.add<TTLTileTypecastToTTKernel>(ctx);
+  // exp carries hardware flags; lowered by a dedicated pattern.
+  patterns.add<TTLTileExpToTTKernel>(ctx);
```
brnorris03: nit: superfluous comment, I'd remove

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
```diff
@@ -291,6 +299,47 @@ struct TTLTileTypecastToTTKernel : OpConversionPattern<TileTypecastOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value dstIdxVal = adaptor.getDstIndex();
+
+    auto expOp = ttk::ExpTileOp::create(rewriter, loc, dstIdxVal);
+    if (op.getApprox()) {
+      expOp->setAttr("approx", rewriter.getBoolAttr(true));
+    }
+    if (op.getInputClamping() != ttl::InputClamping::ClampToNegative) {
+      expOp->setAttr("input_clamping",
```
brnorris03: Consider converting the existing `TTL_InputClampingAttr` to `ttk::InputClampingAttr` here instead of re-emitting it as `i32`. The TTL op already models this as an enum attr, and TTKernel/EmitC reads `getInputClampingAttr()`.

python/ttl/ttl_math.py
```diff
@@ -11,7 +11,86 @@
...
+    scale: float | None = None,
+    skip_clamp_check: bool = False,
+    iterations: int = 8,
+):
+    """Element-wise exponential.
+
+    With default arguments this matches the plain hardware ``exp_tile`` (no
+    approximation, no scaling, clamped). Keyword flags expose the SFPU exp
+    template parameters.
+    """
```
brnorris03, line 80: Please add parameter docs to docstring (e.g., what is iterations)?

python/ttl/ttl_math.py
```diff
@@ -20,5 +99,6 @@
     "transpose",
     "fill",
     "typecast",
+    "exp",
     *_generated_all,
```
brnorris03, line 114: nit: would be nice to alphabetize these (leaving `*_generated_all` at the end); not insisting though

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -623,6 +640,17 @@ static LogicalResult buildFusedCompute(Operation *sinkOp,
...
   // by buildFusedCompute above.
   DenseMap<Value, std::pair<Value, Value>> deferredMatmul;
+  DenseSet<Value> expScaleMulResults;
+  for (Operation *op : trace.opsInOrder) {
+    auto expOp = dyn_cast<ExpOp>(op);
+    if (!expOp || expOp.getScaleAttr()) {
+      continue;
+    }
+    auto mulOp = expOp.getInput().getDefiningOp<MulUnaryConstOp>();
+    if (mulOp && mulOp->hasOneUse() && trace.opsInOrder.contains(mulOp)) {
```
brnorris03: Consider rejecting this fold when a compute-side-effect op sits between the `mul_unary_const` and its `exp` user. Before the fold, a `ttl.dprint {mode = "dst"}` after the multiply observes the scaled tile. After the fold, the multiply only exists as the `tile_exp` scale attr, so the dprint is emitted before the scale is applied.

```mlir
// Before fold: dprint observes %scaled.
%scaled = ttl.mul_unary_const %a, 2.0 : tensor<...> -> tensor<...>
ttl.dprint "after scale" () {mode = "dst"} : ()
%exp = ttl.exp %scaled : tensor<...> -> tensor<...>

// After fold: dprint runs before tile_exp applies scale.
ttl.dprint "after scale" () {mode = "dst"} : ()
%exp = ttl.tile_exp %a {scale = 2.0 : f32} : !ttcore.tile<...> -> !ttcore.tile<...>
```

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
```diff
@@ -291,6 +299,47 @@ struct TTLTileTypecastToTTKernel : OpConversionPattern<TileTypecastOp> {
...
+                     rewriter.getI32IntegerAttr(
+                         static_cast<int32_t>(op.getInputClamping())));
+    }
+    if (op.getIterations() != 8) {
+      expOp->setAttr("iterations",
+                     rewriter.getI32IntegerAttr(op.getIterations()));
+    }
```
brnorris03: What's the significance of 8? Would perhaps be better as a named constant?

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -634,6 +662,10 @@ static LogicalResult buildFusedCompute(Operation *sinkOp,
     }
 
     Value tileResult;
+    if (op->getNumResults() == 1 &&
+        expScaleMulResults.contains(op->getResult(0))) {
```
brnorris03: Can you add a short comment here? It's not obvious why the skip is done. Also, this `continue` is what suppresses the standalone `tile_mul_unary_const` after `expScaleMulResults` marks it for folding into the later `tile_exp` scale attr; without that context it is hard to tell why the multiply is skipped.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -266,6 +266,23 @@ static Value emitTileOpFor(OpBuilder &builder, Location loc,
...
 
+  // ExpOp: unary, but forwards hardware exp flags to the tile op. Handled
+  // explicitly because it is excluded from TTLElementwiseOps.def. Flags are
+  // read via the value getters (which resolve defaults) so that exp ops
+  // written without flags forward well-formed attributes; the optional scale
+  // is forwarded as-is (null when unset).
+  if (auto expOp = dyn_cast<ExpOp>(sourceOp)) {
+    assert(tileOperands.size() == 1 && "exp fusion requires one tile operand");
+    Value dstIndex = createPlaceholderDstIndex(builder, loc);
+    auto tileOp = ExpTileOp::create(
```
brnorris03: Consider factoring exp tile creation into a helper used here and by the scaled-exp fold. `exp` probably needs a dedicated lowering because it carries flags, but the attr construction is now duplicated; a helper with an optional scale override would keep `approx` / `input_clamping` / `iterations` handling in one place.

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
vtsilytskyiTT: Agree.

python/ttl/ttl_math.py
vtsilytskyiTT: Done, moved 3 ops.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Added `hasInterveningComputeSideEffect` and a test.

lib/Dialect/TTL/Transforms/ConvertTTLTileOpsToTTKernel.cpp
vtsilytskyiTT: Good point, now using `ttk::InputClampingAttr inputClampingAttr;`

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
vtsilytskyiTT: Refactored, please take a look.

lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
vtsilytskyiTT: I'm using more enums instead of strings.

zoecarver (approved):

python/ttl/ttl_math.py
```diff
@@ -11,14 +11,105 @@
 # Re-export all generated elementwise operations
 from ._generated_elementwise import *  # noqa: F401,F403
 from ._generated_elementwise import __all__ as _generated_all
-from .operators import broadcast, reduce_sum, reduce_max, transpose, fill, typecast
+from ._src.ttl_ast import syntax
+from .operators import (
+    broadcast,
```
FConstantinos, line 16: I see all these operations being imported from operators.py. So why not do the same with exp?

Also I see exp is defined both under operators.py and here, is there a potential for duplication?

FConstantinos (approved):
LGTM from a sim perspective, but also a minor comment

PR #700, opened 2026-06-23
[ttl][pipes] Add capacity-controlled PipeNet protocol [pipes-4]

brnorris03:
PipeNet PR tree:

```text
main (includes #756, #768, #704, #782, #778, #775, #783, and #765; all merged)
  -> [pipes-4] #700 (this PR; includes [pipes-5] #740 [x] and [pipes-6] #784 [x])
    |-> [pipes-8] #754
    |     `-> [pipes-7] #780
    `-> [pipes-fabric] #734
```

### Review history

#740 extracted the `PipeTransportEmitter` refactor from this work and was independently reviewed and approved. During the final stack reorder, #740 still targeted #700 when #700 was updated to include the approved #740 commits. GitHub therefore recorded #740 as merged into #700. This PR includes those reviewed changes unchanged.

#784 moved PipeNet lowering decisions into immutable plans that are validated before IR mutation. It was also incorporated into #700 during the reorder, and GitHub recorded it as merged into this PR.

### Problem description

Receiver Post synchronization prevents a receiver from posting its next reservation until its current DFB block is released. Computed Address / Capacity Counter (`CA/CC`) permits the sender to use any free receiver slot without waiting for a current post, enabling overlap between sender transfers and receiver work.

Capacity synchronization is correct only when the compiler proves exclusive pipe ownership, one writer, whole-slot reservation and release, and matching transfer counts.

### What's changed

- Prove single-writer, pipe-only receiver DFB ownership and matching whole-slot reserve, wait, push, and pop operations.
- Select `CA/CC` only for proven intra-device point-to-point transfers; other transfers retain Receiver Post synchronization.
- Wait on cumulative capacity before each send and return one unit immediately after the matching receiver pop.
- Allocate capacity counters through the shared local/GlobalSemaphore allocator.
- Add `--ttl-pipe-capacity-sync` and document both option states.

Capacity information is only consulted while lowering the existing pop. Emitting the release there keeps it adjacent to the pop and in the same loop or branch; it does not combine DFB-pop semantics with capacity analysis or require another IR traversal.

### Performance

The pipelined benchmark retains each received block while writing it to four output locations:

```python
with recv_dfb.wait() as recv_block:
    for output_index in range(receiver_work):
        ttl.copy(recv_block, out[output_index, transfer_index]).wait()
# Exiting the context pops the block and returns CA/CC capacity.
```

The Blackhole measurements report the complete interval from the earliest sender or receiver kernel start to the latest completion. Each result is the median of 50 warm dispatches after 5 warmups.

| Transfers | `CA/CC` (us) | `CA/RP` (us) | `RA/RP` (us) | `CA/CC` reduction vs `CA/RP` | `CA/CC` reduction vs `RA/RP` |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 64 | 89.252 | 101.847 | 114.044 | 12.37% | 21.74% |
| 256 | 354.841 | 406.895 | 456.470 | 12.79% | 22.26% |
| 1024 | 1416.960 | 1623.742 | 1823.716 | 12.73% | 22.30% |
| 4096 | 5661.811 | 6495.647 | 7305.041 | 12.84% | 22.49% |

The advantage begins with double-buffering: at 256 transfers the reduction versus `CA/RP` was 1.74% with receiver `block_count = 1` and 12.64% with `block_count = 2`.

Benchmark: https://gist.github.com/brnorris03/b9c79d994dcf9a6524a5a63a9ea2b976

### Validation

New regressions compare BF16 and FP32 `CA/CC`, `CA/RP`, and `RA/RP` transfers and cover capacity proofs, both capacity-sync option states, local and global counters, release placement, and unsupported DFB ownership.

### Checklist


include/ttlang/Dialect/TTL/IR/TTL.h
```diff
@@ -95,6 +95,10 @@ constexpr llvm::StringLiteral
 /// the attribute value is a `ttkernel.thread` enum.
 constexpr llvm::StringLiteral kKernelThreadAttrName("ttl.kernel_thread");
 
+/// TTKernel-side thread tag, set during TTL -> TTKernel conversion and read by
```
arichinsTT: nit: should this be defined in TTKernel.h?

arichinsTT: also in TTLOpsUtils.h kTTKernelThreadAttrName andf kKernelThreadAttrName seem pretty similar, whats the difference?

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -36,6 +36,33 @@ inline mlir::func::FuncOp getEnclosingKernelThread(mlir::Operation *op) {
   return nullptr;
 }
 
+/// Return the TTKernel thread type attached to `func`.
```
arichinsTT, line 40: nit: also TTKernel util not ttl

include/ttlang/Dialect/TTL/Transforms/LaunchNodeDomainAnalysis.h
```diff
@@ -91,6 +91,24 @@ LaunchNodeDomain getPipeSourceLaunchNodeDomain(PipeType pipeType);
...
+
+/// Return true if two launch-node domains may share at least one node.
+///
+/// Unknown domains are treated as overlapping because callers cannot prove
+/// disjointness from incomplete facts.
+bool launchNodeDomainsOverlap(const LaunchNodeDomain &lhs,
+                              const LaunchNodeDomain &rhs);
+
+/// Return true if `domain` contains `coord`, or if the domain is unknown.
+bool launchNodeDomainContains(const LaunchNodeDomain &domain,
```
arichinsTT: for better distinction, maybe launchNodeDomainMayContain? or could the only usage of this function in Pipegraph.cpp be changed to use knownLaunchNodeDomainContains and it's own null check?

arichinsTT (commented):
Could there be a comment highlighting error handling for excluding multicast support currently? is this something that we would eventually want to have limited support for?

include/ttlang/Dialect/TTL/IR/TTL.h
brnorris03: kTTKernelThreadAttrName not used anywhere anymore (now using the actual ttk attribute directly), so removed.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
brnorris03: I think it's fine here since it accepts both ttl and ttkernel.

include/ttlang/Dialect/TTL/Transforms/LaunchNodeDomainAnalysis.h
brnorris03: Removed `launchNodeDomainContains`; its only caller now checks `postDomain.known` before `knownLaunchNodeDomainContains`.

brnorris03:
Multicast is supported with receiver-authored address/receiver-post synchronization (RA/RP) or computed address/receiver-post synchronization (CA/RP); only capacity-counter synchronization (CC) is unsupported. Added an explicit check and debug message; #728 tracks per-receiver CC support.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -661,35 +908,90 @@ LogicalResult lowerPipeTransferPost(PipeTransferPostOp op, Value dst,
...
+  if (auto attr = op.getNumTilesAttr()) {
+    return arith::ConstantIntOp::create(rewriter, loc, attr.getInt(), 32);
+  }
+  auto ttlCbTy = getTTLCBType(originalCb);
+  assert(ttlCbTy && "lowerCBPop already verified the DFB type");
+  return arith::ConstantIntOp::create(rewriter, loc,
+                                      ttlCbTy.getElementsPerBlock(), 32);
+}
+
+LogicalResult lowerCBPop(CBPopOp op, Value cb,
```
arichinsTT, line 1026: combining the lowering of cb pop with the PipeCapacity is kinda confusing, do they need to happen at the same time, or can there be a walk after

python/ttl/kernel_runner.py
```diff
@@ -167,6 +173,7 @@ def build_kernel_descriptors(
 
     # CB indices are 0, 1, 2, ... for each CB (including intermediate CBs).
     cb_indices = list(range(num_cbs))
+    computed_address_base_addresses = pipe_computed_address_base_addresses or {}
```
zoecarver, line 249: Does this need to be part of the hash?

python/ttl/kernel_runner.py
```diff
@@ -466,9 +576,16 @@ def build_cb_descriptors(
...
+    pipe_computed_address_dfb_tensors: Optional[Dict[int, Any]] = None,
 ) -> List[Any]:
     """Return io_tensors for ttnn.generic_op, including pipe SRAM scratch."""
-    io_tensors = list(tensors) + list(pipe_sram_scratch_tensors)
+    computed_address_dfb_tensors = [
+        pipe_computed_address_dfb_tensors[dfb_index]
+        for dfb_index in sorted(pipe_computed_address_dfb_tensors or {})
+    ]
+    io_tensors = (
+        list(tensors) + list(pipe_sram_scratch_tensors) + computed_address_dfb_tensors
```
zoecarver: I don't think there's anything to change here, but maybe a comment here or in the doc if you want. Codex noted: generic_op returns the last tensor so if we ever returned the result of the generic op or used it, it would be the computed address dfb tensor.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: A separate compiler walk is possible, but it would need to find the corresponding lowered pop. Lowering them together keeps the release immediately after the pop, including inside loops and branches.

python/ttl/kernel_runner.py
brnorris03: No. The DFB base is allocation-dependent runtime data, so passed as a runtime arg (so no recompilation needed). Thanks for asking! This is what prompted me to make it a runtime arg

python/ttl/kernel_runner.py
brnorris03: Good catch. `generic_op` returns the last tensor, and `CompiledTTNNKernel.__call__` forwards it to the caller. Compiler-owned tensors now precede user tensors, so the user output remains last. Added unit coverage.

lib/Dialect/TTL/Transforms/PipeLowering.cpp
arichinsTT: I find it pretty confusing, but I understand its a bit more involved of change, can it be a TODO for later? I'd be fine helping with this change later

brnorris03: I may be misunderstanding what you mean by "combining." The capacity plan is already complete here; `lowerCBPop` only checks whether this pop releases sender capacity. If it does, it emits a remote semaphore increment after `cb_pop_front`, telling the sender that one receiver DFB block is reusable.

By "a walk after" do you mean a second traversal after DFB-pop lowering that finds these pops and inserts the increments separately? Why would that be preferable?

jackzhangTT (approved):
Looks good thanks for the detailed description!

zoecarver (approved):

Issue #683, opened 2026-06-13
[external bug report][from claude] convert-ttl-to-compute asserts (isBeforeInBlock) when one compute value is stored across scf.if branches


brnorris03:
## Summary

`convert-ttl-to-compute` aborts with an MLIR assertion when a single compute
value is stored into a DFB from more than one block of a control-flow region
(for example both branches of a runtime `if/else`):

```
mlir/lib/IR/Operation.cpp:382: bool mlir::Operation::isBeforeInBlock(Operation *):
Assertion `other && other->block == block &&
"Expected other operation to have the same parent block."' failed.
```

This occurs in `@ttl.compute()` and is a hard crash (SIGABRT, core dumped), not
a diagnostic.

## Root cause

`ConvertTTLToCompute` fuses a value's stores into a single `ttl.compute` body,
which assumes every store of that value is a sibling in one block. `findLastStore`
(`lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp`) walks all `ttl.store`
users of the value and orders them with `last->isBeforeInBlock(s)`:

```cpp
for (OpOperand &use : op->getResult(0).getUses()) {
  if (auto s = dyn_cast<StoreOp>(use.getOwner())) {
    if (!last || last->isBeforeInBlock(s))   // asserts when last and s are in different blocks
      last = s;
  }
}
```

When the value is stored in two sibling blocks (the `then` and `else` regions of
an `scf.if`), `last` and `s` have different parent blocks and `isBeforeInBlock`
asserts. The same single-block assumption is in `emitTileStores`, which absorbs
all of the value's stores into one compute body.

## Reproducer (MLIR, no device)

```
ttlang-opt repro.mlir -pass-pipeline='builtin.module(func.func(convert-ttl-to-compute))'
```

```mlir
// repro.mlir -- one value (%v) stored into a different DFB in each scf.if branch.
func.func @store_across_branches(%cond: i1)
    attributes {ttl.kernel_thread = #ttkernel.thread<compute>} {
  %in_cb = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
  %out_a = ttl.bind_cb {cb_index = 1, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
[...]

PR #701, opened 2026-06-23
[ttl] CF yield scalar and handle reassignments with type propagation

### LINKED ISSUE #380: [external bug report][from claude] Variable assigned in if/else branches lost during tracing (KeyError in datamovement)

arichinsTT:
### Problem description

Two independent issues are addressed in this PR:

1. **Conditional reassignment (#380):** Variables assigned inside `if/else`
   branches in `@ttl.datamovement()` were lost after the block exits. The AST
   tracer for this landed in #654; this PR adds transfer-handle verifier
   support and Python-level tests for conditional reassignment of scalars and
   transfer handles.

2. **LowerScalarFpTypes rewrite:** Lowers scalar floating-point types and ops
   into their bitwise integer equivalents, rebuilt on MLIR's dialect
   conversion framework.

### What's changed

**Transfer handle verifier (`TTLOpsUtils.h`):** Extended
`traceTransferHandleSource` to trace through `scf.IfOp` results into both
branch yields. Both branches must independently trace to a valid `ttl.copy`
or `ttl.pipe_transfer.send`.

**AST control-flow branch scoping (`kernel_ast.py`):** Branch dependency
analysis now scopes each `if`/`else` branch independently and joins the
results, so loop-carried values assigned in either branch are tracked
correctly.

**Rewrite LowerScalarFpTypes (`LowerScalarFpTypes.cpp`):** Replaced the
hand-rolled two-phase pass with MLIR's dialect conversion framework:

- A `TypeConverter` maps scalar `FloatType` to the corresponding
  `IntegerType` (f32 → i32, bf16 → i16).
- Three `OpConversionPattern`s handle the float operations:
  - `CmpFToSoftFloat` — lowers `arith.cmpf` (ogt/olt) to
    `ttkernel.float32_greater` / `ttkernel.bfloat16_greater`.
  - `TruncFToBitExtract` — lowers `arith.truncf` (e.g. f32 → bf16) to
    `arith.shrui` + `arith.trunci` (bit extraction).
  - `ConstantOpConversion` — converts float constants to integer bit-pattern
    constants.
- `populateSCFStructuralTypeConversionsAndLegality` propagates integer types
  through `scf.if`, `scf.for`, `scf.while`, and `scf.yield`.
- `populateFunctionOpInterfaceTypeConversionPattern` converts `func.func`
  signatures and `func.return` operands.

### TODO / Open items (from review)

- [ ] **`truncf` numeric correctness.** The bit-extraction path for f32 → bf16
  utilizing metal numerical api
 -[ ] ** #739 getting types of variables initialized in if/else statments

### Ticket

closes #380

### Checklist

- [x] New/existing tests provide coverage for changes


include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -91,6 +91,34 @@ traceTransferHandleSource(mlir::Value value, MatchFn match,
         }
       }
     }
+    // Trace through scf.if results -- both branches must independently
+    // yield a valid source.  Each branch is a separate data-flow path,
+    // so give each its own visited set (avoids false negatives when
+    // both branches yield the same outer value).
+    if (auto ifOp = mlir::dyn_cast<mlir::scf::IfOp>(result.getOwner())) {
+      unsigned idx = result.getResultNumber();
```
brnorris03: Suggest replacing "valid source" with the concrete condition this helper checks. For this helper, that likely means each branch yield must trace through `match` to the expected transfer source type, such as the `CopyOp` or `PipeTransferSendOp` expected by the caller. That makes the failure condition clear without requiring readers to inspect `MatchFn`.

include/ttlang/Dialect/TTL/Passes.td
```diff
@@ -733,39 +733,38 @@ def TTLDumpCBFlowGraph
   ];
 }
 
-def TTLLowerScalarCmpF
-    : Pass<"ttl-lower-scalar-cmpf", "::mlir::ModuleOp"> {
-  let summary = "Lower arith.cmpf on scalar floats to TTKernel soft-float ops";
+def TTLLowerScalarFpTypes
+    : Pass<"ttl-lower-scalar-fp-types", "::mlir::ModuleOp"> {
+  let summary = "Lower scalar float types to integer bit patterns";
```
brnorris03: Suggest making this a `func.func` pass unless module scope is needed for call/signature conversion. The implementation rewrites function signatures and bodies; if there are no cross-function call or symbol updates, `Pass<..., "::mlir::func::FuncOp">` would better match the transformation scope. Keeping it as a module pass implies an interprocedural invariant that is not documented here, so if module scope is required, please document that invariant in the pass description.

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+      break;
+    case arith::CmpFPredicate::OLT:
+      if (bitWidth == 32) {
+        result = ttk::Float32GreaterOp::create(
+            rewriter, loc, rewriter.getI1Type(), rhsInt, lhsInt);
+      } else {
+        result = ttk::Bfloat16GreaterOp::create(
+            rewriter, loc, rewriter.getI1Type(), rhsInt, lhsInt);
+      }
+      break;
```
brnorris03, line 62: Suggest adding a small local helper for the greater op creation, e.g. `createGreaterOp(rewriter, loc, bitWidth, lhs, rhs)`. This avoids duplicating the f32/bf16 selection in four branches; otherwise adding another predicate/type or changing the result type requires editing each branch and risks inconsistent operands or op selection.

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+struct ConstantOpConversion : OpConversionPattern<arith::ConstantOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto floatAttr = mlir::dyn_cast<FloatAttr>(op.getValue());
+    if (!floatAttr) {
+      return rewriter.notifyMatchFailure(op, "not a float constant");
+    }
```
brnorris03, line 126: Consider changing this to plain `return failure();`, or avoiding the branch by matching only float constants. Non-float constants are already dynamically legal, so this is not an unsupported user input; it is just a pattern miss.

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+struct TTLLowerScalarFpTypesPass
+    : impl::TTLLowerScalarFpTypesBase<TTLLowerScalarFpTypesPass> {
+  using TTLLowerScalarFpTypesBase::TTLLowerScalarFpTypesBase;
+
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
+    MLIRContext &ctx = getContext();
+
+    TypeConverter typeConverter;
+    typeConverter.addConversion([](Type t) { return t; });
```
brnorris03, line 154: Suggest adding a short comment explaining that the identity conversion marks unchanged types legal for dialect conversion. This mirrors upstream converters such as `QuantizedTypeConverter`; without it, `typeConverter.isLegal(op)` would reject ordinary non-float operand/result types.

python/pykernel/_src/kernel_ast.py
```diff
@@ -150,6 +152,25 @@ def visit_AugAssign(self, node):
...
 
+    def visit_If(self, node):
+        condition_reads = _collect_read_variable_names(node.test)
+        condition_reads = self._expand_dependencies(condition_reads)
+        # Shallow copy: values are the same set objects. The identity check
+        # below works because _record_assignment always creates a *new* set
+        # via set(dependencies), so any entry that was written during the
+        # body visit will be a different object than the snapshot.
+        deps_before = dict(self._dependencies_by_name)
+        for stmt in node.body:
```
brnorris03: Suggest analyzing each branch with a copied dependency map and merging after both branches. The current code visits `body` then `orelse` against the same map, which models mutually exclusive branches as sequential. That can miss loop-carried values: if the then branch makes `tmp` depend on `acc` and the else branch assigns `tmp = other`, the else assignment overwrites the then dependency, so `acc = tmp` after the if is not marked loop-carried even though one branch uses `acc`.

A more robust structure would be:

```python
then_deps = analyze_branch(deps_before, node.body)
else_deps = analyze_branch(deps_before, node.orelse)
self._dependencies_by_name = join_deps(deps_before, then_deps, else_deps)
```

The join should union dependencies for names assigned in either branch because the post-if value may come from either branch. It should track written names explicitly because set identity is an implementation detail of `_record_assignment`; mutating or reusing sets later would silently change the analysis.

lib/Dialect/TTL/Transforms/LowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+namespace ttk = mlir::tt::ttkernel;
+
+struct CmpFToSoftFloat : OpConversionPattern<arith::CmpFOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(arith::CmpFOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Type origFloatTy = op.getLhs().getType();
+    unsigned bitWidth = origFloatTy.getIntOrFloatBitWidth();
```
brnorris03: Suggest guarding the scalar assumption before reading the bit width:

```c++
auto floatTy = dyn_cast<FloatType>(op.getLhs().getType());
if (!floatTy)
  return rewriter.notifyMatchFailure(op, "expected scalar cmpf");
```

Upstream `arith.cmpf` accepts `FloatLike`, including vectors/tensors, and `Type::getIntOrFloatBitWidth()` asserts on those types.

lib/Dialect/TTL/Transforms/LowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+              "and olt are currently supported");
+    }
+
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+
+/// Convert arith.truncf (e.g. f32 -> bf16) into integer bit extraction.
+/// bf16 is the upper 16 bits of the f32 IEEE-754 encoding, so a truncf
```
brnorris03: Suggest reusing upstream BF16 truncation semantics or restricting this to an explicitly bit-level operation. Upstream `BFloat16TruncFOpConverter` in `mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp` handles f32 -> bf16 with default rounding by adding rounding-bias and NaN handling. Without that, this rewrites `arith.truncf` to a different numeric operation for values that require rounding or special NaN handling. At minimum, check `src.isF32()`, `dst.isBF16()`, and `!op.getRoundingmodeAttr()` before rewriting.

lib/Dialect/TTL/Transforms/LowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+  LogicalResult
+  matchAndRewrite(arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto floatAttr = mlir::dyn_cast<FloatAttr>(op.getValue());
+    if (!floatAttr) {
+      return rewriter.notifyMatchFailure(op, "not a float constant");
+    }
+
+    for (Operation *user : op.getResult().getUsers()) {
+      if (isa<ttk::TTKernelDialect>(user->getDialect())) {
```
brnorris03: Suggest leaving a float constant legal only when all users are TTKernel ops. For example:

```c++
if (llvm::all_of(op.getResult().getUsers(), [](Operation *user) {
      return isa<ttk::TTKernelDialect>(user->getDialect());
    }))
  return true;
```

Otherwise a mixed-use constant keeps the scalar-fp user alive through an `unrealized_conversion_cast` instead of replacing it with an integer bit-pattern constant.

lib/Dialect/TTL/Transforms/LowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+    : impl::TTLLowerScalarFpTypesBase<TTLLowerScalarFpTypesPass> {
+  using TTLLowerScalarFpTypesBase::TTLLowerScalarFpTypesBase;
+
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
+    MLIRContext &ctx = getContext();
+
+    TypeConverter typeConverter;
+    typeConverter.addConversion([](Type t) { return t; });
+    typeConverter.addConversion([](FloatType t) -> Type {
```
brnorris03: Suggest making the float-type conversion reject unsupported widths instead of converting every `FloatType`. For example:

```c++
typeConverter.addConversion([](FloatType type) -> std::optional<Type> {
  if (!type.isF32() && !type.isBF16())
    return std::nullopt;
  return IntegerType::get(type.getContext(), type.getWidth());
});
```

Then the f16/f64 invalid tests can check that unsupported scalar float types fail before the pass partially rewrites the surrounding IR.

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                                inputs)
+          .getResult(0);
+    };
+    typeConverter.addSourceMaterialization(materializeCast);
+    typeConverter.addTargetMaterialization(materializeCast);
+
+    ConversionTarget target(ctx);
+    target.addLegalDialect<ttk::TTKernelDialect>();
+    target.addLegalOp<UnrealizedConversionCastOp>();
+    target.addIllegalOp<arith::CmpFOp>();
```
brnorris03, line 194: Suggest making leftover scalar float/int bridge casts illegal for this pass and testing that directly. Otherwise the pass can succeed while leaving exactly the `iN <-> fN` bridge casts it is meant to remove. For example, mark `UnrealizedConversionCastOp` dynamically legal only when it is not an `iN <-> fN` bridge, or add a final walk that emits an error on remaining bridge casts. Then add `CHECK-NOT: builtin.unrealized_conversion_cast` to the `scalar_cmpf.mlir` cases that should be fully lowered.

include/ttlang/Dialect/TTL/Passes.td
brnorris03, line 738: Suggest moving or renaming this as a TTKernel transform unless there is a TTL invariant that requires it to stay here. It runs after `convert-ttl-to-ttkernel`, creates TTKernel ops, and cleans up casts produced by TTKernel lowering. Keeping it under a TTL name makes the ownership unclear: the pass looks like a TTL transform, but its inputs and outputs are TTKernel-side cleanup. A `ttkernel-lower-scalar-fp-types` name/location would make that ownership clearer.

test/ttlang/Conversion/TTLToTTKernel/scalar_cmpf_invalid.mlir
```diff
@@ -57,34 +58,23 @@ module {
 
 // Unsupported float type f16 (neither f32 nor bf16).
 module {
-  func.func @cmpf_f16_unsupported(%a: f16, %b: f16) -> i1 {
-    // expected-error @below {{unsupported float type for scalar comparison:}}
+  func.func @cmpf_f16_unsupported(%a_int: i16, %b_int: i16) -> i1 {
+    %a = builtin.unrealized_conversion_cast %a_int : i16 to f16
+    %b = builtin.unrealized_conversion_cast %b_int : i16 to f16
+    // expected-error @below {{failed to legalize operation 'arith.cmpf'}}
     %cmp = arith.cmpf ogt, %a, %b : f16
```
brnorris03, line 65: Suggest checking unsupported scalar `arith.cmpf` cases before applying the conversion and emitting a specific diagnostic from `runOnOperation()` with `signalPassFailure()`. For example, walk the operation for `arith::CmpFOp`, reject non-scalar or non-f32/bf16 operands there, then let the conversion pattern handle only supported cases. This test reaches the f16 guard today, but `failed to legalize operation` would also match unrelated pattern failure, so the negative test is not checking the intended failure mode.

test/python/test_ssa_cond_assignment.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+@ttl.operation(grid=(2, 1))
+def dm_if_else_copy_kernel(inp, out):
+    """Per-node copy: node 0 reads/writes tile [0,0], node 1 reads/writes [0,1]."""
+    inp_dfb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=2)
+    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+    # Desired compute -- blocked by ISSUE #683
+    # (convert-ttl-to-compute isBeforeInBlock assert when a value produced
```
brnorris03: Nit: I'd remove this commented-out compute block and future assertions, or moving them to an issue if they are useful. The active test only verifies transfer-handle reassignment, so keeping disabled #683 code here makes the test coverage look broader than it is.

test/ttlang/Conversion/TTLToTTKernel/raw_element_ops.mlir
```diff
@@ -180,12 +180,12 @@ module {
 // -----
 
 // Write f32 constant (1.0) to tiled block via cb_reserve -> get_write_ptr.
-// 1.0f = 0x3F800000 = 1065353216.
+// The constant stays as f32; materializeIntBits inserts unrealized_conversion_cast.
 // CHECK-LABEL: func.func @write_tiled_f32_constant
```
brnorris03, line 185: Replace these comments with the tested behavior, e.g. `RawElementWriteLowering` inserts `unrealized_conversion_cast`, and `ttl-lower-scalar-fp-types` removes the scalar bridge. `materializeIntBits` is deleted in this PR.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -91,6 +91,34 @@ traceTransferHandleSource(mlir::Value value, MatchFn match,
...
+        auto elseYield =
+            mlir::cast<mlir::scf::YieldOp>(ifOp.elseBlock()->getTerminator());
+        llvm::SmallPtrSet<mlir::Value, 16> elseSeen(seen.begin(), seen.end());
+        ResultT elseResult = traceTransferHandleSource<ResultT>(
+            elseYield.getOperand(idx), match, elseSeen);
+        if (!elseResult) {
+          return ResultT();
+        }
+      }
+      return thenResult;
```
zoecarver: Who consumes this result? Do they not need to care about the else branch? 

What about something like this:
```
tx = ttl.copy(pipeA, blk)
if node_x: 
  tx = ttl.copy(pipeB, blk)
tx.wait()
```

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
```diff
@@ -0,0 +1,189 @@
...
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+
+/// Convert arith.truncf (e.g. f32 -> bf16) into integer bit extraction.
+/// bf16 is the upper 16 bits of the f32 IEEE-754 encoding, so a truncf
+/// becomes a right shift by (srcWidth - dstWidth) followed by an integer
+/// truncation.
+struct TruncFToBitExtract : OpConversionPattern<arith::TruncFOp> {
```
zoecarver, line 80: Do we need any dtype validation here or is this correct for all types supported?

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -1464,18 +1464,23 @@ struct RawElementWriteLowering : OpConversionPattern<RawElementWriteOp> {
       return rewriter.notifyMatchFailure(op, "block does not trace to a CB");
     }
 
-    auto intVal =
-        utils::materializeIntBits(adaptor.getValue(), intTy, rewriter, loc);
-    if (failed(intVal)) {
```
zoecarver, line 1469: Is it possible to have a bad value here (what would have taken the `failed` path before)? Looks like `raw_element_ops_invalid.mlir` tested this but was removed.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -91,6 +91,34 @@ traceTransferHandleSource(mlir::Value value, MatchFn match,
         }
       }
     }
+    // Trace through scf.if results -- both branches must independently
+    // yield a valid source.  Each branch is a separate data-flow path,
+    // so give each its own visited set (avoids false negatives when
+    // both branches yield the same outer value).
+    if (auto ifOp = mlir::dyn_cast<mlir::scf::IfOp>(result.getOwner())) {
```
zoecarver: Could this be further generalized to other control flow? I think other parts of this util use a more general interface.

brnorris03 (changes requested):
Thank you for working on this! Some feedback inline. I guess the biggest question (to me) is whether this should be e ttkernel pass, not ttl.

Also, make sure `docs/development/DFBManagement.md` is updated based on the changes in this PR.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
arichinsTT: I think it is fine, as all fp scalar values will be converted. The failure was a case that I thought was invalid, but should be fine now (fp scalar args to a kernel)

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
arichinsTT: I'm not really sure how it would work with looping structures (last iteration only?) but generalized it to non looping regions if thats what you are referring to

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
arichinsTT: Ahhh in this scenario the ast will generate an else branch that yields the handle from the copy from pipeA, but we shouldn't allow different sources, fixing to require all branches in the trace to have the same source.  Also from the feedback above, will check all regions, more than just elses,

lib/Dialect/TTL/Transforms/LowerScalarFpTypes.cpp
arichinsTT: awesome I think I'll keep it simple for now with guards and truncation, but long term I think I want to switch to using metal numeric [llk](https://github.com/tenstorrent/tt-metal/blob/004d4778daf6f35e95159e74c09d4506b898689e/tt_metal/hw/inc/api/numeric/bfloat16.h#L30) so we can match packer/unpacker semantics

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
arichinsTT: added type guards

lib/Dialect/TTKernel/Transforms/TTKernelLowerScalarFpTypes.cpp
arichinsTT: gonna leave it here as its semantically the same, but has easier debugging to differentiate pattern misses

python/pykernel/_src/kernel_ast.py
arichinsTT: add test for coverage of this example and fixed, also handling elifs

brnorris03 (commented):
Thanks for addressing the comments! The conditional-reassignment support and the `LowerScalarFpTypes` rewrite look solid now.

About issue #380 -- the tracer only considers variables that already exist in an outer scope (`_get_if_carried_var_names` skips names not found by `_var_exists`, i.e. "fresh names bound inside a branch stay branch-local"), so a variable first assigned inside both branches still isn't visible after the block. From #380 reproducer:

```python
@ttl.datamovement()
def dm_read():
    node_x, _ = ttl.node(dims=2)
    with inp_dfb.reserve() as blk:
        if node_x == 0:
            tx = ttl.copy(inp[0, 0], blk)
        else:
            tx = ttl.copy(inp[0, 1], blk)
        tx.wait()
```

On the current branch this still raises error: `cannot call .wait() on 'tx': expression does not produce a value`. I suggest either adding variables that are defined in all control flow paths (so would handle the above case), or an appropriate TODO in `_get_if_carried_var_names`.

brnorris03 (approved):

PR #702, opened 2026-06-23
[ttl][nfc] Extract DFB acquire-release analysis

brnorris03:
### Problem description

`ttl-insert-cb-sync` computes ownership between DFB acquires and releases when it inserts missing `ttl.cb_push` and `ttl.cb_pop` operations. The computed-address PipeNet PR needs the same reserve->push and wait->pop ownership rules for pipe receiver-stream proofs, so keeping the logic embedded in `ttl-insert-cb-sync` would duplicate proof rules across passes.

### What's changed

This PR extracts the DFB acquire/release ownership logic without changing `ttl-insert-cb-sync` behavior:

- Adds `DFBAcquireReleaseAnalysis` for producer and consumer DFB ownership intervals.
- Moves acquire/release classification, owned-use discovery, and owned-release search into the shared analysis.
- Rewrites `ttl-insert-cb-sync` to call the shared analysis instead of carrying a private copy of the same interval logic.
- Registers the new analysis source in the TTL transform library.

### Tests

Existing coverage for the behavior-preserving refactor:

- MLIR lit `test/ttlang/Dialect/TTL/Transforms/insert_cb_sync.mlir`: verifies missing `ttl.cb_push` / `ttl.cb_pop` insertion after the last transitive DFB use; preservation of explicit releases; idempotence when the pass runs twice; same-DFB producer/consumer interleavings; nested `scf.for` / `scf.if`; data-movement copy/wait chains; deferred consecutive waits/reserves.
- Device pytest `test/python/test_auto_pop_push.py`: exercises auto push/pop placement through Python-generated kernels, including multiple data-movement uses of one acquired block and deferred producer/consumer lifetimes.
- Device pytest `test/python/test_cb_sync_intrathread_hang.py`: covers the nested intra-thread DFB pattern where an outer reserve must be pushed before a loop body waits on the same DFB.
- Device pytest `test/python/test_cb_sync_intrathread_explicit.py`: covers the same intra-thread structure with explicit push/pop operations, verifying that the sync pass recognizes user-authored releases.
- Device pytest `test/python/test_layernorm.py`: keeps larger-kernel regression coverage for DFB sync placement in fused compute patterns.

PipeNet proof, capacity analysis, computed-address selection, and PipeNet runtime/device tests remain in #700 because that PR adds the PipeNet consumers of this shared analysis.

### Ticket

None.

### Checklist

### DIFF
--- a/lib/Dialect/TTL/Transforms/CMakeLists.txt
+++ b/lib/Dialect/TTL/Transforms/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_mlir_dialect_library(TTLangTTLTransforms
   ConvertTTLToTTKernel.cpp
+  DFBAcquireReleaseAnalysis.cpp
   DFBMaterialization.cpp
   PipeGraph.cpp
   PipeLowering.cpp
--- a/lib/Dialect/TTL/Transforms/DFBAcquireReleaseAnalysis.cpp
+++ b/lib/Dialect/TTL/Transforms/DFBAcquireReleaseAnalysis.cpp
@@ -1,0 +1,281 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "DFBAcquireReleaseAnalysis.h"
+
+#include "ttlang/Dialect/TTL/IR/TTLOps.h"
+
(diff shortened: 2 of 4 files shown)


vtsilytskyiTT (commented):
Pure refactoring, LGTM

zoecarver (approved):

Issue #668, opened 2026-06-06
[REFACTOR] Defer intermediate-DFB materialization to after fusion (extend DFBMaterialization, stacked on #651)


brnorris03:
## Summary

`convert-ttl-to-compute` has to relocate circular-buffer release/consumer ops (`cb_push`, re-consume `cb_wait`, `attach_cb`, `cb_pop`) after the compute it forms, because intermediate DFBs are materialized **before** fusion. Deferring intermediate-DFB materialization to **after** compute formation removes the need for that relocation entirely.

## Current state

`TTLInsertIntermediateDFBs` materializes each SSA intermediate consumed by a `DFBInputOp` (reduce/matmul/bcast/transpose) by emitting `bind_cb`/`cb_reserve`/`store`/`cb_wait`/`attach_cb` at the def site, *before* `convert-ttl-to-compute`. Fusing the producer (and other stores of the same value) into a `ttl.compute` then leaves those CB ops out of canonical order, so `convert-ttl-to-compute::emitTileStores` repairs it by sinking the release/consumer run past the compute. That repair is part of the fragile ordering exposed by the #666 deadlock; it exists only because materialization happens early.

## Proposed refactor: materialize after fusion

Run intermediate-DFB materialization after `convert-ttl-to-compute`. For each `DFBInputOp` whose required operand is not yet CB-attached and is produced by a `ttl.compute`, add that DFB as an **extra output of the producing compute**: `cb_reserve` before the compute, a `tile_store` into its body, `cb_push` after it, and `cb_wait`+`attach_cb` before the consumer. Everything is placed in canonical order by construction, so no CB op is ever relocated and the invariants (reserve ≺ compute ≺ push ≺ wait) hold without fixup.

Payoff: removal of the fragile `emitTileStores` release/consumer relocation, addressing the materialization-order problem exposed by #666.

## Relationship to PR #651

#651 already extracted `materializeToDFB` into a shared `DFBMaterialization` module and applied the deferred pattern to **loop state** (`TTLMaterializeLoopState`) — but it still materializes **intermediates** pre-fusion (the #666 deadlock reproduces on a branch that includes #651). This refactor is the continuation of that work: extend `DFBMaterialization` to intermediates, post-fusion.

[...]

PR #704, opened 2026-06-25
[ttl] Defer intermediate DFB materialization [acc-3]

### LINKED ISSUE #666: [BUG] Compute deadlock: loop-carried recurrence + compiler-managed intermediate DFBs
## Summary

A `ttl.compute` kernel that combines a **loop-carried recurrence** (e.g. a
running `max` updated each iteration) with **compiler-managed intermediate
values** (matmul / reduce / transpose results left as SSA, materialized by
`ttl-insert-intermediate-dfbs`) **deadlocks on device**. Removing only the
loop-carried recurrence — keeping the same matmul/reduce/exp/transpose SSA
chain — makes the identical kernel pass.

Reproduces on `main` (commit 8eb21608) on Blackhole.

## Reproducers (complete source)

Single file, two ops selectable via the `REPRO` env var. `v2_reuse` (no
loop-carried state) **PASSES**; `v3_state` (= v2 **plus** one running-max
recurrence `s = max(s_old, reduce_max(qk))`) **HANGS**. That single added
recurrence is the entire delta.

```python
# _repro_grow.py
import os
import torch
import ttl
import ttnn

TILE = 32
N_ITERS = 4
K = 2
M = 2


def to_dram(dev, t):
    return ttnn.from_torch(
        t.contiguous(), dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
        device=dev, memory_config=ttnn.DRAM_MEMORY_CONFIG,
    )


@ttl.operation(grid=(1, 1))
def v2_reuse(a, b, c, out):
    a_cb = ttl.make_dataflow_buffer_like(a, shape=(1, K), block_count=2)
    b_cb = ttl.make_dataflow_buffer_like(b, shape=(M, K), block_count=2)
    qk_cb = ttl.make_dataflow_buffer_like(a, shape=(1, M), block_count=2)
    mx_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
    es_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
    out_cb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=N_ITERS)

    @ttl.compute()
    def compute():
        for _ in range(N_ITERS):
            a_blk = a_cb.wait()
            b_blk = b_cb.wait()
            qk_w = qk_cb.reserve(); qk_w.store(a_blk @ ttl.transpose(b_blk))
            qk = qk_cb.wait()
            mx_w = mx_cb.reserve(); mx_w.store(ttl.math.reduce_max(qk, dims=[1]))
            mx = mx_cb.wait()
            mx_bc = ttl.block.broadcast(mx, dims=[1], shape=(1, M))
            ex = ttl.exp(ttl.sub(qk, mx_bc))
            es_w = es_cb.reserve(); es_w.store(ttl.math.reduce_sum(ex, dims=[1]))
            es = es_cb.wait()
            ow = out_cb.reserve(); ow.store(es)

    @ttl.datamovement()
    def dm():
        for _ in range(N_ITERS):
            ad = a_cb.reserve(); ttl.copy(a[0:1, 0:K], ad)
            bd = b_cb.reserve(); ttl.copy(b[0:M, 0:K], bd)
        for _ in range(N_ITERS):
            ob = out_cb.wait(); ttl.copy(ob, out[0:1, 0:1])

    @ttl.datamovement()
    def dm_unused():
        pass


@ttl.operation(grid=(1, 1))
def v3_state(a, b, c, out):
    a_cb = ttl.make_dataflow_buffer_like(a, shape=(1, K), block_count=2)
    b_cb = ttl.make_dataflow_buffer_like(b, shape=(M, K), block_count=2)
    qk_cb = ttl.make_dataflow_buffer_like(a, shape=(1, M), block_count=2)
    mx_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
    es_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
    st_cb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)  # loop-carried state
    out_cb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=N_ITERS)

    @ttl.compute()
    def compute():
        s0 = st_cb.reserve(); s0.store(ttl.block.fill(-1e30, shape=(1, 1)))
        for _ in range(N_ITERS):
            a_blk = a_cb.wait()
            b_blk = b_cb.wait()
            qk_w = qk_cb.reserve(); qk_w.store(a_blk @ ttl.transpose(b_blk))
            qk = qk_cb.wait()
            mx_w = mx_cb.reserve(); mx_w.store(ttl.math.reduce_max(qk, dims=[1]))
            mx = mx_cb.wait()
            s_old = st_cb.wait()                          # <-- recurrence
            s_new = ttl.math.max(s_old, mx)               # <-- recurrence
            s_next = st_cb.reserve(); s_next.store(s_new)  # <-- recurrence
            mx_bc = ttl.block.broadcast(s_new, dims=[1], shape=(1, M))
            ex = ttl.exp(ttl.sub(qk, mx_bc))
            es_w = es_cb.reserve(); es_w.store(ttl.math.reduce_sum(ex, dims=[1]))
            es = es_cb.wait()
            ow = out_cb.reserve(); ow.store(es)
        _ = st_cb.wait()

    @ttl.datamovement()
    def dm():
        for _ in range(N_ITERS):
            ad = a_cb.reserve(); ttl.copy(a[0:1, 0:K], ad)
            bd = b_cb.reserve(); ttl.copy(b[0:M, 0:K], bd)
        for _ in range(N_ITERS):
            ob = out_cb.wait(); ttl.copy(ob, out[0:1, 0:1])

    @ttl.datamovement()
    def dm_unused():
        pass


def main():
    which = os.getenv("REPRO", "v2_reuse")
    fn = {"v2_reuse": v2_reuse, "v3_state": v3_state}[which]
    dev = ttnn.open_device(device_id=0)
    try:
        a = to_dram(dev, torch.randn(TILE, K * TILE, dtype=torch.bfloat16))
        b = to_dram(dev, torch.randn(M * TILE, K * TILE, dtype=torch.bfloat16))
        c = to_dram(dev, torch.randn(M * TILE, TILE, dtype=torch.bfloat16))
        out = to_dram(dev, torch.zeros(TILE, TILE, dtype=torch.bfloat16))
        print(f"running {which}...")
        fn(a, b, c, out)
        ttnn.synchronize_device(dev)
        print(f"REPRO {which} PASS")
    finally:
        ttnn.close_device(dev)


if __name__ == "__main__":
    main()
```

Run:

```
REPRO=v2_reuse python _repro_grow.py   # prints "REPRO v2_reuse PASS"
REPRO=v3_state python _repro_grow.py   # hangs (no output after "Compiled kernel ready")
```

## Observed behavior

- `v2_reuse`: completes, `REPRO v2_reuse PASS`.
- `v3_state`: kernels compile and dispatch, then the device hangs forever
  (with `TT_METAL_WATCHER=10` the device keeps being polled; no progress).

## Evidence / analysis

- **Not a circular-buffer logic bug.** The IR after `ttl-insert-cb-sync` is
  balanced: every `cb_reserve`/`cb_push` and `cb_wait`/`cb_pop` pairs up
  correctly across the producer/consumer threads.
- **Not back-pressure depth.** Raising the cross-thread DFB `block_count`
  (k/v) from 2 to 4 does **not** fix it.
- **Watcher** (Blackhole) on the single active worker core shows the
  data-movement RISC (NCRISC) blocked at `CRBW` (`cb_reserve_back` wait); the
  compute TRISC waypoints are at *Done* markers (`UABD`, `MWDD`), i.e. the
  compute pipeline is stalled and back-pressures the producer.
- After `ttl-assign-dst`, the loop body is split into many `ttl.dst_section`s
  (one per compute op). The recurrence adds a value (`s_new`) that is
  multi-used — materialized into a compiler-allocated DFB for the broadcast
  **and** written back to the state DFB — which appears to tip the
  DST-section / unpack-math-pack scheduling into a deadlocking order.

Conclusion: the deadlock is in DST-register / `ttl.dst_section` scheduling
(`ttl-assign-dst` -> `ttl-lower-to-loops`), surfaced by combining a
loop-carried recurrence with compiler-managed intermediate DFBs.

## Workaround

Manually stage **every** intermediate in its own single-use DFB instead of
relying on `ttl-insert-intermediate-dfbs` (this keeps DST live-ranges short).
A fully hand-staged flash-attention `flash_shard_kd` (~25 DFBs) runs correctly;
reducing it to rely on compiler-managed intermediates reintroduces this hang.

## Impact

Blocks authoring minimal-DFB online-softmax / flash-attention-style kernels:
any loop-carried recurrence (running max/sum/output) combined with
compiler-materialized op intermediates can deadlock, forcing users to
hand-stage all intermediates.

brnorris03:
Chain:

```text
#704 (this PR)
|-> #733
`-> #778 -> #775
```

### Problem description

Compiler-allocated intermediate DFBs were materialized before `ComputeOp`
creation. Later conversion then had to infer legality and repair ordering from
IR that earlier rewrites had already changed. In #666, a loop-state update plus
a compiler-managed intermediate could place the consumer wait before the
producer push and hang on device.

The same materialization area must preserve SSA dominance for branch-local
DFB-input consumers in #683-class control flow and reject invalid DFB
lifecycles before changing IR.

### What's changed

- Computes immutable per-kernel `ComputeOp` creation and materialization plans
  before rewriting IR. DFB value availability uses MLIR dataflow analysis and
  treats unresolved reachable control flow conservatively.
- Moves intermediate DFB insertion after producer `ComputeOp` creation and
  removes the redundant pre-creation `ttl-auto-sync` stage.
- Rebuilds each `ttl.compute` once with all required compiler-allocated DFB
  outputs. When one result requires materialization, every surviving consumer
  reads the same pushed and waited attachment.
- Validates DFB acquisition and release counts before mutation. Duplicate
  releases now produce a compiler diagnostic; this exposed real redundant
  releases in context-managed broadcast and fused-matmul kernels that main
  accepted.
- Keeps final DFB-input consumer conversion in `convert-ttl-to-compute` without
  compiler-intermediate lifecycle relocation.
- Adds deterministic analysis printers and documents the planning algorithms,
  correctness argument, assumptions, limitations, and future work.

### Test coverage

The representative hardware case publishes an intermediate to a user DFB while
the same value feeds reduction, broadcast, and elementwise consumers.
Compilation depends on one shared compiler DFB for the surviving consumers, and
runtime checks both published results.

MLIR coverage also exercises loops, control flow, chained fixed-point
requirements, multi-output producers, f32 multi-tile matmul, invalid
lifecycles, and final conversion. Device coverage validates recurrence,
branch-local consumers, and fused post-ops across the supported configurations.

### Tickets

Fixes #666.
Follow-up to #668.
Partially addresses #683 by fixing branch-local compiler-intermediate
materialization; #683 remains open for cross-branch store fanout.

### Performance

This PR changes compiler correctness and does not add an optimization. The
measurements below establish the baseline for the subsequent in-DST
accumulation work in #733 (see the `bnorris/microbenchmarks` branch).

| Variant | us/accumulation | Fixed us | vs acc_dst |
|---|---:|---:|---:|
| ttlang | 0.460 | 0.910 | 10.9x |
| cpp_acc_dst | 0.042 | 0.898 | 1.0x |
| cpp_acc_l1 | 0.038 | 0.792 | 0.9x |


lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
```diff
@@ -6,9 +6,164 @@
...
+  BindCBOp bindDFB;
+  {
+    OpBuilder::InsertionGuard guard(builder);
+    bindDFB =
+        createCompilerAllocatedDFB(tensorType, loc, funcOp, moduleOp, builder);
+  }
+
+  builder.setInsertionPoint(computeOp);
+  auto reserve =
+      CBReserveOp::create(builder, loc, tensorType, bindDFB.getResult());
```
zoecarver: Hmm, so let's say this is around a branch, we unconditionally reserve/push before the if-statement, but then conditionally wait/pop inside the branch? I think that could produce a hang.

zoecarver: A meta point: DFB materializatoin should only happen for thread-local DFBs, right? So why do we even need synchronization at all... could we just reserve at the top of the scope and use it as scratch throughout the body?

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
```diff
@@ -6,9 +6,164 @@
...
+  Value init =
+      createEmptyTensorLike(tensorType, loc, computeOp.getInputs(), builder);
```
zoecarver: `computeOp.getInputs()` seems problematic. Can you add a test for where the compute input is a different shape than the use here? I'm thinking a broadcast would work. Wouldn't this cause us to create an incorrect size?

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
```diff
@@ -77,7 +85,12 @@ struct TTLInsertIntermediateDFBsPass
     }
 
     OpBuilder builder(funcOp.getContext());
-    llvm::DenseMap<Value, Value> materialized;
+    DominanceInfo dominanceInfo(funcOp);
+    // A shared consumer-side acquire is valid only when its attach dominates
+    // the next consumer. Incomparable control-flow regions need separate
+    // compiler DFB outputs so each dynamic execution consumes exactly one
+    // pushed slot.
+    llvm::DenseMap<Value, SmallVector<Value>> materialized;
```
zoecarver: Are you planning to re-visit re-use in a follow up? I think we can do a lot better here. We should allow re-using CBs that have different dtype or shape, taking the largest logical size. 

I also think this might not handle the case where a ttl.compute returns multiple outputs that each feed out of order consumers. E.g.

```
a, b = ttl.compute

ttl.compute(a) # materialized rebuilt map key = a
ttl.compute(b) # rebuilt again, a is now stale 
ttl.compute(a) # should re-use a, but cannot even though it dominates here 
```

If it's easy to model a multi-output re-use, maybe a good thing to have a test for, as I'm struggling a bit to map this in my head 😁

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
brnorris03: > Hmm, so let's say this is around a branch, we unconditionally reserve/push before the if-statement, but then conditionally wait/pop inside the branch? I think that could produce a hang.

Are you asking that analysis be added for this scenario? There is no current static analysis checking for unconditional reserves followed by conditional wait/pops

zoecarver:
While you're here, maybe we can also make the default `block_count=1` for these since they're thread local I don't think we get anything from double buffering.

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
brnorris03: There were two uses of `getInputs()` here. Passing it as the compute inputs is fine — the rebuilt op just duplicates the selected output, so the inputs and iteration domain don't change and a broadcast input can't affect the new output's size.

The issue was using it as the shape exemplar in `createEmptyTensorLike`: static dims came from the result type, but dynamic dims were read off the first input via `tensor.dim`, which is the wrong extent under a broadcast. Added `broadcast_result_then_reduce` (`[1,1]` -> `[1,2]` -> reduce) to confirm that the DFB and extra output size comes from the broadcast result, not the input.

We don't support dynamic tile shapes, meaning that branch is never exercised atm with the current tt-lang, so instead of leaving that code in, I added an ~~error~~ assert for non-static shapes (ttl.compute verifier ensures static shapes).

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
brnorris03: I think that's definitely doable, but I would prefer to do in a separate subsequent PR and just tighten the constraints on what is supported (control-flow-wise) in this one -- emitting errors that will later be removed once the analysis is in.

brnorris03: >A meta point: DFB materializatoin should only happen for thread-local DFBs, right? So why do we even need synchronization at all... could we just reserve at the top of the scope and use it as scratch throughout the body?

Not quite -- the packer and unpacker are separate processors as you know so there are three concurrent threads involved in execution operations from the in compute kernel. For loops, having block count 1 can end up clobbering a value that is not completely read yet; 2 is not guaranteeing this can never happen, the proper control flow analysis is required to do this correctly long term. Issue #724 created to add this. 

After rereading the metal docs, I think that setting block count to 1 should actually work. Need to figure out why it doesn't (a number of tests fail).

brnorris03:
> While you're here, maybe we can also make the default `block_count=1` for these since they're thread local I don't think we get anything from double buffering.

This is only generally (obviously) correct when `dst_full_sync_en=true` (so can do that then for sure), when packer and compute are serialized. I can certainly check the option and set it then.

If there is an actual guarantee somewhere that single buffering ensures sequential access (I think the metal docs imply it), then I think it should work as you suggest, but I believe it doesn't currently (at least when I tried I ran into frontend issues).

Update: It's 1 by default now (the issues I had run into earlier were not related to packer/unpacker concurrency); there is an issue to possibly optimize later if microbenchmarks indicate it's worth it.

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
brnorris03: Okay, the default block count is 1 now and optimizing it is issue #727 (if microbenchmarks determine it's worth doing).

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
brnorris03: Sorry, misunderstood the question initially. Looking into it more carefully, I don't think that compute should be rebuilt. I will change this to build each compute exactly once.

brnorris03: There is no more multiple compute rebuilds (new details in the `DFBManagement.md` doc); also added the above scenario to both the mlir and pytests.

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
```diff
@@ -36,17 +39,333 @@ namespace mlir::tt::ttl {
...
+    if (!clonedStore) {
+      continue;
+    }
+
+    Value storeDFB = getAttachedCB(clonedStore.getView());
+    for (MaterializedOutput &output : materializedOutputs) {
+      if (storeDFB != output.sourceDFB) {
+        continue;
+      }
+      auto materializedStore = TileStoreOp::create(
```
zoecarver, line 103: I am confused about the semantics of this. Can you explain the situations in which we implicitly materialize a store? 

If any assignment turns into a store, then why do we have the `.store` operation, and if it only sometimes turns into a store, that seems really confusing because the user would have to run through the rules in their head to know whether an assignment would actually write to memory or not.

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
```diff
@@ -36,17 +39,333 @@ namespace mlir::tt::ttl {
...
+  unsigned resultIndex;
+  SmallVector<ConsumerUse> uses;
+};
+
+struct ComputeMaterializationPlan {
+  ComputeOp producer;
+  SmallVector<ResultMaterializationPlan> results;
+};
+
+struct StandaloneTensorMaterializationUse {
```
zoecarver: Can this just be `ConsumerUse`?

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
```diff
@@ -36,17 +39,333 @@ namespace mlir::tt::ttl {
...
+  }
+
+  MaterializedOutput output;
+  output.sourceResultIndex = resultIndex;
+  output.tensorType = tensorType;
+  output.sourceDFB = sourceDFB;
+  return output;
+}
+
+static LogicalResult cloneComputeBodyWithMaterializedStores(
```
zoecarver: nit: never returns failure?

test/python/test_recurrence_multi_output_dfb.py
```diff
@@ -0,0 +1,348 @@
...
+
+    @ttl.compute()
+    def compute():
+        for i in range(N_BRANCH_ITERS):
+            a_block = a_dfb.wait()
+            b_block = b_dfb.wait()
+            sum_block = a_block + b_block
+            with sum_dfb.reserve() as sum_out:
+                sum_out.store(sum_block)
+            if i < 1:
```
zoecarver: Hmm, this might not be expressible today, but I think this won't compile to a branch, so I don't know if this test is testing what it thinks it is.

zoecarver: Maybe we can use an scalar read in the condition to force it to be a runtime decision?

test/ttlang/Dialect/TTL/Transforms/insert_intermediate_dfbs_deferred.mlir
```diff
@@ -0,0 +1,348 @@
...
+// FULL-LABEL: func.func @branch_local_reductions
+// FULL: %[[INTERMEDIATE_DFB:.*]] = ttl.bind_cb{{.*block_count = 1.*}} {ttl.compiler_allocated} : <[1, 1], {{.*}}, 1>
+// FULL: ttl.cb_push %[[INTERMEDIATE_DFB]]
+// FULL: %[[INTERMEDIATE_WAIT:.*]] = ttl.cb_wait %[[INTERMEDIATE_DFB]]
+// FULL: %[[INTERMEDIATE_ATTACHED:.*]] = ttl.attach_cb %[[INTERMEDIATE_WAIT]], %[[INTERMEDIATE_DFB]]
+// FULL: scf.if
+// FULL: ttl.compute ins(%[[INTERMEDIATE_ATTACHED]],
+// FULL: else
+// FULL: ttl.compute ins(%[[INTERMEDIATE_ATTACHED]],
+// FULL: ttl.cb_pop %[[INTERMEDIATE_DFB]]
```
zoecarver, line 235: nit: add a check: } to make sure this doesn't match some random pop elsewhere

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
brnorris03: Assignments don't (can't) implicitly become stores. `ttl.compute` already contains a `ttl.tile_store` produced by the explicit block (user) `store`; materialization preserves that store and replicates its tile into a compiler DFB only when a downstream operand requires a DFB-attached value. I clarified the semantics and algorithm in [`DFBManagement.md`](https://github.com/tenstorrent/tt-lang/blob/cb65f96609975ab81af89139ed3901868ee18cde/docs/development/DFBManagement.md#L248-L275) and the replication site in [`TTLInsertIntermediateDFBs.cpp`](https://github.com/tenstorrent/tt-lang/blob/cb65f96609975ab81af89139ed3901868ee18cde/lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp#L168-L187).

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
brnorris03: Yes, good idea. I removed the duplicate type and now use `ConsumerUse` for both compute-result and standalone tensor materialization uses.

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
brnorris03: Good catch. The helper now returns `void`, and I removed the unreachable failure branch at its call site in `TTLInsertIntermediateDFBs.cpp`.

test/python/test_recurrence_multi_output_dfb.py
brnorris03: Agreed. I replaced it with a runtime `ttl.node()` x-coordinate condition on a two-core grid, so the generated compute kernel contains an `if` and the hardware test executes both branches: [`test_recurrence_multi_output_dfb.py`](https://github.com/tenstorrent/tt-lang/blob/cb65f96609975ab81af89139ed3901868ee18cde/test/python/test_recurrence_multi_output_dfb.py#L198-L233). `raw_element_read` is restricted to data movement kernels, so the node coordinate keeps this test in the compute kernel.

test/ttlang/Dialect/TTL/Transforms/insert_intermediate_dfbs_deferred.mlir
brnorris03: Added `FULL-NOT: ttl.cb_pop` before the expected compiler-DFB pop (I don't think the exact same pop with that intermediate itself is worth checking for, but correct me if I'm misunderstanding).

zoecarver (approved):

PR #705, opened 2026-06-25
[nfc] Uplift toolchain dependencies

brnorris03:
Uplift of LLVM, TT-Metal, and tt-mlir plus associated test updates.

| Package | SHA | Date | Ref |
|---|---|---|---|
| third-party/llvm-project | e568ab3b0fceeeba1b43400b199ab07f719b1092 | 2026-06-29T10:18:57+08:00 | main |
| third-party/tt-metal | c04ae2758fc87f3a49ca19a7d339464db90e995d | 2026-06-28T23:25:10Z | v0.74.0-dev20260629 |
| third-party/tt-mlir | ba8aa2aca750fa530773edf5c903bf3964915bd1 | 2026-06-27T15:58:28Z | main |

ttnn PyPI pin (`third-party/tt-metal-version`): `0.73.1`, built from tt-metal `v0.73.1-rc5`. `TT_METAL_TAG` builds against `v0.74.0-dev20260629`.


zoecarver (approved):

cmake/modules/BuildTTMetal.cmake
```diff
@@ -220,6 +220,12 @@ endif()
 # Sentinel file: if this exists, tt-metal is already built.
 set(_TTNN_SO "${TTMETAL_BUILD_DIR}/ttnn/_ttnn.so")
 
+if((TTLANG_BUILD_TOOLCHAIN OR TTLANG_FORCE_TOOLCHAIN_REBUILD) AND
```
vtsilytskyiTT, line 223: What is the difference between these two variables?

test/ttlang/Translate/TTLToCpp/dma_loop_multi_tile_nontrivial_cb.mlir
```diff
@@ -14,25 +14,25 @@
 //   for (USER_ITER in 0..4) {              ← User loop from input MLIR
 //     for (tile_y in 0..2) {               ← Generated tile loop for Tensor1
 //       for (tile_x in 0..2) {             ← Generated tile loop for Tensor1
-//         noc_async_read_tile(...)         ← Tensor1 → CB1
+//         noc0.async_read(...)         ← Tensor1 → CB1
```
vtsilytskyiTT: nit: alignment (here and below)

third-party/patches/ttmlir-memref-to-emitc-typeconv.patch
vtsilytskyiTT, line 1: Are we going merge this directly into repo? I hope the patch is a temporary solution. Maybe open a ticket to keep track of this.

requirements-runtime.txt
```diff
@@ -1,7 +1,11 @@
 # Runtime dependencies for the tt-lang wheel; also read by pyproject.toml.
 pydantic<3
 torch>=1.9.0
-numpy>=1.24.4,<2
+numpy>=1.20.0
```
vtsilytskyiTT, line 4: I didn't find an explanation for numpy downgrade. Please add the justification somewhere (PR description/comment/issue).

cmake/modules/BuildTTMetal.cmake
brnorris03: One indicates that the toolchain should be built. The other forces a rebuild if it was already built and installed.

requirements-runtime.txt
brnorris03: It's an upgrade -- not downgrade. It's explained in the subsequent lines, which are comments in front of the original restriction, which applies only to old intel macs. Limiting to <2 makes tt-lang incompatible with tt-xla.

third-party/patches/ttmlir-memref-to-emitc-typeconv.patch
brnorris03: This is because tt-mlir has not updated their llvm since last October. They are very aware of that. There is no need for a tt-lang ticket since it's not anything in tt-lang. We have another patch for similar reasons. It's temporary if tt-mlir ever does uplift, otherwise not.

requirements-runtime.txt
vtsilytskyiTT: Indeed, my bad. You extended range from [1.24.4, 2) to [1.20.0, +inf). That is an upgrade.
I was confused by 1.24 -> 1.20, which is not explained by the comment below.

vtsilytskyiTT (approved):

requirements-runtime.txt
brnorris03: I think that's what our original lower limit was before the mac-x86-inspired update, so this just reverts to what it was before while adding the mac special case.

phizalev-TT (approved):

Issue #612, opened 2026-05-21
fp32 `0.5 * a + b` corrupts when scalar unary feeds SFPU add


brnorris03:
### Problem

A standalone fp32 kernel for `0.5 * a_blk + b_blk` compiles and runs, but the output contains corrupted fp32 values. The equivalent bf16 case passes.

### Standalone Reproducer

Save this as `repro_scalar_mul_fp32.py` in a tt-lang checkout and run it with the normal test environment:

```bash
python repro_scalar_mul_fp32.py
```

```python
import torch
import ttnn

import ttl


TILE = 32


@ttl.operation(grid=(1, 1))
def scalar_mul_add_kernel(input_a, input_b, output):
    input_a_dfb = ttl.make_dataflow_buffer_like(
        input_a, shape=(2, 2), block_count=2
    )
    input_b_dfb = ttl.make_dataflow_buffer_like(
        input_b, shape=(2, 2), block_count=2
    )
    output_dfb = ttl.make_dataflow_buffer_like(
        output, shape=(2, 2), block_count=2
    )

    @ttl.compute()
    def compute_fn():
        with (
            input_a_dfb.wait() as input_a_block,
            input_b_dfb.wait() as input_b_block,
            output_dfb.reserve() as output_block,
        ):
            output_block.store(0.5 * input_a_block + input_b_block)

    @ttl.datamovement()
    def dm_read():
        with input_a_dfb.reserve() as input_a_block:
            ttl.copy(input_a[slice(0, 2), slice(0, 2)], input_a_block).wait()
        with input_b_dfb.reserve() as input_b_block:
            ttl.copy(input_b[slice(0, 2), slice(0, 2)], input_b_block).wait()

    @ttl.datamovement()
    def dm_write():
        with output_dfb.wait() as output_block:
            ttl.copy(output_block, output[slice(0, 2), slice(0, 2)]).wait()


def to_l1_tensor(torch_tensor, device):
    ttnn_dtype = ttnn.float32 if torch_tensor.dtype == torch.float32 else ttnn.bfloat16
    tensor = ttnn.from_torch(
        torch_tensor,
        dtype=ttnn_dtype,
        layout=ttnn.TILE_LAYOUT,
        device=device,
        memory_config=ttnn.DRAM_MEMORY_CONFIG,
    )
    return ttnn.to_memory_config(tensor, memory_config=ttnn.L1_MEMORY_CONFIG)


def main():
    torch.manual_seed(0)
[...]

PR #706, opened 2026-06-25
Use correct unpack mode for fp32

vtsilytskyiTT:
### Problem description
The `TTLSetComputeKernelConfig` pass in `TTLSetComputeKernelConfig.cpp` misclassifies non-FPU-eligible strategy-dependent binary ops (like `ttl.tile_add` when LHS is not a block arg). These ops lower to `add_binary_tile` (DST-based SFPU path), but since they lack `TTLDSTInputsTrait`, their f32 CB operands don't get added to sfpuCBs, so CB1 isn't marked for UnpackToDestFp32. When copy_tile_init(CB1) runs in Default mode, FP32 data is read as BF16, causing the massive errors.

### What's changed
Strategy-dependent binary ops that are NOT FPU-eligible will lower to `add_binary_tile` (DST-path), so their f32 CB inputs need UnpackToDestFp32.

### Ticket
https://github.com/tenstorrent/tt-lang/issues/612

### Checklist

### DIFF
--- a/lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
@@ -77,7 +77,9 @@ static bool isDstInputTileComputeOp(Operation *op) {
   if (isFPUEligibleBinaryOp(op)) {
     return false;
   }
-  return op->hasTrait<TTLDSTInputsTrait>() ||
+
+  return op->hasTrait<TTLStrategyDependentBinaryOpTrait>() ||
+         op->hasTrait<TTLDSTInputsTrait>() ||
          isa<TileBcastOp, TileTransposeOp>(op);
 }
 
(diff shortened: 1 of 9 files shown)


zoecarver (approved):
Awesome!!

lib/Dialect/TTL/Transforms/TTLSetComputeKernelConfig.cpp
```diff
@@ -77,7 +77,9 @@ static bool isDstInputTileComputeOp(Operation *op) {
   if (isFPUEligibleBinaryOp(op)) {
     return false;
   }
-  return op->hasTrait<TTLDSTInputsTrait>() ||
+
+  return op->hasTrait<TTLStrategyDependentBinaryOpTrait>() ||
+         op->hasTrait<TTLDSTInputsTrait>() ||
          isa<TileBcastOp, TileTransposeOp>(op);
 }
```
brnorris03, line 84: I think that it's not completely correct to just add this as a static check since TTLStrategyDependentBinaryOpTrait doesn't necessarily imply that this will have DST inputs. Given that this is a helper used in only one place, I think the better solution would be to remove it and just perform the correct checks in the (current) caller location.

PR #707, opened 2026-06-25
Simulator: Add multidevice functional support

FConstantinos:


python/sim/ttnnsim.py
```diff
@@ -2375,7 +2468,7 @@ def all_reduce(
 
     t = input_tensor.to_torch()
 
-    axes = [cluster_axis] if cluster_axis is not None else range(2)
+    axes = [cluster_axis] if cluster_axis is not None else range(len(msi.mesh_shape))
```
brnorris03: I'd add bounds-checking `cluster_axis` before it indexes `msi.dims[k]` to produce a more helpful error than the IndexError you'd get otherwise.

```suggestion
    if cluster_axis is not None and not (0 <= cluster_axis < len(msi.mesh_shape)):
        raise ValueError(
            f"cluster_axis {cluster_axis} out of range for a {len(msi.mesh_shape)}-axis "
            f"mesh {tuple(msi.mesh_shape)}"
        )
    axes = [cluster_axis] if cluster_axis is not None else range(len(msi.mesh_shape))
```

The docstring above ("For 2D meshes", "cluster_axis (0 or 1)") is also outdated now that N-D is supported.

python/sim/ttnnsim.py
```diff
@@ -2454,7 +2547,7 @@ def all_gather(
...
```
brnorris03: Same as `all_reduce`: consider validating `cluster_axis` against `len(msi.mesh_shape)` here?

```suggestion
    if cluster_axis is not None and not (0 <= cluster_axis < len(msi.mesh_shape)):
        raise ValueError(
            f"cluster_axis {cluster_axis} out of range for a {len(msi.mesh_shape)}-axis "
            f"mesh {tuple(msi.mesh_shape)}"
        )
    axes = [cluster_axis] if cluster_axis is not None else range(len(msi.mesh_shape))
```

python/sim/copyhandlers.py
```diff
@@ -84,6 +84,22 @@ def _get_or_create_pipe_entry(pipe: AnyPipe) -> PipeEntry:
...
+    mesh axes); on-chip NoC pipes and single-device grids return False.  Used
+    only to annotate trace events; it does not affect transfer semantics.
+    """
+    from .nodecontext import pipe_crosses_mesh
+
+    scheduler = get_context().scheduler
+    grid = tuple(getattr(scheduler, "grid", ()) or ()) if scheduler is not None else ()
+    if not grid:
+        return False
+    return pipe_crosses_mesh(pipe.src, pipe.dst, grid)
```
brnorris03: `pipe_crosses_mesh` raises `ValueError` for a pipe whose coordinates are shorter than the grid rank on an N-D grid. This helper only annotates trace events (see the docstring), so enabling tracing would crash a run that succeeds with tracing off. Consider making the classification non-raising, or validating pipe-coordinate rank at `Pipe`/`PipeNet` construction (the spec already requires node ranges to match grid rank)?

python/sim/copy.py
```diff
@@ -125,6 +126,19 @@ def __init__(
...
+        match src:
+            case Tensor():
+                validate_mesh_access(src, "read")
+            case _:
+                pass
+        match dst:
+            case Tensor():
+                validate_mesh_access(dst, "write")
+            case _:
+                pass
```
brnorris03: I think the simpler `isinstance` is sufficient here perhaps?

```suggestion
        if isinstance(src, Tensor):
            validate_mesh_access(src, "read")
        if isinstance(dst, Tensor):
            validate_mesh_access(dst, "write")
```

python/sim/ttnnsim.py
```diff
@@ -1046,19 +1046,52 @@ def set_fabric_config(config: Any) -> None:
...
-    def __init__(self, rows: int, cols: int) -> None:
-        self.rows = rows
-        self.cols = cols
+    Accepts any number of axis sizes, e.g. ``MeshShape(8)`` (1D),
+    ``MeshShape(2, 4)`` (2D), or ``MeshShape(2, 2, 2)`` (3D).  ``rows`` and
+    ``cols`` remain available for 2D call sites; for higher-rank meshes use
+    :attr:`dims` (or indexing / iteration) directly.
+    """
+
+    def __init__(self, *dims: int) -> None:
```
brnorris03, line 1057: MeshShape now defines `__repr__`/`__len__`/`__iter__`/`__getitem__` but no `__eq__`, so `MeshShape(2, 2) == MeshShape(2, 2)` is `False`. Not sure it's intended, but seems wrong so perhaps add `__eq__`/`__hash__`?

python/sim/ttlang_sim.py
```diff
@@ -217,7 +217,11 @@ def _write_jsonl_trace(path: Path, events: list) -> None:
     """Write trace events to a JSON Lines file.
 
     Each line is a self-contained JSON object with flat fields:
-    event, tick, kernel, and any event-specific data.
+    event, tick, kernel, node, device, and any event-specific data.
+
+    ``node`` is the linear-node label and ``device`` is the device-mesh
```
brnorris03: Is `node` the global node index or the local within-device index?

docs/TRACING.md
```diff
@@ -65,6 +65,31 @@ A span covering each `copy()` call:
...
+
+Instant events at each pipe send/receive:
+
+- **pipe_send** -- a `copy(block, pipe)` posted a payload to a pipe. Fields: pipe
+  name, tile count, `fabric`, node, kernel.
+- **pipe_recv** -- a `copy(pipe, block)` drained a payload from a pipe. Fields: pipe
+  name, tile count, `fabric`, node, kernel.
+
+The boolean **fabric** field classifies the transfer by topology, derived from the
+pipe's endpoints and the launch grid: a grid's leading `len(grid) - 2` dimensions are
```
brnorris03: I don't think the simulator defines topologies matching the actual hardware? I am perhaps reading too much in the use of "topology" here, but maybe worth making it clear whether that's simulator-specific or reflecting actual valid topologies.

brnorris03 (commented):
Thank you for adding this, will be great for debugging!

brnorris03:
Some additional tests to consider as you see fit
1. Device-scoped ownership on a real core grid (>1 core per device)

Every mesh-access and device-attribution test uses a 1×1 core grid ((2,1,1), (2,2,1,1)), so one node = one device. But real launch grids usually give each device a larger core grid — the PR's own example is (2,2,8,8). So I think right now nothing tests that ownership is per-device, not per-node.

On grid (2,2,2) (2 devices × 2×2 cores): all four nodes of device 0 may read device 0's shard; any device-0 node reading device 1's shard raises. I verified this works, so no bug — it just isn't covered.

2. Multi-tile-per-device shard (kernel loops over its local tiles)

Every device currently owns exactly one tile (64 rows / 2 = 32 = one tile). Real kernels own many tiles and loop over them. E.g. 128 rows / 2 devices = 2 tiles/device: device 0 reads tiles 0 and 1 (allowed), tile 2 raises. Just confirms that the owned-slice math holds across a multi-tile boundary, not just at a single tile.

3. Column / last-dim shard (matmul-style)

Access tests shard dim 0 (rows). Matmul shards the contraction/output dim, i.e. the last dim — `ShardTensorToMesh(dim=1)` on a 3D grid. I verified that a device owning `cols[64:128)` and reading a `cols[0:32)` tile raises an exception, so just adding coverage (not a bug).

4. Replicated + sharded in one kernel

Real data-parallel: replicated weights/bias (every device reads the whole tensor) alongside a sharded input (each device reads only its shard). `ReplicateTensorToMesh` sets no `mesh_shard_info` so it's skipped. I also verified that the replicated read by a foreign device is allowed, so not a bug, again just better coverage.

python/sim/copyhandlers.py
FConstantinos: fixed

python/sim/ttnnsim.py

python/sim/ttnnsim.py

python/sim/ttnnsim.py
FConstantinos: fixed

docs/TRACING.md
FConstantinos: made it more clear

python/sim/ttlang_sim.py
FConstantinos: global one

python/sim/copy.py
FConstantinos: fixed

FConstantinos:
> Some additional tests to consider as you see fit
> 
> 1. Device-scoped ownership on a real core grid (>1 core per device)
> 
> Every mesh-access and device-attribution test uses a 1×1 core grid ((2,1,1), (2,2,1,1)), so one node = one device. But real launch grids usually give each device a larger core grid — the PR's own example is (2,2,8,8). So I think right now nothing tests that ownership is per-device, not per-node.
> 
> On grid (2,2,2) (2 devices × 2×2 cores): all four nodes of device 0 may read device 0's shard; any device-0 node reading device 1's shard raises. I verified this works, so no bug — it just isn't covered.
> 
> 2. Multi-tile-per-device shard (kernel loops over its local tiles)
> 
> Every device currently owns exactly one tile (64 rows / 2 = 32 = one tile). Real kernels own many tiles and loop over them. E.g. 128 rows / 2 devices = 2 tiles/device: device 0 reads tiles 0 and 1 (allowed), tile 2 raises. Just confirms that the owned-slice math holds across a multi-tile boundary, not just at a single tile.
> 
> 3. Column / last-dim shard (matmul-style)
> 
> Access tests shard dim 0 (rows). Matmul shards the contraction/output dim, i.e. the last dim — `ShardTensorToMesh(dim=1)` on a 3D grid. I verified that a device owning `cols[64:128)` and reading a `cols[0:32)` tile raises an exception, so just adding coverage (not a bug).
> 
> 4. Replicated + sharded in one kernel
> 
> Real data-parallel: replicated weights/bias (every device reads the whole tensor) alongside a sharded input (each device reads only its shard). `ReplicateTensorToMesh` sets no `mesh_shard_info` so it's skipped. I also verified that the replicated read by a foreign device is allowed, so not a bug, again just better coverage.

I added some tests along those lines

PR #742, opened 2026-07-14
  [ttkernel] Add ttkernel-specialize-cores per-core specialization pass

jackzhangTT:
### Problem description
A kernel's core coordinate (ttl.node) is only known per-core at runtime (it lowers to the hardware reads ttkernel.my_logical_x_ / my_logical_y_), so kernels cannot constant-fold control flow that branches on it. By cloning such a kernel per launch coordinate and replacing the coordinate reads with per-core constants, the branch conditions become constant and upstream passes like CSE can be leveraged  (dead code elimination/constant folding can go brrrrrr).

The specialized clones also need a runtime path that can dispatch more than one reader/writer binary: each clone must keep its NOC role and launch only on its own cores. NOC kernels roles need to be updated to be determined via `noc_kernel_idx` rather than position.

### What's changed

- Add an opt-in ModuleOp pass that runs at the TTKernel level, right before EmitC. For every kernel whose control flow branches on a core coordinate (an scf.if whose condition derives from ttkernel.my_logical_x_ / my_logical_y_), the pass clones the function per launch coordinate to leverage canonicalize/cse. 
- Wire --ttl-specialize-cores through the C++ / Python / ME2E pipelines (disabled by default).
- Runtime bridge (_compile_ttnn_kernel): assign Reader/Writer from ttl.noc_index instead of using positional counter, and dispatch each clone with a per-kernel CoreRangeSet from ttl.core_coord / ttl.crta_indices
- Emitted runner: bake per-kernel core ranges and NOC roles into the generated template so standalone cold runs match in-process dispatch.

NOTE: This pass is disable by default as I wasn't sure if we wanted it to be part of the main pipeline yet or not

**TODOs:**

-  Currently the pass clones the function once per launch coordinate but we should consider leveraging LaunchDomainAnalysis from TTL to further minimize clones, or do further analysis so clones are not duplicate

- Per-core compile-time / runtime arg specialization and inter-function `SymbolRefAttr` fixups are not performed.

Tests:
- Lit coverage for per-core cloning, dead-branch folding, the data-only no-clone case, and pipe participants.
- End-to-end add/matmul (coordinate used only for addressing, so no clone) and a column-swap reader that branches on core_x (cloned per core, matches a torch reference and the unspecialized path).

### Ticket
[Link](https://github.com/tenstorrent/tt-lang-ops-and-models/issues/6)

### DIFF
--- a/include/ttlang/Dialect/TTL/Passes.td
+++ b/include/ttlang/Dialect/TTL/Passes.td
@@ -848,8 +848,9 @@ def TTKernelSpecializeCores
     branch conditions and delete untaken regions. The runtime bridge reads
     ttl.core_coord to build per-kernel core ranges for dispatch.
 
-    NOTE: functions with symbol uses (e.g. `func.call` targets) are skipped so erasing the
-    original does not leave dangling `SymbolRefAttr`s.
+    NOTE: functions with symbol uses (e.g. `func.call` targets) are left
+    unspecialized with a warning so erasing the original does not leave dangling
+    `SymbolRefAttr`s; unrelated functions in the module are still specialized.
 
     Example (`ttl.launch_grid = [1, 2]`):
     ```mlir
--- a/include/ttlang/Dialect/TTL/Passes.td
+++ b/include/ttlang/Dialect/TTL/Passes.td
@@ -837,10 +837,65 @@ def TTKernelSpecializeCores
     : Pass<"ttkernel-specialize-cores", "::mlir::ModuleOp"> {
   let summary = "Clone kernels per core when control flow branches on a coordinate";
   let description = [{
-    Clone and rewrite the graph with ttl.node values propagated as constants
-    to leverage DCE and other upstream canolicalization passes. To minimize
-    clones we only clone kernels whose control flow branches on a core
-    coordinate. Clone functions once per coordinate.
+    Per-core specialization for TTKernel functions. Requires the 
+    `ttl.launch_grid` attribute, missing or malformed `ttl.launch_grid` is a hard error.
+
+    For each kernel whose scf.if region branches on a core coordinate, clones
+    the function once per launch coordinate, replaces
+    ttkernel.my_logical_x_ / ttkernel.my_logical_y_ with arith.constant's
+    for that core, and tags the cloned function with ttl.core_coord (the coordinate it
+    serves). Downstream canonicalize / cse then fold the now-constant
+    branch conditions and delete untaken regions. The runtime bridge reads
+    ttl.core_coord to build per-kernel core ranges for dispatch.
+
+    NOTE: functions with symbol uses (e.g. `func.call` targets) are skipped so erasing the
+    original does not leave dangling `SymbolRefAttr`s.
+
+    Example (`ttl.launch_grid = [1, 2]`):
+    ```mlir
+    // Before:
+    func.func @k() -> index {
+      %c0 = arith.constant 0 : index
+      %c7 = arith.constant 7 : index
+      %c9 = arith.constant 9 : index
+      %y = ttkernel.my_logical_y_ : () -> index
+      %pred = arith.cmpi eq, %y, %c0 : index
+      %r = scf.if %pred -> (index) {
+        scf.yield %c7 : index
+      } else {
+        scf.yield %c9 : index
+      }
+      return %r : index
+    }
+
+    // After:
+    func.func @k_c0_0() -> index attributes {ttl.core_coord = [[0, 0]]} {
+      %c0 = arith.constant 0 : index
+      %c7 = arith.constant 7 : index
+      %c9 = arith.constant 9 : index
+      %y = arith.constant 0 : index
+      %pred = arith.cmpi eq, %y, %c0 : index
+      %r = scf.if %pred -> (index) {
+        scf.yield %c7 : index
+      } else {
+        scf.yield %c9 : index
+      }
+      return %r : index
+    }
+    func.func @k_c0_1() -> index attributes {ttl.core_coord = [[0, 1]]} {
+      %c0 = arith.constant 0 : index
+      %c7 = arith.constant 7 : index
+      %c9 = arith.constant 9 : index
+      %y = arith.constant 1 : index
+      %pred = arith.cmpi eq, %y, %c0 : index
+      %r = scf.if %pred -> (index) {
+        scf.yield %c7 : index
+      } else {
+        scf.yield %c9 : index
+      }
+      return %r : index
+    }
+    ```
   }];
 
   let dependentDialects = [
--- a/include/ttlang/Dialect/TTL/Passes.td
+++ b/include/ttlang/Dialect/TTL/Passes.td
@@ -837,7 +837,7 @@ def TTKernelSpecializeCores
     : Pass<"ttkernel-specialize-cores", "::mlir::ModuleOp"> {
   let summary = "Clone kernels per core when control flow branches on a coordinate";
   let description = [{
-    Per-core specialization for TTKernel functions. Requires the 
+    Per-core specialization for TTKernel functions. Requires the
     `ttl.launch_grid` attribute, missing or malformed `ttl.launch_grid` is a hard error.
 
     For each kernel whose scf.if region branches on a core coordinate, clones
--- a/include/ttlang/Dialect/TTL/Passes.td
+++ b/include/ttlang/Dialect/TTL/Passes.td
@@ -833,6 +833,22 @@ def TTKernelInsertInits
   ];
 }
 
+def TTKernelSpecializeCores
+    : Pass<"ttkernel-specialize-cores", "::mlir::ModuleOp"> {
+  let summary = "Clone kernels per core when control flow branches on a coordinate";
+  let description = [{
+    Clone and rewrite the graph with ttl.node values propagated as constants
+    to leverage DCE and other upstream canolicalization passes. To minimize
+    clones we only clone kernels whose control flow branches on a core
+    coordinate. Clone functions once per coordinate.
+  }];
+
+  let dependentDialects = [
+    "::mlir::arith::ArithDialect",
+    "::mlir::func::FuncDialect"
+  ];
+}
+
 def TTKernelCombinePackTiles
     : Pass<"ttkernel-combine-pack-tiles", "::mlir::func::FuncOp"> {
   let summary = "Combine consecutive pack_tile ops into pack_tile_block";
--- a/lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
+++ b/lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
@@ -1,0 +1,179 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//===----------------------------------------------------------------------===//
+// TTKernelSpecializeCores (per-core specialization)
+//
+// A single module pass that runs at the TTKernel level, right before EmitC
+// conversion. For every kernel function whose control flow branches on a core
+// coordinate (i.e. an `scf.if` whose condition is derived from
+// `ttkernel.my_logical_x_` / `ttkernel.my_logical_y_`), the pass clones the
+// function once per launch coordinate. In each clone, the coordinate reads are
+// replaced by `arith.constant`s for that core, so the following
+// `canonicalize` / `cse` fold the now-constant branch conditions and delete the
+// untaken regions. Each clone is tagged with a `ttl.core_coord` attribute (the
+// coordinate it serves) and the runtime bridge (ttl_api.py) turns that into a
+// per-kernel core range for dispatch.
+//===----------------------------------------------------------------------===//
+
+#include "ttlang/Dialect/TTKernel/IR/TTKernel.h"
+#include "ttlang/Dialect/TTKernel/IR/TTKernelOps.h"
+#include "ttlang/Dialect/TTL/Passes.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/Twine.h"
+
+namespace ttk = mlir::tt::ttkernel;
+
+namespace mlir::tt::ttl {
+
+#define GEN_PASS_DEF_TTKERNELSPECIALIZECORES
+#include "ttlang/Dialect/TTL/Passes.h.inc"
+
+namespace {
+
+// Attribute names. These are part of the frontend / runtime contract and keep
+// the `ttl.` prefix even though this pass runs at the TTKernel level:
+// `ttl.launch_grid` (the launch extent) is set on the module by the Python
+// frontend, and `ttl.core_coord` is read back by the ttnn runtime bridge for
+// dispatch.
+constexpr llvm::StringLiteral LaunchGridAttrName = "ttl.launch_grid";
+constexpr llvm::StringLiteral CoreCoordAttrName = "ttl.core_coord";
+
+/// Parse the launch extent from an i64 array attribute into (gridX, gridY).
+///
+/// NOTE: operations.py specifies that only dims=2 is supported for now.
+///       this should be updated once operations.py is updated
+static bool readGrid(ArrayAttr attr, int64_t &gridX, int64_t &gridY) {
+  if (!attr || attr.size() != 2) {
+    return false;
+  }
+  auto x = llvm::dyn_cast<IntegerAttr>(attr[0]);
+  auto y = llvm::dyn_cast<IntegerAttr>(attr[1]);
+  if (!x || !y) {
+    return false;
+  }
+  gridX = x.getInt();
+  gridY = y.getInt();
+  return gridX > 0 && gridY > 0;
+}
+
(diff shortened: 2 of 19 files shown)


python/ttl/ttl_api.py
```diff
@@ -908,13 +1042,21 @@ def _compile_ttnn_kernel(
         else:
             kernel_arg_specs.append([])
 
+    # On the specialization path get_ttkernel_names returns 3*N clones, so the
+    # positional thread_tensor_indices (one entry per original thread) no longer
+    # lines up; use the per-clone indices recovered from ttl.crta_indices.
+    kernel_tensor_indices = (
```
brnorris03, line 1040: This alignment fix isn't propagated to the `TTLANG_EMIT_RUNNER` block further down, which still does `tensor_indices = thread_tensor_indices[kernel_idx]` and omits `core_ranges=`.

Confirmed on device (n150, grid 2x2, `TTLANG_EMIT_RUNNER=... pytest test/python/test_specialize_cores.py::test_specialize_cores_branch_matches_reference`):

```
>               tensor_indices = thread_tensor_indices[kernel_idx]
E               IndexError: list index out of range
python/ttl/ttl_api.py:1079: IndexError
```

On the specialization path `kernel_paths` has one entry per clone (here 6: 1 compute + 4 reader clones + 1 writer) while `thread_tensor_indices` has one per original thread (`[[], [0], [1]]`), so it runs off the end.

Reusing the values already computed here fixes the call site, and the test then passes:

```python
spec = KernelSpec(
    path=kernel_path,
    thread_type=thread_type,
    tensor_indices=kernel_tensor_indices[kernel_idx],
    config=kernel_configs[kernel_idx],
    core_ranges=kernel_core_ranges[kernel_idx],
)
```

I verified this doesn't regress the default path (`test_emit_runner.py` still passes). But it isn't sufficient on its own: the emitted runner template still can't express per-clone dispatch, and executing it then dies with `TT_FATAL: Illegal NOC usage ... results in hangs` (details in my comment on `kernel_runner.py`). So the two need to land together, or EMIT_RUNNER + specialization should raise a clear error until the template is updated.

`test_emit_runner.py` never sets `--ttl-specialize-cores`, which is why CI is green -- worth adding that combination as the regression test.

python/ttl/ttl_api.py
```diff
@@ -889,17 +999,41 @@ def _compile_ttnn_kernel(
             thread_to_kernel["TRISC_1"] = name
             thread_to_kernel["TRISC_2"] = name
         elif thread_type == "noc":
-            if noc_kernel_idx == 0:
+            noc_role = _get_kernel_noc_index(module, name)
+            if noc_role is None:
```
brnorris03: This looks unreachable. `_compile_ttnn_kernel` has a single caller, and the frontend sets `ttl.noc_index` on every datamovement thread, so `_get_kernel_noc_index` never returns `None` in this branch. `noc_role_per_core` is dead, too.

Also should probably mention in the PR description (which becomes the commit comment): this switches the default path from the positional `noc_kernel_idx` to attribute-based role assignment. The two agree now only because the frontend assigns `ttl.noc_index` with the same positional counter.

python/ttl/ttl_api.py
```diff
@@ -747,6 +754,70 @@ def _get_kernel_i32_array_attr(module, kernel_name: str, attr_name: str):
     return list(attr)
 
 
+def _get_kernel_core_coords(module, kernel_name: str):
+    """Read the `ttl.core_coord` attribute set by `ttkernel-specialize-cores`.
+
+    We expect the aray to be of length 2 since node dim currently only supports 2D.
```
brnorris03: Typo: `aray` -> `array`.

python/ttl/ttl_api.py
```diff
@@ -747,6 +754,70 @@ def _get_kernel_i32_array_attr(module, kernel_name: str, attr_name: str):
...
+    if attr is None:
+        return None
+    return int(IntegerAttr(attr).value)
+
+
+def _get_kernel_crta_indices(module, kernel_name: str):
+    """Read the `ttl.crta_indices` attribute as a list of global tensor indices.
+
+    Used by the per-core specialization path where clones cannot be aligned
+    positionally with the original thread list. Returns an empty list when the
```
brnorris03: The docstring says it returns an empty list when the attribute is missing, but the code raises `ValueError`. Raising is correct, just need to fix the docstring.

lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
```diff
@@ -0,0 +1,179 @@
...
+  moduleBuilder.insert(clone);
+}
+
+struct TTKernelSpecializeCoresPass
+    : impl::TTKernelSpecializeCoresBase<TTKernelSpecializeCoresPass> {
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+
+    int64_t gridX = 0, gridY = 0;
+    if (!readGrid(module->getAttrOfType<ArrayAttr>(LaunchGridAttrName), gridX,
```
brnorris03: Silent no-op: the user explicitly asked for this pass via `--ttl-specialize-cores`, so a missing or malformed `ttl.launch_grid` should produce a diagnostic rather than quietly returning an unspecialized binary.

This is `runOnOperation`, so `module.emitOpError() << ...; signalPassFailure();` is available. Worth distinguishing "attribute absent or malformed" (a hard error when the pass was requested) from `gridX * gridY <= 1` (a legitimate no-op).

lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    Builder builder(&getContext());
+    for (func::FuncOp func : targets) {
+      OpBuilder moduleBuilder(func);
+      for (int64_t y = 0; y < gridY; ++y) {
+        for (int64_t x = 0; x < gridX; ++x) {
+          emitCoreClone(func, x, y, moduleBuilder, builder);
+        }
+      }
+      func.erase();
```
brnorris03, line 195: The PR notes that inter-function `SymbolRefAttr` fixups aren't performed, but nothing enforces that precondition. Erasing the original while creating renamed clones leaves any `func.call` dangling, which surfaces as a confusing downstream verifier error instead of a targeted one.

Suggest checking before any mutation:

```cpp
if (auto uses = SymbolTable::getSymbolUses(func, module); uses && !uses->empty()) {
  func.emitOpError() << "cannot specialize a function with symbol uses";
  signalPassFailure();
  return;
}
```

This also covers the currently unchecked `_c<x>_<y>` symbol-collision case.

lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
```diff
@@ -0,0 +1,179 @@
...
+// frontend, and `ttl.core_coord` is read back by the ttnn runtime bridge for
+// dispatch.
+constexpr llvm::StringLiteral LaunchGridAttrName = "ttl.launch_grid";
+constexpr llvm::StringLiteral CoreCoordAttrName = "ttl.core_coord";
+
+/// Parse the launch extent from an i64 array attribute into (gridX, gridY).
+///
+/// NOTE: operations.py specifies that only dims=2 is supported for now.
+///       this should be updated once operations.py is updated
+static bool readGrid(ArrayAttr attr, int64_t &gridX, int64_t &gridY) {
```
brnorris03: `bool` + out-params is a sentinel-style failure. Prefer `FailureOr<std::pair<int64_t, int64_t>>` (or `std::optional`) to match the error-handling conventions.

lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
```diff
@@ -0,0 +1,179 @@
...
+  });
+  return found;
+}
+
+/// Emit one clone of `func` for core `(x, y)`, replacing every coordinate read
+/// with the matching constant and tagging the clone with `ttl.core_coord`.
+/// TODO: See if we can leverage LaunchDomainAnalysis in an earlier pass
+/// to further minimize clones.
+static void emitCoreClone(func::FuncOp func, int64_t x, int64_t y,
+                          OpBuilder &moduleBuilder, Builder &builder) {
```
brnorris03: `OpBuilder` derives from `Builder`, so `builder` is redundant — `moduleBuilder` can both construct the attributes and insert.

The `xReads` / `yReads` collect-then-replace loops below are also near-identical; a small lambda templated on the op type collapses them into one.

include/ttlang/Dialect/TTL/Passes.td
```diff
@@ -833,6 +833,22 @@ def TTKernelInsertInits
   ];
 }
 
+def TTKernelSpecializeCores
+    : Pass<"ttkernel-specialize-cores", "::mlir::ModuleOp"> {
+  let summary = "Clone kernels per core when control flow branches on a coordinate";
+  let description = [{
```
brnorris03, line 839: Typo: `canolicalization` -> `canonicalization`.

The description should mention that it requires `ttl.launch_grid` on the module, and it produces `ttl.core_coord` on each clone, which the runtime depends on for dispatch. Also include before/after mlir snippets -- similar to upstream pass descriptions.

Also should probably mention that only `scf.if` conditions are inspected, so `scf.index_switch`, `cf.cond_br`, and any other specific constraints.

test/python/test_specialize_cores.py
```diff
@@ -0,0 +1,343 @@
...
+tagging each clone with `ttl.core_coord`. The runtime bridge
+(`_compile_ttnn_kernel`) turns each `ttl.core_coord` into a per-kernel core
+range for dispatch.
+
+Crucially, a kernel is cloned only when it *branches* on a core coordinate
+(`scf.if` on `ttkernel.my_logical_x_` / `my_logical_y_`). Kernels that use the
+coordinate only as data (e.g. addressing `a[y, x]`) are left as a single
+whole-grid binary, because the runtime coordinate reads already give each core
+its own coordinate -- no clone is needed. Modules that use pipes are never
+specialized (cloning a pipe endpoint deadlocks at runtime).
```
brnorris03: This contradicts both the pass and the lit test, which says the pass no longer special-cases pipes — and the pass indeed has no pipe handling.

If cloning a pipe endpoint really can deadlock, then the lit test is codifying unsafe behavior as correct and the pass needs a guard.Why is cloning a pipe endpoint a problem (or does this mean replicating -- so you have larger number of receives, for example, than sends)?

Could you confirm and drop it? A claimed deadlock hazard left unresolved in a docstring is worse than no comment IMO.

test/python/test_specialize_cores.py
```diff
@@ -0,0 +1,343 @@
...
+  * The specialized result matches a torch reference (numerical correctness).
+  * The specialized result matches the default (unspecialized) path.
+  * The dumped final MLIR shows these data-addressing kernels are NOT cloned
+    (no `ttl.core_coord`), i.e. specialization safely declines to clone them.
+
+A third kernel (`branch_swap_*`) *does* branch on the coordinate: its reader
+selects a different source column depending on `core_x`, so specialization
+clones the reader per core and const-folds the branch. That test exercises the
+full clone-and-dispatch path end to end (the MLIR-only clone/fold checks live in
+the lit test test/ttlang/Dialect/TTL/Transforms/specialize_cores.mlir).
```
brnorris03: Stale path: the lit test is in `test/ttlang/Dialect/TTKernel/Transforms/`, not `Dialect/TTL/Transforms/`.

test/python/test_specialize_cores.py
```diff
@@ -0,0 +1,343 @@
...
+        x, y = ttl.node(dims=2)
+        with out_dfb.wait() as o:
+            tx = ttl.copy(o, out[y, x])
+            tx.wait()
+
+
+# Separate op object so its compilation cache is independent from the default
+# matmul. This one is always invoked with options="--ttl-specialize-cores".
+@ttl.operation(grid=(GRID_X, GRID_Y))
+def matmul_specialized(a, b, out):
```
brnorris03: This body is byte-identical to `matmul_default` (same for the `branch_swap` pair). The comment explains why the op *objects* must be distinct but don't have to duplicate the bodies, suggest a helper:

```python
def _make_matmul_op():
    @ttl.operation(grid=(GRID_X, GRID_Y))
    def matmul(a, b, out):
        ...
    return matmul

matmul_default = _make_matmul_op()
matmul_specialized = _make_matmul_op()
```

python/ttl/compiler_options.py
```diff
@@ -90,6 +90,15 @@ def _make_parser() -> argparse.ArgumentParser:
         action=argparse.BooleanOptionalAction,
         help="Insert compiler-allocated intermediate DFBs for fused computations (default: enabled).",
     )
+    p.add_argument(
+        "--ttl-specialize-cores",
```
brnorris03, line 94: `docs/sphinx/reference/compiler-options.md` needs updating as well — it documents every other option across the codegen table, the pipeline options table, the pipeline step list, and the pass reference. `Passes.td` and `compiler_options.py` are covered by this PR; that file isn't.

python/ttl/compiler_options.py
```diff
@@ -90,6 +90,15 @@ def _make_parser() -> argparse.ArgumentParser:
         action=argparse.BooleanOptionalAction,
         help="Insert compiler-allocated intermediate DFBs for fused computations (default: enabled).",
     )
+    p.add_argument(
```
zoecarver, line 93: I wonder if we should make the default on here? At least for testing maybe?

test/python/test_specialize_cores.py
```diff
@@ -0,0 +1,343 @@
...
+    The kernels here use the coordinate only for addressing, never for a branch,
+    so specialization must decline to clone them: no `ttl.core_coord` clone tag
+    and no `_c<x>_<y>` clone suffixes should appear.
+    """
+    with open(final_mlir_path) as fd:
+        mlir = fd.read()
+    assert (
+        "ttl.core_coord" not in mlir
+    ), "data-addressing kernels must not be cloned (found ttl.core_coord)"
+    clones = re.findall(r"func\.func @\w+_c\d+_\d+", mlir)
```
zoecarver, line 67: This is not usually how we do tests. Would you mind converting this to a lit test instead? I think it's slightly easier to verify just by looking at the CHECKs.

zoecarver: Didn't read far enough, sorry. I see there's a lit test below, so up to you if you want to keep this one as well.

test/ttlang/Dialect/TTKernel/Transforms/specialize_cores.mlir
```diff
@@ -0,0 +1,152 @@
+// RUN: ttlang-opt %s --split-input-file -pass-pipeline='builtin.module(ttkernel-specialize-cores)' | FileCheck %s
```
zoecarver, line 1: Maybe an end-to-end lit test too (a python file that checks generated cxx)?

zoecarver (approved):
LGTM!

python/ttl/kernel_runner.py
```diff
@@ -193,9 +197,15 @@ def build_kernel_descriptors(
         else:
             kernel_compile_time_args = cb_indices + list(tensor_accessor_args)
 
+        # Note: core_range is used outside this function to dispatch the kernel to the correct cores
+        # and should not be overridden
+        kernel_ranges = (
+            spec.core_ranges if spec.core_ranges is not None else core_ranges
+        )
```
brnorris03: Nit: this comment seems to contradict the code right below it -- we're overwriting the per-kernel core ranges here?

Bigger issue: the emitted-runner template can't express per-clone dispatch. It re-derives reader/writer from a positional `noc_idx` and passes only the whole-grid `core_ranges`, so the reader clones get labeled as writers and every clone lands on every core.

Confirmed on device: on a cold program cache the emitted runner dies with `TT_FATAL: Illegal NOC usage`. It only works if the same program already ran in-process -- the runner's `custom_program_hash` reuses the cached (correct) program, which masks it. So a test has to run it cold (emit under `TTLANG_COMPILE_ONLY=1`).

Fix: emit per-kernel core ranges + NOC roles (`ttl.noc_index`) instead of the positional counter. Worth a test.

python/ttl/ttl_api.py
```diff
@@ -1648,6 +1790,14 @@ def _compile_kernel(
             "cse",
             "lower-affine",
             "ttl-lower-signpost-to-emitc",
+        ]
+        if compiler_options.specialize_cores:
+            pipeline_passes += [
+                "ttkernel-specialize-cores",
+                "canonicalize",
+                "cse",
+            ]
```
brnorris03, line 1875: There are three pipeline updates required with any new passes the cpp one (used by ttlang-opt tool), the me2e tests and this one, so please make sure they are all equivalently updated. Also any new options should be added to the sphinx docs.

lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
jackzhangTT: I agree with the check however I don't think we should fail the pass. I think the current func should be skipped if it's used else where in the module and handle that case in the future.

brnorris03: Agreed!

test/python/test_specialize_cores.py
jackzhangTT: sorry I forgot to update this comment it's left over from the original two pass implementation

python/ttl/ttl_api.py
jackzhangTT: Just a quick update it seems like the ttl.noc_index isn't actually propagated at the ttk level so when I removed the none branch it actually broke the tests. But I think it's cleaner to read the attribute to figure out whether it's a read/writer than to determine it positionally so I've propagated it through.

brnorris03 (approved):
Looks good, thanks for all the updates! Optionally I thought maybe could add a test that runs the new pass twice (to check idempotency); can be mlir lit test.

PR #743, opened 2026-07-14
[spec] Externalize examples

phizalev-TT:
## Summary

The spec's code examples were previously inlined as Markdown code fences in `TTLangSpecification.md`, so they couldn't be run or tested. This PR **externalizes** them into standalone files under `examples/spec/` (grouped by spec section) and adds a generator that stitches them back into the rendered spec.

Most examples are illustrative snippets carried over verbatim and are not meant to run standalone. Four are complete, runnable programs that open a device, execute, and assert against a Torch reference while still rendering as concise snippets: `dataflow_buffer/dataflow_buffer.py`, `dataflow_buffer/tiled_tensor_shape.py`, `dataflow_buffer/row_major_tensor_shape.py`, and `operation_function/multi_kernel_operation.py`.

## How it works

- **Authoring source** — `docs/sphinx/specs/TTLangSpecification.externalized.md` is the file humans edit. Code examples are replaced by directives naming a file under `examples/spec/`:
  ```
  <!-- @spec:example pipe/gather.py -->
  ```
- **Markers** — the region embedded from each file is delimited by `# spec:begin` / `# spec:end` comments and dedented. A file may contain several such sections; they are concatenated (no separating line) and dedented as a whole. This keeps the runnable examples real programs while showing only the illustrative parts: scaffolding (device open/close, `try`/`finally`, concrete `ttl.copy`/`store` calls adjacent to the `# ...` placeholders, and the final `assert`s) lives *outside* the markers, so it runs but never appears in the spec.
- **Generator** — `docs/sphinx/specs/build_spec.py` expands directives into `py` code blocks and writes `TTLangSpecification.md`:
  - `python docs/sphinx/specs/build_spec.py` — regenerate
  - `python docs/sphinx/specs/build_spec.py --check` — verify up to date (non-zero if stale; for CI)

  Malformed markers (unclosed `spec:begin`, stray `spec:end`, nested `spec:begin`) are reported rather than silently mis-rendered.

## Files

- **`examples/spec/<section>/*.py`** (19 files) across `operation_function/`, `grid/`, `dataflow_buffer/`, `block/`, `pipe/`, `tensor_slice/`, `copy/`, `semaphore/`, and `performance_and_debugging/`. The four named above are runnable programs; the rest are illustrative snippets not intended to run standalone.
- **`docs/sphinx/conf.py`** — added `specs/TTLangSpecification.externalized.md` to `exclude_patterns` so Sphinx renders only the generated `TTLangSpecification.md`, not the authoring source with unexpanded directives.

## Notes

- Most examples render identically to before (modulo Python formatting) — a pure refactor of *how* the spec is authored; `dataflow_buffer/dataflow_buffer.py` renders unchanged despite now being runnable.
- Three render with small, intentional changes: the two shape examples show executable `assert … == …` in place of the old `# prints …` comments, and the multi-kernel example shows `@ttl.operation(grid=(1, 1))` plus its device-scoped `try`/`finally` wrapper.


docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Generate TTLangSpecification.md from its template and externalized examples.
+
+The specification is authored as ``TTLangSpecification.externalized.md``, a
+Markdown file whose code examples are replaced by include directives of the
+form::
+
+    <!-- @spec:example <file> -->
+    <!-- @spec:example <file>:<tag> -->
+
```
brnorris03, line 11: Claude suggested adding this to `.pre-commit-config.yaml` (right before `  - repo: https://github.com/espressif/check-copyright/`, so the spec is updated automatically (precommit is also already called in CI).
```
  - repo: local
    hooks:
      - id: build-spec
        name: regenerate TTLangSpecification.md from externalized examples
        entry: python3 docs/sphinx/specs/build_spec.py
        language: system
        pass_filenames: false
        files: ^(docs/sphinx/specs/TTLangSpecification\.externalized\.md|examples/spec/.*)$
```

docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+    return textwrap.dedent("".join(lines[start + 1 : end_idx]))
+
+
+def render(template_text: str) -> str:
+    """Expand every include directive in ``template_text`` into a code block."""
+    out: list[str] = []
+    for line in template_text.splitlines(keepends=True):
+        match = DIRECTIVE.match(line.rstrip("\n"))
+        if match is None:
+            out.append(line)
```
brnorris03, line 109: A malformed directive (e.g., if it contains a typo) does not result in an error now.

I changed one directive to `@spec:exmaple` (typo) and regenerated: the generator exited **0**, and the line passed straight through into the generated `.md` verbatim as an HTML comment -- which is invisible when rendered. The example just disappears from the spec, with no error anywhere. The same applies to a directive that is indented, since `DIRECTIVE` is `^`-anchored.

```suggestion
        if match is None:
            if "@spec:" in line:
                raise SystemExit(f"Malformed directive: {line.strip()}")
            out.append(line)
```

docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+        stripped = line.strip()
+        if stripped == begin and start is None:
+            start = i
+        elif stripped == end and start is not None:
+            end_idx = i
+            break
+
+    if start is None or end_idx is None:
+        marker = "spec:begin/spec:end" if tag is None else f"spec:begin/end '{tag}'"
+        raise SystemExit(f"Could not find {marker} markers")
```
brnorris03: I'd add the file name to the error message, e.g., append "in <fname>"

docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+            out.append(line)
+            continue
+
+        example = EXAMPLES_DIR / match["file"]
+        if not example.is_file():
+            raise SystemExit(f"Example file not found: {example}")
+
+        region = extract_region(example.read_text(encoding="utf-8"), match["tag"])
+        if not region.endswith("\n"):
+            region += "\n"
```
brnorris03: I'd check for an empty region and emit an error

docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+import textwrap
+from pathlib import Path
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+REPO_ROOT = SCRIPT_DIR.parents[2]
```
brnorris03: This will break if the file happens to be moved, consider not using a specific parent count by doing something like this

```python
def _repo_root() -> Path:
    """Nearest ancestor containing pyproject.toml."""
    for candidate in Path(__file__).resolve().parents:
        if (candidate / "pyproject.toml").is_file():
            return candidate
    raise SystemExit("build_spec.py: no pyproject.toml found in any ancestor")
```

examples/spec/pipe/gather.py
```diff
@@ -0,0 +1,86 @@
...
+# ---------------------
+# gather from row y to (0, y) with unicast.
+#
+# The pipe net is sized from the active set, not the launch extent.
+# ROWS and COLS describe the rectangle that bounds the active set.
+# Nodes outside the active rectangle (row 0..ROWS-1, column 0..COLS-1)
+# skip the operation body.
+
+ROWS = ...  # rows participating in the gather
+COLS = ...  # columns participating in the gather
```
brnorris03, line 43: Are these values supposed to be present? I thought the *.py should be runnable standalone, but most are not complete right now, so cannot run. Maybe I missed something in the transformations.

Only 2 of 17 import against the real `ttl` stack: `ROWS`/`COLS` are module-level `Ellipsis`, so `range(1, COLS)` on line 46 raises `TypeError`, and the rest fail on undefined `M`/`N`/`device`/`dfb`. That leaves the `.py` vs `.py.txt` split not meaning much -- the `.py` files are as un-runnable as the two `.py.txt` ones.

The dedent in `build_spec.py` looks like the intended answer, but nothing uses it: all 19 regions start at column 0. `examples/eltwise_add.py` nests its body in `@ttl.operation` and imports fine with no device -- doing the same here would make the file importable and leave the rendered spec unchanged.

If these are meant to stay illustrative instead, dropping `.py.txt` and the unused dedent support would say so more clearly.

examples/spec/pipe/gather.py
```diff
@@ -0,0 +1,86 @@
...
+#
+# Everything outside the markers (imports, scaffolding) exists so the file can
+# stand on its own and is not copied into the specification.
+
+import math
+
+import torch
+
+import ttl
+import ttnn
```
brnorris03, line 20: optional (minor): `math`, `torch`, and `ttnn` are unused here, so can be removed (only `ttl` is used).

All 17 example files carry this same copy-pasted import header, and in most of them all four imports are unused -- `dataflow_buffer/row_major_tensor_shape.py` does not even use `ttl`. The header comment above (line 13) says imports exist "so the file can stand on its own", which is fine, but not quite accurate for imports that are unused.

examples/spec/pipe/forward_neighbor.py
```diff
@@ -0,0 +1,80 @@
...
+import torch
+
+import ttl
+import ttnn
+
+# spec:begin
+# ---------------------
+# forward to a +1 neighbor in a column x
+
+grid_x, grid_y = ttl.grid_size()
```
brnorris03, line 26: **[info -- not a regression, no action needed in this PR]**

`ttl.grid_size()` has signature `grid_size(*, dims)` with `dims` required, so this line raises `TypeError` and the same call appears in three published spec regions.

This is pre-existing, not introduced here: `git show origin/main:docs/sphinx/specs/TTLangSpecification.md` has the identical no-arg call at lines 847/891/939, so the PR carried the content over faithfully.

Flagging it because it is the best argument *for* this PR: it is exactly the spec-vs-compiler drift the machinery here would catch, the moment the examples import and the generator runs in CI.

examples/spec/operation_function/multi_kernel_operation.py
brnorris03, line 1: nit: even though this has the .txt suffix, it's markdown, so maybe should be .md?

docs/sphinx/specs/build_spec.py
```diff
@@ -0,0 +1,131 @@
...
+marker comments::
+
+    # spec:begin
+    ...code included in the specification...
+    # spec:end
+
+A file may contain several named regions, selected by ``:<tag>`` on both the
+directive and the markers (``# spec:begin <tag>`` / ``# spec:end <tag>``). The
+embedded region is dedented, so an example may nest the marked lines inside a
+function while the specification shows them at the left margin.
```
brnorris03: **[medium] This documented capability is unused -- no example nests its region.**

I checked all 19 files under `examples/spec/`: every one has region lines starting at column 0, so `textwrap.dedent()` in `extract_region()` is a no-op in every case. Nothing nests marked lines inside a function.

That matters because this is precisely the mechanism that would make the examples importable, which is what the `.py` extension currently claims for them. The import failures (2 of 17 import; see the comment on `pipe/gather.py`) all trace to spec code executing at *module scope*.

`examples/eltwise_add.py` shows the convention that avoids it: the body sits inside `@ttl.operation`, which defers execution to call time -- I confirmed `eltwise_add.py` imports cleanly with no device attached. If the spec examples nested their regions inside `@ttl.operation` and marked only the inner lines, dedent would strip the wrapper indentation and the rendered spec would stay byte-identical, while the files became importable.

So this feature is doing real work in the design, just not in the examples. Either they should use it, or it is dead code and should be removed along with the `.py.txt` split.

brnorris03 (commented):
My main concern is the ability to run the examples, otherwise just a few minor things.

phizalev-TT:
> My main concern is the ability to run the examples, otherwise just a few minor things.

Agreed. I will get them to run section by section in the follow up PRs.

examples/spec/pipe/gather.py
phizalev-TT: This header part is a placeholder. I will minimize imports while making examples to run.

examples/spec/pipe/forward_neighbor.py
phizalev-TT: Good point. Will make separate change for this.

brnorris03 (approved):
LGTM, thank you for all the updates!

zoecarver (approved):
LGTM

arichinsTT (approved):
LGTM

FConstantinos (approved):
LGTM

PR #744, opened 2026-07-15
[uplift-A] Uplift llvm to latest main and tt-metal to v0.74.0 (PyPI-aligned)

brnorris03:
### Problem description

The toolchain needs to move onto the latest LLVM `main` tip, and tt-lang needs a tt-metal pin that matches the newest public `ttnn` wheel so the published wheel's `ttnn` dependency resolves on public PyPI. Before this change `TT_METAL_TAG` was `v0.74.0-dev20260629` while the public `ttnn` provenance tag was `v0.73.1-rc5`; those disagree on their `vX.Y.Z` component, so `ttnn_pypi_aligned` fails and `publish-pypi.yml` refuses the release, forcing S3-only wheels.

This is the first of two stacked uplifts. It lands the publishable pin; `[uplift-B]` then advances tt-metal to the latest tag, which is deliberately S3-only.

### What's changed

- Bump `third-party/llvm-project` to `4279d524cc78` (llvm/main) and `third-party/tt-metal` to `v0.74.0` (`c49bb7625e6a`).
- Set `TTNN_PYPI=0.74.0`, `TTNN_PYPI_TT_METAL_TAG=v0.74.0` and `TT_METAL_TAG=v0.74.0`. Public `ttnn` 0.74.0 is now on PyPI, and tt-metal publishes a plain `v0.74.0` tag pointing at the same commit as `v0.74.0-rc8`, so both tags reduce to `v0.74.0`, the alignment check passes, and the wheel's `ttnn==0.74.0` dependency is publishable to public PyPI.
- Drop the `DANGER` note warning that public publishing is blocked, which no longer applies at this pin. `[uplift-B]` restores it, where it is true again.
- Define `ComputeOp`'s new four-argument `TilingInterface::getTiledImplementation` overload. Upstream MLIR added an `InnerTileAlignment`-hint variant of that method, and because `TTLOps.td` lists `getTiledImplementation` by name, ODS now treats the op as overriding both overloads and dispatches to a definition that did not exist, failing the link. The override forwards to the existing hint-less implementation, matching the interface default, since `ttl.compute` does not consult pack/unpack inner-tile alignment.
- Document the two-phase uplift workflow, and correct four things the uplift procedure got wrong: toolchain reuse is keyed on file existence rather than submodule SHA (so rebuilding an uplift into a populated toolchain directory silently keeps the old LLVM and tt-metal, with only an `AUTHOR_WARNING`), the shallow `llvm-project` submodule needs an explicit `--depth 1` SHA fetch, uplift failures land in three distinct stages with different remedies, and `pyproject.toml` never changes on an uplift because its version and dependencies are dynamic.

### Testing

Built the LLVM and tt-metal toolchain from these pins into a separate prefix, leaving the default toolchain intact, then built tt-lang against it and ran the MLIR lit suite: `check-ttlang-mlir` passes with 171 of 172 tests and one expected failure. Both tt-metal patches apply cleanly at `v0.74.0`. The full hardware suite and the PyPI alignment gate run in CI on this PR.

### DIFF
--- a/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
+++ b/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
@@ -314,7 +314,9 @@ inline bool isFullFp32ReduceSupported(TileReduceOp reduceOp) {
     return false;
   }
 
-  // TODO(#533): Blackhole REDUCE_ROW full-fp32 produces incorrect results.
+  // Blackhole cannot write the low 16 bits while the 32-bit dest is enabled, so
+  // the row-reduce LLK disables fp32 accumulation internally (tt-metal #47311).
+  // Report unsupported so DST_ACCUM_MODE is not enabled for it.
   return !isBlackholeTarget(reduceOp) ||
          reduceOp.getReduceDim() != mlir::tt::ttkernel::ReduceDim::Row;
 }
--- a/.github/workflows/call-test-hardware.yml
+++ b/.github/workflows/call-test-hardware.yml
@@ -61,6 +61,7 @@ jobs:
       RUNS_ON: ${{ inputs.runs-on }}
       PYTHONDONTWRITEBYTECODE: 1
       TTLANG_TEST_SEED: 42
+      HW_PYTEST_CHIPS: 1
 
     steps:
       - name: Checkout tt-lang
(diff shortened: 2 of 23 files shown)


phizalev-TT (approved):

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -314,7 +314,9 @@ inline bool isFullFp32ReduceSupported(TileReduceOp reduceOp) {
     return false;
   }
 
-  // TODO(#533): Blackhole REDUCE_ROW full-fp32 produces incorrect results.
```
zoecarver, line 317: happy to see this resolved :)

zoecarver (approved):

PR #751, opened 2026-07-17
[ci] Fix manual PyPI publish condition

brnorris03:
### Problem description

Manual PyPI publishing with an existing Docker tag built and tested the wheel, but skipped the final publish job. The skipped `build-docker` job propagated through the dependency chain because terminal job conditions lacked a status-check function. The dry-run summary had the same defect.

### What's changed

Use `!cancelled()` to override transitive skip propagation while explicitly requiring each direct dependency to succeed. Add regression coverage for both terminal jobs.

### Checklist

- [x] Packaging workflow tests cover the changed conditions
- [x] `actionlint` and pre-commit pass

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,13 +11,13 @@
 * @tenstorrent/tt-lang-maintainers
 
 # Protect repository policy and automation configuration.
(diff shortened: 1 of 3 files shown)


phizalev-TT (approved):

jackzhangTT:
LGTM

PR #740, opened 2026-07-08
[ttl][nfc] Extract PipeTransportEmitter for PipeNet NoC lowering [pipes-4]

brnorris03:
PipeNet PR tree:

```text
main (includes #756, #768, #704, #782, #778, #775, and #783; all merged)
  -> [pipes-3] #765
    -> [pipes-4] #700
      -> [pipes-5] #740 (this PR)
        -> [pipes-6] #784
          |-> [pipes-7] #780
          |-> [pipes-8] #754
          `-> [pipes-fabric] #734
```

### Problem description

PipeNet protocol decisions and NoC emission were interleaved. Fabric transport needs to reuse the protocol proofs and resource decisions without duplicating NoC-specific lowering.

### What's changed

- Add `PipeTransportEmitter` as the transport emission interface.
- Move existing NoC address publication, sender-ready signaling, payload writes, barriers, and completion signaling into `NocPipeTransportEmitter`.
- Keep protocol selection, address proofs, synchronization, and resource decisions transport-independent.

This is a refactoring: generated intra-device behavior is unchanged.

### Validation

Updated regressions preserve payload writes, synchronization order, completion
increments, and capacity releases across the emitter extraction.

### Checklist

- [x] New and existing tests provide coverage for the changes.

PR #734, opened 2026-07-07
[ttl][pipes] Add multidevice fabric PipeNets [fabric-2]

brnorris03:
Stack: `main` -> **#734**

### Problem description

PipeNets previously had no device-level transfer relation or target binding, so they could not express topology-independent communication across a device mesh. Start review with [Pipes on Fabric](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipes-multidevice-integrated-poc/docs/development/PipesOnFabric.md), especially System overview, Design, and Software architecture. [PipeNets](https://github.com/tenstorrent/tt-lang/blob/bnorris/pipes-multidevice-integrated-poc/docs/development/PipeNets.md) defines the shared protocol and proofs.

### What's changed

- Add logical device domains and structured point-to-point, neighbor, gather, scatter, and all-to-all transfer relations.
- Lower graph PipeNets to compact selected-record loops with device-qualified launch, receiver-address, and synchronization proofs.
- Plan routes, links, runtime arguments, manager interference, and semaphore capacity before descriptor mutation.
- Preserve logical manager intervals for ownership and link assignment while opening each function's host-specialized connection records through one physical manager lifetime.
- Bind generated and external managers through one validated target plan, including emitted runners and program caching.
- Discover topology after selecting the fabric configuration and exercise the full fabric suite on Galaxy CI.

### Checklist

- [x] Current-head full-Galaxy fabric suite: 59 passed, including mesh and torus routing plus consecutive BF16 and FP32 tree all-reduce.
- [x] 128 pipe tests passed; one multi-device test skipped on the one-card runner and one strict xfail remains for #623.
- [x] 472 supported MLIR tests passed; 2 compile-time measurement tests were unsupported in this configuration.
- [x] Focused host, Docker build, simulator, kernel-runner, and packaging tests passed.
- [x] Pre-commit passed.

PR #756, opened 2026-07-19
[ttl][analysis] Add static execution analysis

brnorris03:
Stack: #756 (this PR) -> #700 -> #740 -> #734

### Problem description

Several compiler components need the exact number of times an operation executes within structured control flow. Implementing this reasoning inside PipeNet graph construction would couple a generic control-flow proof to one consumer and prevent reuse by other compiler analyses and cost models.

### What's changed

- Adds a shared `IntegerExpressionEvaluator` that uses MLIR folding semantics, consumer-provided facts, and memoized iterative evaluation to prove integer and index values without changing IR.
- Adds `ExecutionCountAnalysis`, which returns an exact unsigned 64-bit dynamic-instance count or unknown relative to one invocation of a specified root region.
- Uses `LoopLikeOpInterface` and `RegionBranchOpInterface` for structured control flow and models finite block control-flow graphs with MLIR dead-code analysis, LLVM strongly connected components, and post-dominance.
- Computes rectangular loop nests by multiplication and uses bounded iteration enumeration only when nested control flow depends on induction variables.
- Supports consumer-provided symbol and region-invocation evaluators for compile-time facts not encoded in IR.
- Reuses one structured-control proof for all operation queries in the same block.
- Uses sparse constant-propagation facts forwarded through block and region arguments when evaluating control-flow selectors.
- Uses checked arithmetic and returns unknown for overflow, unsupported control flow, cyclic execution, unresolved runtime selection, or exceeded proof limits.
- Refactors launch-node domain analysis to use the shared integer evaluator instead of maintaining separate arithmetic and predicate evaluators.
- Adds `ttlang-op-stats`, a test-only example client that reports static operation occurrences and aggregate dynamic instances, plus an end-to-end SDPA-style kernel test.
- Documents the analysis semantics, control-flow model, algorithm, extension points, and use in operation-type cost models.
- Adds an explicit build target for test-only lit tools used by hardware CI.

### Test coverage

- Integer and index evaluation covers constants, consumer facts, induction-variable bindings, standard operation folds, signed and unsigned semantics, mismatched bit widths, and shared expression graphs.
- Structured-control coverage includes straight-line sequences, nested and zero-trip loops, large rectangular nests, induction-dependent branches, selected and unreachable regions, and consumer-defined region invocation counts.
- Block-control-flow coverage includes unconditional sequences, compile-time and runtime branches, constants forwarded through block and region arguments, exact merge counts, unreachable and duplicate edges, reducible and irreducible cycles, non-exiting loops, and disconnected cycles.
- Failure coverage verifies unknown results for unresolved runtime data, unsupported operations and regions, operations outside the root region, cyclic execution, enumeration-limit exhaustion, and arithmetic overflow.
- Integration coverage exercises the public analysis API, deterministic operation statistics on MLIR, the SDPA-style Python kernel, and launch-node predicate evaluation through the shared integer evaluator.

### Checklist


zoecarver (approved):

arichinsTT (approved):

jackzhangTT (approved):
LGTM

PR #765, opened 2026-07-21
[ttl][pipes] Scale PipeNet synchronization counters [pipes-3]

brnorris03:
PipeNet PR tree:

```text
main (includes #756, #768, #704, #782, #778, #775, and #783; all merged)
  -> [pipes-3] #765 (this PR)
    -> [pipes-4] #700 (includes the independently reviewed #740 and #784 layers)
      |-> [pipes-7] #780
      |-> [pipes-8] #754
      `-> [pipes-fabric] #734
```

### Problem description

PipeNet completion and readiness counters could exhaust the 16 hardware-local semaphore IDs and fail lowering even though GlobalSemaphore storage was available. Some applications also need to reserve every local semaphore ID for their own synchronization.

### What's changed

One shared counter descriptor and allocator now serves completion and readiness synchronization. By default, it allocates hardware-local semaphore IDs first, then uses host-created GlobalSemaphores whose addresses are passed through the existing runtime resource plan. Global readiness indices are reused across source cores because each core has independent storage.

`--ttl-pipe-global-semaphores-only` allocates every compiler-managed PipeNet completion and readiness counter in GlobalSemaphore storage, preserving all local semaphore IDs for application use. Counter coloring and reuse are unchanged.

TT-Metal validates target-dependent runtime-argument and L1 capacity during program construction; the compiler does not duplicate those runtime limits.

The Python PipeNet graph no longer estimates semaphore counts. Resource allocation depends on IR liveness, receiver overlap, and selected protocols, so the compiler resource plan is the single source of truth.

### Validation

New regressions exhaust local semaphore IDs, verify global completion and readiness allocation and source-core reuse, and prove that global-only mode emits no local semaphore lookup. Device tests compare local-first computed addressing, local-first receiver-published addressing, and global-only computed addressing across BF16 and FP32 transfers.

### Checklist


zoecarver (approved):
As a follow up is fine: would it be possible to add a mode that only uses global semaphores because some projects may want to reserve local semaphores for other things.

brnorris03:
> As a follow up is fine: would it be possible to add a mode that only uses global semaphores because some projects may want to reserve local semaphores for other things.

Yeah, that's easy enough, will do.

PR #766, opened 2026-07-21
[ttl] call_extern_func

arichinsTT:
### Problem description

tt-lang has no escape hatch for calling hand-written or library C/C++ kernel
code. Users who need a custom LLK, an external op from a shared header, or a
one-off shim have no way to invoke it by name from a `@ttl` thread. This adds
`ttl.call_extern_func`, an opaque call that emits a direct C++ function call
plus the `#include` for the header that declares it.

#### New ops

- `ttl.opaque_call` / `ttkernel.opaque_call` -- emit a call to an external
  C/C++ function. `callee` and `header` are string attributes; `arg_operands`
  are forwarded positionally; `template_arg_vals` are i32 SSA values that
  become C++ template arguments. The foreign signature is not type-checked
  (opaque by design). Both carry `MemRead, MemWrite` effects and a verifier.
- `ttl.get_dfb_id` / `ttkernel.get_dfb_id` -- resolve a dataflow buffer to its
  compile-time CB index as an i32 SSA value. Keeping the index in SSA (rather
  than an attribute) lets it flow into `opaque_call` template args while
  holding the DFB live through normal dataflow.

#### Lowering

`convert-ttl-to-ttkernel` lowers both ops; `TTKernelToEmitC` lowers
`ttkernel.opaque_call` to `emitc.call_opaque` (tagging the header on the call)
and `ttkernel.get_dfb_id` to an integer literal. `TTKernelToCpp` collects the
tagged headers and emits the corresponding `#include` directives. Template
args are resolved to integer literals at EmitC time (`arith.constant`,
`get_dfb_id`, or their already-converted emitc forms).

```mlir
// TTL
%id = ttl.get_dfb_id %cb : !ttl.cb<[1,1], !ttcore.tile<32x32, bf16>, 2>
ttl.opaque_call "negate_tile_shim" template_args(%id) () {header = "negate_tile_op.hpp"} : () -> ()

// EmitC / generated C++
#include "negate_tile_op.hpp"
negate_tile_shim<2>();
```

#### Python DSL

`ttl.call_extern_func(header, callee, template_args=[...], func_args=[...],
include_paths=[...])` lowers to `ttl.opaque_call`. `include_paths` are
forwarded to the JIT compiler as `-I` flags via a new
`KernelDescriptor.compiler_include_paths` field.

```python
@ttl.compute()
def compute():
    with in_dfb.wait() as in_blk, out_dfb.reserve() as out_blk:
        ttl.call_extern_func(
            HEADER,                                  # header that declares the callee
            "negate_tile_shim",                      # callee name
            template_args=[ttl.get_dfb_id(in_dfb),   # -> C++ template args
                           ttl.get_dfb_id(out_dfb)],
        )
```

### Supported and unsupported arguments

**Supported**

| Kind | Accepts | Lowers to |
|------|---------|-----------|
| `template_args` | integer literal / module-level int constant | C++ template arg `callee<N>()` |
| `template_args` | `ttl.get_dfb_id(dfb)` | the DFB's compile-time CB index, as a template arg |
| `func_args` | scalar integer | forwarded as-is |
| `func_args` | scalar float | raw bit pattern in an integer register (tt-metal scalar-arg convention; the C++ param may be `float`/`bf16` or an unsigned integer) |
| `func_args` | dataflow buffer | its i32 CB index (`get_compile_time_arg_val`) |
| `include_paths` | list of strings | `-I` flags for the JIT compiler |

Template args must be compile-time evaluable; a runtime value or block
argument is rejected by the verifier.

**Not yet supported/Future work**

- Semaphores as arguments. #767 
- Raw tensor operands passed directly (only dataflow-buffer CB-index passing
  is supported today) #767 
- library style invocation (maybe via atoms)

### Tests

- Round-trip lit tests for `ttl`/`ttkernel` `opaque_call` and `get_dfb_id`.
- Invalid lit tests (`--verify-diagnostics`): empty callee/header, and
  block-argument / runtime template args.
- Conversion lit tests through `--convert-ttl-to-ttkernel`: void call,
  constant template args, DFB template arg, and DFB `func_arg`.
- `get_dfb_id.py`: compile-only check that `get_dfb_id` template args reach the
  generated C++ as raw integers (`my_shim<2, 1, 1>()`).
- `test_call_extern_func.py`: end-to-end negate-tile op via a struct-based
  external header, checked against a torch reference.

### Checklist


test/python/include/negate_tile_op.hpp
```diff
@@ -0,0 +1,70 @@
...
+} // namespace test_ops
+
+// Shim callable from ttl.call_extern_func via template_args.
+template <uint32_t InCB, uint32_t OutCB>
+void negate_tile_shim() {
+  using Config = test_ops::NegateTileConfig<InCB, OutCB>;
+  test_ops::NegateTileOp<Config> op;
+  op.init();
+  op.run();
+}
```
zoecarver, line 70: Can you add some more tests? CBs passed as templates and as normal args? Ints/bools passed as both? Tensors passed (or I guess that can be a follow up PR). I think it would be good to have runtime tests for all of these. And runtime tests where tt-lang deals with an unbalanced CB, so maybe it reserves it then calls a c++ function.

zoecarver: More test ideas: scalar float as arg and include path runtime test

python/ttl/_src/ttl_ast.py
```diff
@@ -599,6 +609,36 @@ def _is_ttl_block_access(self, node):
...
+            return True
+        if (
+            isinstance(func, ast.Attribute)
+            and func.attr == "get_dfb_id"
+            and isinstance(func.value, ast.Name)
+            and func.value.id == "ttl"
+        ):
+            return True
+        return False
+
```
zoecarver, line 627: optional nit: merge into single function that takes string as param since these are otherwise almost identical

lib/Dialect/TTKernel/IR/TTKernelOps.cpp
```diff
@@ -629,4 +629,29 @@ void UnpackStallOnPackOp::getCanonicalizationPatterns(
   });
 }
 
+::mlir::LogicalResult OpaqueCallOp::verify() {
```
zoecarver, line 633: Seems like it's the same here and in ttlops.cpp

zoecarver (approved):
LGTM pending added tests. Also a few optional nits.

test/python/include/negate_tile_op.hpp
arichinsTT: Yes for int/bool/float template and func args.  Added test with dfb as func args.  The unbalanced CBs I want to leave for later, and I'm not sure how to test the include path runtime, as I think it would need to reference something custom outside of our and metals repos

PR #768, opened 2026-07-22
[ttl][analysis] Add value origin analysis

brnorris03:
#768 (this PR) -> #700

### Problem description

Pipe lowering must conservatively identify every producer that can reach a handle, token, or pipe value through control flow and tensor containers. Missing or ambiguous origins can generate incorrect synchronization, while performing this analysis in op verifiers is non-local and repeats expensive work.

### What's changed

- Makes `ValueOriginAnalysis` conservative, cached, bounded, and reusable across consumers.
- Moves transfer provenance checks into a reachability-aware pass.
- Converts transfer state through SCF and tensor containers while preserving each dynamic receive token's completion sequence.

### Checklist

- [x] New and existing tests cover the changes.
- [x] `check-ttlang-all` passes.


include/ttlang/Analysis/ValueOriginAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+/// `tensor.extract`, it also follows the extracted tensor element through
+/// `tensor.cast` and `tensor.insert`. When traversal reaches an unmodeled
+/// producer, its result remains an opaque origin so consumers cannot treat
+/// unsupported value flow as a proven source.
+class ValueOriginAnalysis {
+public:
+  /// Limits compile-time enumeration of finite loop and tensor index domains.
+  struct Options {
+    /// Maximum loop iterations examined by one access-domain comparison.
+    std::uint64_t maxEnumeratedLoopIterations = 1'000'000;
```
zoecarver: This seems... high. Have you checked compile time? Maybe make this configurable?

zoecarver (approved):

include/ttlang/Analysis/ValueOriginAnalysis.h
brnorris03: Yes, compile time has been measured -- small fraction of a second at last check

brnorris03: Also there were some more improvements committed since then which reduced that value as well

brnorris03: Anyway, just reran the exact 1,000,000-iteration version -- it averages 4.12 ms per complete parse-and-analysis invocation.

PR #769, opened 2026-07-22
[ttl] Add data movement subtile support 

jackzhangTT:
### Problem description
TT-Lang assumed a default 32x32 tile in the Python frontend when building tensor types, dataflow buffers, and CB descriptors. Sub-tile shapes from tensor.get_tile() (e.g. 16x32, 1x16) were dropped, so DFBs and CB page sizing did not match the real tile bytes. 

### What's changed
Threaded tile shape through the Python frontend so DFBs, tensor/layout typing, and CB page sizing honor tensor.get_tile(). Added DM-only pytest coverage for page tensor to CB to tensor round trip copies. Including single-tile round-trips and a neighbor-tile check when page < aligned, to confirm the NOC transfer uses logical tile size rather than oversizing to the aligned page.

### Ticket
[Issue](https://github.com/tenstorrent/tt-lang-ops-and-models/issues/9) 

### Checklist

### DIFF
--- a/python/ttl/kernel_runner.py
+++ b/python/ttl/kernel_runner.py
@@ -421,14 +421,16 @@ def build_cb_descriptors(
                     data_format,
                     page_size,
                     total_size,
+                    None,
                     f"  CB[{i}]: compiler-allocated num_tiles={cb.num_tiles} "
                     f"block_count={cb.block_count} format={cb.data_format} -> {total_size} bytes",
                 )
             )
         else:
             data_format = _cb_data_format(cb)
-
-            page_size = tile_bytes_from_dtype(data_format)
+            tile = ttnn.Tile(cb.tile)
+            tile_descriptor = ttnn.TileDescriptor(tile)
+            page_size = tile.get_tile_size(data_format)
             num_tiles = cb.shape[0] * cb.shape[1] * cb.block_count
             total_size = num_tiles * page_size
             rows.append(
@@ -436,6 +438,7 @@ def build_cb_descriptors(
                     data_format,
                     page_size,
                     total_size,
+                    tile_descriptor,
                     f"  CB[{i}]: shape={cb.shape} block_count={cb.block_count} -> {total_size} bytes",
                 )
             )
@@ -454,7 +457,7 @@ def build_cb_descriptors(
     # Must stay aligned with MLIR ttl-validate-cb-budget (TileType::getSizeBytes) and
     # tile_bytes_from_dtype; see issue #511.
     if total_cb_bytes > remaining_bytes:
-        breakdown = "\n".join(r[3] for r in rows)
+        breakdown = "\n".join(r[-1] for r in rows)
         raise ValueError(
             "Total circular buffer allocation ("
             f"{total_cb_bytes} bytes) exceeds L1 budget ({remaining_bytes} bytes). "
@@ -464,11 +467,14 @@ def build_cb_descriptors(
         )
 
     cb_descriptors = []
-    for i, (data_format, page_size, total_size, _) in enumerate(rows):
+    for i, row in enumerate(rows):
+        data_format, page_size, total_size = row[:3]
+        tile_descriptor = row[3]
         cb_format = ttnn.CBFormatDescriptor(
             buffer_index=i,
             data_format=data_format,
             page_size=page_size,
+            **({"tile": tile_descriptor} if tile_descriptor is not None else {}),
         )
         cb_desc = ttnn.CBDescriptor(
             total_size=total_size,
--- a/python/ttl/operators.py
+++ b/python/ttl/operators.py
@@ -98,6 +98,17 @@ def _get_constant_float(val) -> float:
     return v
 
 
+def _tile_hw(elem_type) -> Optional[Tuple[int, int]]:
+    """Return (H, W) when ``elem_type`` is a TileType, else None."""
+    from ttl.dialects import ttcore
+
+    tile = ttcore.ir.TileType.maybe_downcast(elem_type)
+    if tile is None:
+        return None
+    shape = list(tile.shape)
+    return (int(shape[0]), int(shape[1]))
+
+
 # Type aliases for common patterns
 CoreCoordinate = Tuple[int, int]
 IndexedTensor = Union["TensorBlock", Tuple["TensorBlock", Tuple[int, ...]]]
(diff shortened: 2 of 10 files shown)


jackzhangTT:
One thing to note in `TTLValidateCBBudget.cpp` there exists this function:
```
static uint64_t bytesPerCbElement(mlir::Type elemTy) {
  if (auto tileTy = mlir::dyn_cast<mlir::tt::ttcore::TileType>(elemTy)) {
    return tileTy.getSizeBytes();
  }
  return mlir::tt::ttcore::TileType::get(elemTy).getSizeBytes();
}
```
Should I update this to check if `elemTy` is already a `TileType` like I did in `ConvertTTLToCompute.cpp`? This path wasn't executed in my test but maybe I should as a preventive measure?

test/python/test_subtile_dfb.py
```diff
@@ -0,0 +1,224 @@
...
+
+import pytest
+import torch
+
+ttnn = pytest.importorskip("ttnn", exc_type=ImportError)
+
+import ttl
+from ttlang_test_utils import assert_allclose
+
+SUBTILE_SHAPES = [
```
zoecarver: what about sub-tile shapes that are something like 8x32? these don't need to get wired through compute, but we should be able to construct a CB with that tile descriptor so that we can send them to external metal functions. Ideally we could also take a tensor tilized like that and pass it through data movement, into a CB, and then into a metal function.

test/python/test_subtile_dfb.py
```diff
@@ -0,0 +1,224 @@
...
+        torch_tensor,
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        tile=tile,
+        memory_config=memory_config,
+    )
+
+
+def _assert_initial_mlir_tile(
```
zoecarver: I think it would be better to do this kind of checking with lit rather than asserts.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -28,6 +28,17 @@ static RankedTensorType getTensorType(Value v) {
   return dyn_cast<RankedTensorType>(v.getType());
 }
 
+/// Resolve a tensor element type to the tile type used as a compute block
```
zoecarver: Cool that this was able to be wired all the way through compute, but I wonder if we don't need some more robust testing here. If we want to wire sub-tiles through compute, I think we need to ensure the sub-tile is a face, i.e. the four cases you laid out in your test (and not 1x32 for example). Can you please raise an error if a non 16-bit-aligned tile shape makes its way into compute? 

Then we also need to test all of the ops with these tiles, so the eltwise sweep, matmul, reduce, broadcast, fill (and maybe some others?).

zoecarver: Also @brnorris03 do you see any dragons with sub-blocking or dst allocation? I would have thought we might need updates there.

zoecarver: Again, I think the compute side of this is optional, so if this ends up becoming too large of a change, feel free to make it a follow up.

zoecarver:
One thing I'd like to confirm: a really small tile with maybe a small dtype, does an tensor tile copy have a lower bound for the size we copy? If you copy a single page/tile from from a tensor with our current lowering, I think it can copy a minimum of 64 bytes, so a very small tile might be less than that, and we might end up putting garbage into the destination. Can you create a test for that and see what happens?

python/ttl/kernel_runner.py
```diff
@@ -418,21 +418,24 @@ def build_cb_descriptors(
                     data_format,
                     page_size,
                     total_size,
+                    None,
```
brnorris03, line 424: This would introduce a bug because compiler-allocated DFBs lose the tile dimensions here. `ttl.compiler_allocated_dfbs` retains the full `!ttcore.tile<HxW, dtype>`, but `_extract_compiler_allocated_dfbs` keeps only the dtype. This branch therefore uses the 32x32-only `tile_bytes_from_dtype` result and omits the tile descriptor. A 16x16 bf16 compiler-allocated DFB is 512 bytes per page in MLIR but 2048 bytes per page at runtime ([reproducer](https://gist.github.com/brnorris03/17e24cc83d6bcd9ed92aec05001f0872)). This affects compiler-generated intermediate DFBs in the accumulation stack and would also make PR #700's computed receiver-address stride disagree with the runtime DFB after restacking (which would break pipes that may involve subtiles). You can preserve the tile dimensions in `CompilerAllocatedDFBConfig` and use the same tile-aware descriptor construction for compiler-allocated DFBs.

test/python/test_subtile_dfb.py
```diff
@@ -0,0 +1,224 @@
...
+def _to_device_with_tile(torch_tensor, device, tile_hw, memory_config):
+    """Create a TILE_LAYOUT tensor with an explicit tile shape.
+
+    Prefer direct placement in the target memory space. DRAM->L1 via
+    ttnn.to_memory_config currently rejects mismatched tile shapes.
+    """
+    tile = ttnn.Tile(tile_hw)
+    return ttnn.from_torch(
+        torch_tensor,
+        dtype=ttnn.bfloat16,
```
brnorris03: I think this test should be parameterized for the supported sub-tile cases over bf16 and fp32 with dtype-appropriate tolerances (there are other similar tests but let me know and I'll point to specific examples if not obvious).

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
brnorris03: Yes, I agree that compute may be better as a follow up (ensuring dm-only in this PR)

zoecarver:
Another review comment I just thought of: what if we try to copy (store) between CBs with different tilization? We should probably add an error for that. 

And for tensors we should either make it work or also add an error. So the full matrix of tensor -> tensor, cb -> cb, tensor -> cb, and cb -> tensor where LHS and RHS are different tiles.

python/ttl/operators.py
```diff
@@ -378,6 +384,39 @@ def _get_cb_shape(cb_val):
     return list(cb_type.shape)
 
 
+def _tile_hw(elem_type) -> Optional[Tuple[int, int]]:
+    """Return (H, W) when ``elem_type`` is a TileType, else None."""
+    from ttl.dialects import ttcore
```
zoecarver: nit move to top of file

zoecarver (approved):
Looks great, thank you!

PR #771, opened 2026-07-25
Simulator: Run all spec examples

FConstantinos:
This commit makes sure that the simulator runs all TT-Lang specification examples except the ones that include semaphores (not yet implemented). Specifically:

 - All spec examples added to the simulator test suite
 - Semaphore examples expected to fail, all others expected to pass
 - New spec examples need to have a corresponding test in the simulator (could either be passing or be an xfail)

Features:

- atom_split.py is now reused in the simulator to split monokernels into multiple kernels
- kernel independent code (in ttl.operation but outside ttl.compute/datamovement) now runs on every node before kernels run
- ttnnsim more faithfully simulates actual ttnn
  - 0/1-D supported in from_torch
  - add ttnn.Shape/zeros/abs/exp to ttnnsim
  - allow open slices
  - logical and padded shapes now both kept
  
 - Compatibility break: NdShardSpec now takes the number of cores as num_cores=. Callers using the previous keyword argument name must update to num_cores=; positional calls are unaffected.


docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -88,10 +88,8 @@ def __add(
     # and the addition on the compute thread.
     a_dst_blk = a_dfb.reserve()
     b_dst_blk = b_dfb.reserve()
-    a_tx = ttl.copy(a[0:1, 0:1], a_dst_blk)
-    b_tx = ttl.copy(b[0:1, 0:1], b_dst_blk)
-    a_tx.wait()
-    b_tx.wait()
+    ttl.copy(a[0:1, 0:1], a_dst_blk).wait()
+    ttl.copy(b[0:1, 0:1], b_dst_blk).wait()
```
brnorris03, line 93: I would double check this works with the compiler -- I think it should, but just in case.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -409,7 +406,6 @@ y_dfb = ttl.make_dataflow_buffer_like(y, shape=(N_BLOCK_SIZE, 1))
 # of each tile in a row of M_TILES tiles
 z_dfb = ttl.make_dataflow_buffer_like(z, shape=(1, M_TILES))
 
-
```
brnorris03, line 417: I am wondering whether all the formatting changes are something that will cause some churn -- do you know why all these blank lines happened to be removed now?

python/sim/ttnnsim.py
```diff
@@ -1299,11 +1313,39 @@ def _maybe_resolve_nd_shard_spec_for_tensor(
...
+    """Constructible tuple of dimensions, mirroring ``ttnn.Shape``.
+
+    Accepts ``Shape([d0, d1, ...])``, ``Shape((d0, d1, ...))``, or
+    ``Shape(d0, d1, ...)``. Instances are ordinary tuples, so they interoperate
+    with the tuple-based shapes (``Shape``) used throughout the simulator:
+    indexing, equality with plain tuples, passing to ``torch`` factory
+    functions, etc.
+    """
+
+    def __new__(cls, *dims: "int | Sequence[int]") -> "Shape":
```
brnorris03: Real `ttnn.Shape` takes a single sequence and isn't a tuple, so this is looser in both directions -- `Shape(32, 64)`, slicing, concatenation and `isinstance(..., tuple)` all work here and none work on device, while `.rank` / `.to_rank()` work on device and are missing here. I think the simulator should be stricter than ttnn rather than looser, otherwise code passes in sim and fails on hardware.

python/sim/ttnnsim.py
```diff
@@ -1299,11 +1313,39 @@ def _maybe_resolve_nd_shard_spec_for_tensor(
...
+            return super().__new__(cls, dims[0])
+        return super().__new__(cls, cast("Sequence[int]", dims))
+
+
 def _dtype_element_size(dtype: torch.dtype) -> int:
     """Return the element size in bytes for a torch dtype."""
     return torch.tensor([], dtype=dtype).element_size()
 
 
+class Tile:
```
brnorris03: `t.tile` builds a fresh `Tile` on every access and the class has no `__eq__`, so `t.tile == t.tile` is `False` here and `True` on device. Real `Tile` also exposes `face_shape`, `num_faces` and `get_tile_size(dtype)` -- suggest a module-level singleton plus `__eq__` at minimum.

python/sim/ttnnsim.py
```diff
@@ -1447,27 +1530,27 @@ def _normalize_index(selector: Selector) -> slice:
...
-        """Validate a tile-coordinate slice has explicit bounds and no step.
+    def _resolve_tile_slice(
+        s: slice, tile_count: int, dim_name: str
+    ) -> Tuple[int, int]:
+        """Resolve a tile-coordinate slice to explicit ``(start, stop)`` tile bounds.
+
+        Open ends follow Python slice semantics: a missing start defaults to 0
+        and a missing stop defaults to ``tile_count`` (the full extent along the
+        dimension). This lets ``t[i, :]`` / ``t[:, j]`` select whole rows/columns
+        of tiles, matching ttnn. Steps remain unsupported.
```
brnorris03: I don't think this matches ttnn -- `ttnn.Tensor.__getitem__` mimics PyTorch element indexing over the logical shape and drops the dimension for an int index, while this is tile-coordinate addressing. On a 64x64 tensor `t[0:32, 0:32]` returns the whole tensor here and a 32x32 element view on device, and `t[0, :]` gives `(32, 64)` vs `(64,)`. Now that `.shape` reports logical element extents the two conventions are easy to confuse -- worth stating in the class docstring that indexing is tile-space.

python/sim/ttnnsim.py
```diff
@@ -1447,27 +1530,27 @@ def _normalize_index(selector: Selector) -> slice:
...
-            raise ValueError(
-                f"Tile slice '{dim_name}' must have explicit stop value, "
-                f"got slice({s.start}, {s.stop}, {s.step})"
-            )
         if s.step is not None:
             raise ValueError(
                 f"Tile slice '{dim_name}' must not have a step value, "
                 f"got slice({s.start}, {s.stop}, {s.step}). Only simple slices are supported."
             )
+        start = 0 if s.start is None else s.start
```
brnorris03, line 2110: No bounds check on the resolved indices -- `t[0:100, 0:100]` on a 2x2-tile tensor returns the whole tensor instead of raising, because torch clamps the element slice. More reachable now that an open end resolves against the tile count. Can we assert `0 <= start <= stop <= tile_count`?

python/sim/program.py
```diff
@@ -189,201 +207,363 @@ def _build_node_context(self, node: int) -> Dict[str, Any]:
...
+    ctx = get_context()
+
+    node_plans: Dict[int, tuple[Dict[str, Any], tuple[BindableTemplate, ...]]] = {}
+    all_nets: List[Any] = []
+    dfb_count = 0
+    l1_bytes = 0
+
+    body_globals = getattr(body, "__globals__", {})
+    try:
+        for node in range(total_nodes):
```
brnorris03, line 249: This runs the body for every node, so PipeNet-inactive nodes still execute setup and allocate dataflow buffers -- with a net covering only nodes 0 and 1 on a `(2, 2)` grid I get the body running on `[0, 1, 2, 3]`. `_schedule_and_run` says inactive nodes mirror the compiler's `scf.if` guard, but that only holds for the kernels.

examples/spec/grid/grid_size.py
```diff
@@ -9,23 +9,47 @@
...
+    # for (8, 8) single-chip or SPMD grid gets x_size = 8, y_size = 8, z_size = 1
+    x_size, y_size, z_size = ttl.grid_size(dims=3)
+    # spec:end
+
+    # Grid size is node-independent; on a (8, 8) grid the successive forms yield
+    # 64, then (8, 8), then (8, 8, 1). (Scaffolding assertion, not rendered.)
+    assert (x_size, y_size, z_size) == (8, 8, 1)
+
+    @ttl.compute()
+    def _noop_compute() -> None:
```
brnorris03: I don't think these three are needed -- with no kernels the body is classified unified and the splitter synthesizes all three. I dropped them and `grid_size_example()` still runs and passes; same for `node.py`. Keeps the example closer to what it documents.

examples/spec/block/batched_matmul_bias.py
```diff
@@ -19,170 +22,203 @@
...
-
-                    # End of "with" scope:
-                    # Push c_blk to make it ready for matmul_compute
-
-                # Repeat for each K block
-                for k_block in range(K_BLOCKS):
-                    k_slice = slice(
-                        k_block * K_BLOCK_SIZE, (k_block + 1) * K_BLOCK_SIZE
+# Concrete compile-time sizes for a single-block run (scaffolding, not rendered).
+TILE_SIZE = 32
```
brnorris03: "(scaffolding, not rendered)" is kind of weird, IMO, is it something that is required for parsing these (I don't recall), or can it be omitted (it's in all examples)?

examples/spec/block/batched_matmul_bias.py
```diff
@@ -19,170 +22,203 @@
...
+                                    dims=[0],
+                                    shape=(L_BLOCK_SIZE, K_BLOCK_SIZE, N_BLOCK_SIZE),
+                                )
+
+                                # Accumulate dot product between L_BLOCK_SIZE×M_BLOCK_SIZE×K_BLOCK_SIZE a_blk and
+                                # L_BLOCK_SIZE×K_BLOCK_SIZE×N_BLOCK_SIZE b_bcast in y_final
+                                y_final += a_blk @ b_bcast
+
+                                # End of "with" scope:
+                                # Pop a_blk and b_blk to make them available for matmul_read to load and push next blocks
```
brnorris03: nit: this comment makes it sound like the example is incomplete (but I know it means that pops will be inserted), I'd reword to make it clearer

examples/spec/block/elementwise_broadcast_reduce.py
```diff
@@ -19,217 +22,275 @@
...
+    c_dfb = ttl.make_dataflow_buffer_like(c, shape=(1, M_TILES))
+    # When tiled, the scalar value d of shape () is placed at position (0, 0)
+    # of a single tile
+    d_dfb = ttl.make_dataflow_buffer_like(d, shape=(1, 1))
+    # When untiled, the vector y is formed from column 0
+    # of each tile in a column of N_TILES tiles
+    y_dfb = ttl.make_dataflow_buffer_like(y, shape=(N_BLOCK_SIZE, 1))
+    # When untiled, the vector z is formed from row 0
+    # of each tile in a row of M_TILES tiles
+    z_dfb = ttl.make_dataflow_buffer_like(z, shape=(1, M_TILES))
```
brnorris03: Nit: I don't think 80-character wrapping is necesary, it would be more readable with 100 or 120 (comments)

examples/spec/block/elementwise_broadcast_reduce.py
```diff
@@ -19,217 +22,275 @@
...
-
-                # Accumulate-add partial z_final
-                z_final += z_partial
+                a_xf.wait()
+                b_xf.wait()
 
                 # End of "with" scope:
-                # Pop a_blk and b_dfb to make them available for elwise_read to load and push next blocks;
-                # Push y_blk to make it ready for elwise_write
+                # Push a_blk and b_blk to make them ready for elwise_compute
```
brnorris03: Same comment as above for the pop (and elsewhere), perhaps add "implicit" in front of the push/pop in the comments at the end of `with` statements

examples/spec/pipe/gather.py
```diff
@@ -9,78 +9,154 @@
...
+    # (0, 2) (1, 2) (2, 2) (3, 2)
+    # (0, 3) (1, 3) (2, 3) (3, 3)
+
+    # ---------------------
+    # gather from row y to (0, y) with unicast.
+    #
+    # The pipe net is sized from the active set, not the launch extent.
+    # ROWS and COLS describe the rectangle that bounds the active set.
+    # Nodes outside the active rectangle (row 0..ROWS-1, column 0..COLS-1)
+    # skip the operation body.
```
brnorris03, line 52: I know that's pre-existing, but I wonder if it should be in the examples at all, it's an implementation detail that may vary by implementation and also may fall out of sync with the spec itself easily I think. @phizalev-TT, what do you think?

examples/spec/semaphore/many_to_one_barrier.py
```diff
@@ -9,34 +9,56 @@
...
-node_0_barrier = my_barrier.get_remote((0, 0))
-non_0_node_count = grid_size(dims=1) - 1
+# Concrete grid for a standalone run (scaffolding, not rendered).
+GRID_X, GRID_Y = 2, 2
+
+
+@ttl.operation(grid=(GRID_X, GRID_Y))
+def many_to_one_barrier() -> None:
+    # The rendered spec spells the grid helper unqualified; bind it to the ttl
+    # helper so the marked lines below stay verbatim.
```
brnorris03: Seems like a superfluous and kind of confusing comment?

examples/spec/semaphore/one_to_many_barrier.py
```diff
@@ -9,33 +9,52 @@
...
-import math
-
-import torch
+# Everything outside the markers (imports, the @ttl.operation wrapper, the
+# no-op kernels) exists so the file can run standalone; it is not copied into
+# the specification. The marked lines are nested inside @ttl.operation and
+# dedented on render, so the rendered spec is unchanged.
+#
+# NOTE: the user-facing ttl.Semaphore barrier API is not implemented in the
+# simulator (or the compiler) yet, so this example is expected to fail at the
```
brnorris03, line 18: This is not about the current PR, but it reminded me that we should add an issue for implementing `ttl.Semaphore` in the simulator and compiler if it doesn't exist (I think it doesn't) and reference it here

python/sim/ttnnsim.py
```diff
@@ -1318,14 +1360,28 @@ def __init__(
...
         self._tensor: torch.Tensor = tensor
         self._layout: IndexType = layout
+        # The logical (user-visible) shape, mirroring ttnn.Tensor.shape: the
+        # dimensions of the actual data, before tile-alignment padding and
+        # before low-rank inputs are lifted for storage.  ``padded_shape``
+        # reports the physical stored shape.  ``logical_shape`` defaults to the
+        # backing tensor's shape for tensors produced by internal ops (e.g.
+        # arithmetic / slicing), whose data is already at physical extent; the
+        # creation entry points (from_torch / rand / empty / zeros) pass the
+        # caller's original shape so ``.shape`` matches ttnn.
```
brnorris03, line 1932: "before low-rank inputs are lifted for storage" -- what does "lifted for storage" mean exactly?

python/sim/unified_operation.py
```diff
@@ -0,0 +1,272 @@
...
+    """
+    try:
+        fn_def = _parse_operation_funcdef(func)
+    except (OSError, TypeError, ValueError):
+        return False
+    for node in ast.walk(fn_def):
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            if any(_is_kernel_decorator(d) for d in node.decorator_list):
+                return False
+    return True
```
brnorris03: nit: I'd group private methods together and public ones together if possible

brnorris03 (commented):
Mostly minor questions and comments in addition to the things shared in Slack.

docs/sphinx/specs/TTLangSpecification.md
FConstantinos: It does, there are tests for the compilers that exercise just that

docs/sphinx/specs/TTLangSpecification.md
FConstantinos: It's just that this code now resides in runnable files and it needs to be properly nested inside ttl.operation blocks that don't appear here. Maybe we could make the renderer not care for that, I'm not sure. But it's a one time thing

python/sim/ttnnsim.py

python/sim/ttnnsim.py
FConstantinos: fixed

examples/spec/grid/grid_size.py
FConstantinos: fixed

examples/spec/block/batched_matmul_bias.py
FConstantinos: not required, removed from examples

examples/spec/block/batched_matmul_bias.py
FConstantinos: reworded

examples/spec/block/elementwise_broadcast_reduce.py
FConstantinos: Made it wider

examples/spec/block/elementwise_broadcast_reduce.py
FConstantinos: done

examples/spec/semaphore/many_to_one_barrier.py
FConstantinos: reworded

examples/spec/semaphore/one_to_many_barrier.py
FConstantinos: I've mentioned some preexisting tickets

python/sim/ttnnsim.py
FConstantinos: expanded on a comment for that

python/sim/unified_operation.py
FConstantinos: done

python/sim/ttnnsim.py
FConstantinos: I've updated the docstring

python/sim/ttnnsim.py

python/sim/program.py
FConstantinos: Hmm, I'm not sure how this is supposed to work, mainly because the spec is silent on where and when the setup runs. @phizalev-TT should we address where the part of the operation that is outside the kernels (let's call it "setup phase") runs?

Another issue is that PipeNets are defined in the setup phase, so it has to run/analyzed before we know the active set of nodes. If we make it so that the nodes inside the pipenets are compile time constant then I can see that working, but what if they are defined e.g by user input?

Then, Claude tells me that the compiler also defines the dfb allocations for all nodes, regardless of whether they are in PipeNets or not, so it would be inconsistent with the compiler as well to not do that.

So I'm not sure what to do here. I'd prefer if we tackled this on another PR

brnorris03: > Another issue is that PipeNets are defined in the setup phase, so it has to run/analyzed before we know the active set of nodes. If we make it so that the nodes inside the pipenets are compile time constant then I can see that working, but what if they are defined e.g by user input?
> 
> Then, Claude tells me that the compiler also defines the dfb allocations for all nodes, regardless of whether they are in PipeNets or not, so it would be inconsistent with the compiler as well to not do that.

We don't allow dynamic pipenets at the moment, I believe. For the dfb allocation -- that's actually moving toward having node-specific allocation, but I would consider that optimization, I don't expect the sim to match total number of physical dfbs, pretty sure it's impossible since the compiler generates them as needed for intermediate storage.

python/sim/__init__.py
```diff
@@ -153,7 +150,6 @@ def signpost(*args: Any, **kwargs: Any) -> _SignpostContextManager:
     "copy",
     "CopyTransaction",
     "GroupTransfer",
```
brnorris03, line 158: Since spec 0.2 removed `ttl.Program`, I think we should remove it from the compiler exports too and update `test_ttl_api_basic`. Otherwise the compiler and simulator expose different APIs, and the test fails under sim.

python/sim/program.py
```diff
@@ -93,297 +96,427 @@ def get_max_l1_bytes() -> int:
...
+        # A node index left behind would answer ttl.node() outside a run.
+        body_globals.pop("_node", None)
+
+    pipenets = build_pipenets(_dedupe_pipe_nets(all_nets))
+    pipenets.validate()
+
+    _schedule_and_run(
+        node_plan_for=lambda n: node_plans[n],
+        candidate_nodes=range(total_nodes),
+        analysis_templates=node_plans[0][1],
```
brnorris03: I think this needs to analyze every node's kernels. When kernel selection depends on `ttl.node()`, using only node 0 means some kernels don't get copy-wait injection. `_visited` should already dedupe shared code.

python/sim/program.py
```diff
@@ -93,297 +96,427 @@ def get_max_l1_bytes() -> int:
...
+
+    The compiler evaluates the body once instead, at compile time, so a body
+    that mutates state of the enclosing scope sees those effects repeated here
+    and not there.  The specification does not currently say how many times an
+    implementation may evaluate the body.
+    """
+    from .decorators import clear_kernel_registry, get_registered_kernels
+    from .pipe import build_pipenets, discover_pipe_nets_from_closures
+
+    total_nodes = _total_nodes(grid)
```
brnorris03, line 235: `grid=(0, 2)` leaves `node_plans` empty, then `node_plans[0]` raises `KeyError`. Should this reject non-positive grid dimensions when validating `@ttl.operation`?

test/sim/test_examples.py
```diff
@@ -68,56 +65,52 @@ def reset_simulator_context():
...
+    "spec/pipe/scatter_gather.py",
+    "spec/pipe/forward_neighbor.py",
+    "spec/pipe/gather.py",
+    # Shape introspection: these assert the shapes they demonstrate.
+    "spec/dataflow_buffer/tiled_tensor_shape.py",
+    "spec/dataflow_buffer/row_major_tensor_shape.py",
+    "spec/dataflow_buffer/dataflow_buffer.py",
+    "spec/operation_function/multi_kernel_operation.py",
+    # Run for their structure, with nothing to compare against: the debugging
+    # snippets are about what they print.  Reaching the end without a dataflow or
```
brnorris03: Agent finding: `debug_printing.py` and `signpost.py` only register module-scope kernels; neither invokes an operation, so their bodies never execute. Please wrap and call them through `@ttl.operation`. `debug_printing.py` should also assert its captured output.

python/sim/ttnnsim.py
```diff
@@ -2107,40 +3063,157 @@ def to_memory_config(tensor: Tensor, memory_config: MemoryConfig) -> Tensor:
...
+                return _logical_view(arg)
+            case _:
+                return arg
+
+    try:
+        result = golden_fn(
+            *(unpadded(a) for a in args),
+            **{k: unpadded(v) for k, v in kwargs.items()},
+        )
+    except Exception:
```
brnorris03, line 3120: Agent finding, haven't confirmed: `except Exception` also hides invalid logical operands. For example, `(3, 5) @ (7, 9)` fails in PyTorch, but the padded rerun succeeds on 32x32 storage and returns a tensor. Please distinguish unsupported logical extents from invalid operand semantics and propagate genuine shape errors.

python/ttl/_src/atom_rules.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def non_resource_assignment_names(fn_def: ast.FunctionDef) -> Set[str]:
+    """Names the body assigns itself, excluding the lifted construction.
+
+    These exist only where the body runs, so a lifted construction cannot read
+    them.
+    """
+    names: Set[str] = set()
+    for statement in fn_def.body:
```
brnorris03, line 288: `non_resource_assignment_names` misses names bound by `for`, `with ... as`, walrus expressions, imports, and `except ... as`. A hoisted construction that reads one then fails with `NameError` instead of the intended diagnostic. Please collect all Python binding forms and add focused tests.

python/sim/ttnnsim.py
```diff
@@ -266,22 +301,54 @@ class NdShardSpec:
...
     """
 
-    shard_shape: Shape
+    shard_shape: Sequence[int]
     core_ranges: Optional["CoreRangeSet"] = None
-    shard_grid: Optional[Shape] = None
+    shard_grid: Optional[ShardGrid] = None
     distribution: ShardDistributionStrategy = ShardDistributionStrategy.ROUND_ROBIN_1D
-    num_cores: Optional[int] = None
+    round_robin_cores: Optional[int] = None
```
brnorris03, line 315: I had Claude do a sweep of things that work on main but not here, so one of them -- `NdShardSpec(num_cores=...)` worked before this PR and now raises `TypeError`. Please document the keyword rename in the PR description as a compatibility break.

examples/spec/operation_function/operation_function.py
```diff
@@ -36,10 +34,8 @@ def __add(
     # and the addition on the compute thread.
     a_dst_blk = a_dfb.reserve()
     b_dst_blk = b_dfb.reserve()
-    a_tx = ttl.copy(a[0:1, 0:1], a_dst_blk)
-    b_tx = ttl.copy(b[0:1, 0:1], b_dst_blk)
-    a_tx.wait()
-    b_tx.wait()
+    ttl.copy(a[0:1, 0:1], a_dst_blk).wait()
```
brnorris03, line 37: Agent claim and suggestion, I haven't verified: `ttl.copy(...).wait()` serializes the two loads that this introductory example previously issued together. Please either document the splitter limitation in the rendered specification or keep the overlapping form there and use a separate serialized test fixture.

python/sim/unified_operation.py
```diff
@@ -0,0 +1,674 @@
...
+    ]
+    for path in candidates:
+        if path.is_file():
+            name = f"ttl_sim_{Path(filename).stem}"
+            spec = importlib.util.spec_from_file_location(name, str(path))
+            assert spec is not None and spec.loader is not None
+            module = importlib.util.module_from_spec(spec)
+            # Register before exec so @dataclass in the module can resolve its
+            # own __module__ via sys.modules during class processing.
+            sys.modules[spec.name] = module
```
brnorris03, line 89: If `exec_module()` raises, the partially initialized module remains in `sys.modules`, leaving this in a bad state. Maybe remove `spec.name` in an exception handler before re-raising?

python/sim/unified_operation.py
```diff
@@ -0,0 +1,674 @@
...
+    which is why this looks at the object at all.
+
+    Resolution can also withhold recognition the spelling would have given: a
+    body's own ``compute`` that is not this API's decorator resolves to something
+    else and is not read as a kernel. So this is not the compiler's rule plus
+    extra recognition -- the compiler applies only the spelling
+    (``atom_rules.defines_kernels_by_spelling``) -- and the two can disagree about
+    a body that reuses a kernel decorator's name for something else.
+
+    Misclassifying a multi-kernel body as unified splits it and silently produces
```
brnorris03: nit: `silently` is too cryptic, I'd be explicit about misclassification returns a wrong result. Also I'd check for that elsewhere, there were other similar comments.

python/sim/ttnnsim.py
```diff
@@ -1299,17 +1635,282 @@ def _maybe_resolve_nd_shard_spec_for_tensor(
...
+                return NotImplemented
+
+    def __ne__(self, other: object) -> bool:
+        result = self.__eq__(other)
+        if result is NotImplemented:
+            return result
+        return not result
+
+    def __hash__(self) -> int:
+        """The equal tuple's hash.
```
brnorris03: nit: `__hash__` needs one sentence explaining why defining `__eq__` requires an explicit hash. The remaining text repeats details documented above (I'd also remove comments and docstrings that restate code).

brnorris03 (approved):
So sorry for the delayed review! Looks good overall, mostly minor remaining things.

python/sim/program.py
FConstantinos: fixed

python/sim/program.py
FConstantinos: fixed

python/sim/__init__.py
FConstantinos: fixed

test/sim/test_examples.py
FConstantinos: fixed

python/sim/ttnnsim.py
FConstantinos: done

python/sim/unified_operation.py
FConstantinos: fixed

python/sim/ttnnsim.py
FConstantinos: done

python/sim/program.py
FConstantinos: I've filed  #881 and #882 for this

python/sim/ttnnsim.py
FConstantinos: tracked in #884

python/ttl/_src/atom_rules.py
FConstantinos: tracked in #883

python/ttl/atom.py
```diff
@@ -365,9 +187,7 @@ def _build_atom_spec(fn: Callable) -> _AtomSpec:
     captured_values = _captured_values(fn)
     external_pipenets = dict(inlined_pipenets)
```
brnorris03, line 271: I think there may have been an incorrect conflict resolution here -- should preserve main’s logical-kernel implementation and layer the simulator changes onto it? `inline_atom_calls()` now returns `(pipenets, logical_kernels)`, but this code treats it as one dict, so `@ttl.operation` on the python lit tests.

zoecarver (approved):

python/ttl/atom.py
FConstantinos: fixed

phizalev-TT (approved):

PR #776, opened 2026-07-27
[ttl] Add TileFillOp to  classifyTileOp

jackzhangTT:
### Problem description
classifyTileOp has no case for TileFillOp, so a tile_fill falls through
to TileOpCategory::Unknown and ttl-schedule-operations never schedules it.

### What's changed
Classify TileFillOp as a dedicated TileOpCategory::Fill so the fill
participates in scheduling instead of bounding it. The depth analysis already
models WAW and WAR hazards on DST, so the fill sorts after the last reader of
its slot and before its consumer while the operations around it stay in one
group.

### Test coverage

`schedule_fill_participation.mlir` places an independent broadcast after a
fill in two cases. The broadcast can only reach the front of the DST section if
the fill is scheduled alongside it; while the fill is a boundary, the section
splits in two, each half is already sorted, and nothing moves. The second case
additionally overwrites a register an earlier operation reads, so the
write-after-read dependency still pins the fill behind that reader even though
it now participates in the sort.

### DIFF
--- a/docs/development/DST_Utilization.md
+++ b/docs/development/DST_Utilization.md
@@ -547,7 +547,7 @@ algorithm assigns each op a depth (longest path from any root in the
 dependency graph), then sorts by a 6-component key:
 
 1. depthLevel — dependency depth (correctness constraint)
-2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L178-L187) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < CopyDst < DstIndex (no Reduce entry yet — reduction ops do not exist in TTL; when added, they would likely slot between Bcast and Transpose as full-init CB-input ops)
+2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < CopyDst < DstIndex < Fill (Fill is a constant `fill_tile` write to DST with no tile operands)
 3. opName — groups identical op types for init sharing (string comparison for determinism)
 4. initAffinity — groups ops sharing one init call (e.g., COL vs ROW bcasts, copies from different CBs)
 5. dstIdx — DST register index for deterministic ordering
--- a/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
+++ b/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
@@ -481,6 +481,7 @@ enum class TileOpCategory : uint8_t {
   SFPUBinary = 5, // DST -> DST binary (MATH-only init)
   CopyDst = 6,    // DST -> DST copy
   DstIndex = 7,   // Zero-cost SSA name for one existing DST slot
+  Fill = 8,       // const -> DST (MATH-only fill_tile_init, no tile operands)
   Unknown = 255
 };
 
--- a/lib/Dialect/TTL/IR/TTLOpsUtils.cpp
+++ b/lib/Dialect/TTL/IR/TTLOpsUtils.cpp
@@ -756,6 +756,9 @@ TileOpCategory classifyTileOp(Operation *op) {
   if (isa<TileTransposeOp>(op)) {
     return TileOpCategory::Transpose;
   }
+  if (isa<TileFillOp>(op)) {
+    return TileOpCategory::Fill;
+  }
 
   if (op->hasTrait<TTLStrategyDependentBinaryOpTrait>()) {
     FailureOr<TileExecutionStrategy> strategy =
--- a/test/python/test_dst_fill_clobber.py
+++ b/test/python/test_dst_fill_clobber.py
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression: an inline ``ttl.block.fill(k)`` operand under DST pressure.
+
+The kernel computes ``out = tanh(a) * sigmoid(a) * ttl.block.fill(k)`` (k a
+compile-time constant) and must match the golden. ``a`` has two consumers so
+AssignDST inserts a copy_tile for the sigmoid subtree; the expression has to be
+this deep for the inline fill to share that copy_tile's DST slot (shallower forms
+like ``(a + b) * fill(k)`` don't exercise the hazard).
+
+Parameterized over bf16 and f32: f32 halves the usable DST slots (8 -> 4), so the
+allocation pressure that forces the fill to share a slot differs between them.
+Both consumers of ``a`` are SFPU (tanh/sigmoid) so a single f32 input CB is legal;
+an FPU+SFPU mix on one CB is rejected in f32.
+"""
+
+import pytest
+import torch
+import ttl
+
+ttnn = pytest.importorskip("ttnn", exc_type=ImportError)
+
+from ttlang_test_utils import assert_allclose, to_dram
+
+K = 3.0
+
+_DTYPE_PARAMS = [
+    pytest.param(torch.bfloat16, id="bf16"),
+    pytest.param(torch.float32, id="f32"),
+]
+_DTYPE_TOL = {
+    torch.bfloat16: dict(rtol=5e-2, atol=1.0),
+    torch.float32: dict(rtol=1e-3, atol=1e-3),
+}
+
+
+@ttl.operation(grid=(1, 1))
+def fill_inline_kernel(a, out):
+    """out = tanh(a) * sigmoid(a) * fill(K)  -- the regression."""
+    a_dfb = ttl.make_dataflow_buffer_like(a, shape=(1, 1), block_count=2)
+    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+    @ttl.compute()
+    def compute():
+        with a_dfb.wait() as av:
+            with out_dfb.reserve() as o:
+                core = ttl.math.tanh(av) * ttl.math.sigmoid(av)
+                # Inline fill as an operand to the outer multiply -- clobbered.
+                # Typing the fill from o covers both dtypes with one kernel: a
+                # thread closure may only capture ints, floats, tensors and
+                # buffers, so the dtype cannot be passed in as a parameter.
+                o.store(core * ttl.block.fill(K, shape=o.shape, dtype=o.dtype))
+
+    @ttl.datamovement()
+    def dm_read():
+        with a_dfb.reserve() as a_blk:
+            ttl.copy(a[0, 0], a_blk).wait()
+
+    @ttl.datamovement()
+    def dm_write():
+        with out_dfb.wait() as blk:
+            ttl.copy(blk, out[0, 0]).wait()
+
+
+def _golden(a, k):
+    return torch.tanh(a.float()) * torch.sigmoid(a.float()) * float(k)
+
+
+def _inputs(dtype):
+    torch.manual_seed(0)
+    return torch.randn(32, 32, dtype=dtype)
+
+
+@pytest.mark.parametrize("dtype", _DTYPE_PARAMS)
+def test_fill_inline_matches_golden(device, dtype):
+    """The inline fill constant survives DST scheduling and reaches the mul."""
+    a = _inputs(dtype)
+    expected = _golden(a, K).to(dtype)
+
+    a_t = to_dram(a, device)
+    out_t = to_dram(torch.zeros(32, 32, dtype=dtype), device)
+
+    fill_inline_kernel(a_t, out_t)
+
+    result = ttnn.to_torch(out_t)
+    assert_allclose(result.float(), expected.float(), **_DTYPE_TOL[dtype])
+
+
+if __name__ == "__main__":
+    import sys
+
+    sys.exit(pytest.main([__file__, "-v", "-rxX", "--tb=short"]))
--- a/test/ttlang/Dialect/TTL/Transforms/ScheduleOperations/schedule_fill_participation.mlir
+++ b/test/ttlang/Dialect/TTL/Transforms/ScheduleOperations/schedule_fill_participation.mlir
@@ -1,0 +1,68 @@
+// Summary: A classified tile_fill participates in scheduling instead of
+// bounding it. Companion to schedule_ordering_boundaries.mlir, which covers the
+// operations that do bound a schedule. Both cases below place an independent
+// broadcast after the fill: the broadcast may only reach the front of the
+// section if the fill is scheduled with everything else, because an
+// unclassified fill is an ordering boundary that splits the section into two
+// independently sorted groups.
+// RUN: ttlang-opt %s --split-input-file \
+// RUN:   -pass-pipeline='builtin.module(func.func(ttl-schedule-operations))' \
+// RUN:   | FileCheck %s
+
+// The fill writes a register no other operation touches, so it carries no
+// dependency and sorts by category alone. It lands after the operations that
+// configure MATH from a dataflow buffer and before the exponential that depends
+// on the named register.
+
+// CHECK-LABEL: func.func @schedule_across_fill
+// CHECK:       ttl.dst_section {
+// CHECK-NEXT:    %[[BCAST:.*]] = ttl.tile_bcast
+// CHECK-NEXT:    %[[DST:.*]] = ttl.dst_index
+// CHECK-NEXT:    %[[FILL:.*]] = ttl.tile_fill
+// CHECK-NEXT:    %[[EXP:.*]] = ttl.tile_exp %[[DST]]
+// CHECK-NEXT:  }
+func.func @schedule_across_fill(
+    %input: !ttcore.tile<32x32, bf16>,
+    %output: !ttcore.tile<32x32, bf16>) {
+  %dst0 = arith.constant 0 : index
+  %dst1 = arith.constant 1 : index
+  %dst2 = arith.constant 2 : index
+  ttl.dst_section {
+    %dst = ttl.dst_index %input[%dst0] : !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %exp = ttl.tile_exp %dst into dst[%dst0] : !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %fill = ttl.tile_fill 3.000000e+00 into dst[%dst1] : !ttcore.tile<32x32, bf16>
+    %bcast = ttl.tile_bcast %input, %output 1 : i32 into dst[%dst2] : (!ttcore.tile<32x32, bf16>, !ttcore.tile<32x32, bf16>) -> !ttcore.tile<32x32, bf16>
+    ttl.yield
+  }
+  return
+}
+
+// -----
+
+// Participation does not weaken the register hazards. Here the fill overwrites
+// the register the exponential reads, so the write-after-read dependency gives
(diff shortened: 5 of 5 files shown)


docs/development/DST_Utilization.md
```diff
@@ -549,7 +549,7 @@ algorithm assigns each op a depth (longest path from any root in the
 dependency graph), then sorts by a 6-component key:
 
 1. depthLevel — dependency depth (correctness constraint)
-2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L178-L187) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < CopyDst < DstIndex (no Reduce entry yet — reduction ops do not exist in TTL; when added, they would likely slot between Bcast and Transpose as full-init CB-input ops)
+2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L178-L187) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < Fill < CopyDst < DstIndex (Fill is a constant `fill_tile` write to DST with no tile operands; no Reduce entry yet — reduction ops do not exist in TTL; when added, they would likely slot between Bcast and Transpose as full-init CB-input ops)
```
brnorris03: First, while you are here -- can you also remove the incorrect "no Reduce entry yet — reduction ops do not exist in TTL; when added, they would likely slot between Bcast and Transpose as full-init CB-input ops" portion of this sentence? Thanks!

Second, #L178-L187 in `[`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L178-L187)` is pointing at normalizeDim -- TileOpCategory is at L370 on main. It's a blob/main link, so the range re-resolves as the file grows. Can you either drop the line range or use a commit permalink?

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -374,8 +374,9 @@ enum class TileOpCategory : uint8_t {
   FPUBinary = 3,  // CB -> DST FPU (UNPACK+MATH init)
   SFPUUnary = 4,  // DST -> DST in-place (MATH-only init)
   SFPUBinary = 5, // DST -> DST binary (MATH-only init)
-  CopyDst = 6,    // DST -> DST copy
-  DstIndex = 7,   // Zero-cost SSA name for one existing DST slot
+  Fill = 6,       // const -> DST (MATH-only fill_tile_init, no operands)
+  CopyDst = 7,    // DST -> DST copy
+  DstIndex = 8,   // Zero-cost SSA name for one existing DST slot
```
brnorris03: I think it's better to add `Fill` than move existing enums, so leaving CopyDst and DstIndex where they are, and add `Fill = 8`.

test/ttlang/Dialect/TTL/Transforms/ScheduleOperations/schedule_fill_hazard.mlir
```diff
@@ -0,0 +1,83 @@
...
+// AssignDST emits the fill AFTER the mul that last reads dst1 (= sigmoid(a)).
+//
+// BUG (fixed): ttl-schedule-operations classified `tile_fill` as Unknown and
+// excluded it from scheduling. The remaining tile ops were then repositioned
+// around the stranded fill, hoisting `fill_tile(->dst1)` ahead of the
+// copy_tile/sigmoid that also write dst1. The fill constant was overwritten
+// before the final multiply read it, so the result silently dropped the
+// constant factor (observed: magnitude collapse). The fix classifies
+// `tile_fill` (TileOpCategory::Fill) so the scheduler tracks its DST write
+// hazards and keeps it after the last reader of that slot.
```
brnorris03: Kind of verbose, I would typically refrain from describing implementation details (and any implementation history beyond mentioning what the regression is) in a test comment -- the implementation should have enough information, so the test should only describe what it's testing briefly.

lib/Dialect/TTL/IR/TTLOpsUtils.cpp
```diff
@@ -187,6 +187,9 @@ TileOpCategory classifyTileOp(Operation *op) {
   if (isa<TileMatmulBlockOp>(op)) {
     return TileOpCategory::FPUBinary;
   }
+  if (isa<TileFillOp>(op)) {
+    return TileOpCategory::Fill;
+  }
   // TODO: add TileOpCategory::Transpose case when TTL transpose op is added.
```
brnorris03: This TODO is stale -- `ttl.tile_transpose` exists (TTLOps.td:1278), which also means `TileOpCategory::Transpose` is declared but never returned by anything.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -374,8 +374,9 @@ enum class TileOpCategory : uint8_t {
   FPUBinary = 3,  // CB -> DST FPU (UNPACK+MATH init)
   SFPUUnary = 4,  // DST -> DST in-place (MATH-only init)
   SFPUBinary = 5, // DST -> DST binary (MATH-only init)
-  CopyDst = 6,    // DST -> DST copy
-  DstIndex = 7,   // Zero-cost SSA name for one existing DST slot
+  Fill = 6,       // const -> DST (MATH-only fill_tile_init, no operands)
```
brnorris03: Nit: it does take a `$dst_index` operand -- `no tile operands`, like the doc line says, is more accurate.

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,153 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression: an inline ``ttl.block.fill(k)`` operand under DST pressure.
```
brnorris03, line 5: Same note as on the MLIR test -- the `Historic bug (fixed)` paragraph is implementation history. Please cut it to what the two kernels assert and why the expression has to be this deep. I'd shorten the comment significantly.

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,153 @@
...
+            tk.wait()
+
+    @ttl.datamovement()
+    def dm_write():
+        with out_dfb.wait() as blk:
+            tx = ttl.copy(blk, out[0, 0])
+            tx.wait()
+
+
+def _inputs(device):
```
brnorris03: `device` is unused.

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,153 @@
...
+    torch.manual_seed(0)
+    a = torch.randn(32, 32, dtype=torch.bfloat16)
```
brnorris03: bf16 only -- fp32 halves the DST slot count, so the pressure that makes the fill share a slot is different there. Can you parameterize over both? Look at `test_fill.py` or other dtype-parameterized pytests for examples.

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,153 @@
...
+    b = torch.randn(32, 32, dtype=torch.bfloat16)
+    return a, b
+
+
+def test_const_input_matches_golden(device):
```
brnorris03: This one passes with or without the fix (I checked) -- is the control worth the device time, or can we drop it?

lib/Dialect/TTL/IR/TTLOpsUtils.cpp
```diff
@@ -187,6 +187,9 @@ TileOpCategory classifyTileOp(Operation *op) {
   if (isa<TileMatmulBlockOp>(op)) {
     return TileOpCategory::FPUBinary;
   }
+  if (isa<TileFillOp>(op)) {
+    return TileOpCategory::Fill;
+  }
```
brnorris03, line 761: `tile_transpose` and `tile_reduce` are still in the state `tile_fill` was -- neither is classified, so the scheduler drops them too. Swapping `tile_transpose` into your new lit test reproduces the identical hoist (`transpose_wh_tile` lands above the `copy_tile`/`sigmoid` that reuse its slot).

Rather than adding cases one op at a time, can we change the filter in `TTLScheduleOperationsPass::runOnOperation` from `classifyTileOp(&op) != TileOpCategory::Unknown` to `isa<DstAccessOpInterface>(&op)`, and leave `classifyTileOp` used only where `TileOpSortKey` is built? Every `TTL_TileOp` implements that interface, so an op that writes DST gets its hazards tracked regardless of whether it has a category -- an unclassified one sorts last within its depth instead of being dropped.

On the enum itself: `Fill` no longer affects correctness once the filter changes, since an unclassified `tile_fill` lands at `Unknown`, sorts last within its depth, and fill has no init-ordering constraint. I would still keep it -- the root cause here was an op with no category, and leaving another one uncategorized will result in a similar problem. Once `tile_transpose` and `tile_reduce` are classified too (they need real categories anyway, per the CB-input-precedes-`copy_tile` note on the enum), `Unknown` would mean the op has no category, and you could assert on that in `scheduleOpsInRegion`. So `Fill = 8` as above.

test/ttlang/Dialect/TTL/Transforms/ScheduleOperations/schedule_fill_hazard.mlir
```diff
@@ -0,0 +1,83 @@
...
+       %out_tile: !ttcore.tile<32x32, bf16>):
+    %i = ttl.iter_index 0 : index
+    %j = ttl.iter_index 1 : index
+    %c0 = arith.constant 0 : index
+    %ab = ttl.tile_mul %a_tile, %b_tile into dst[%c0] : !ttcore.tile<32x32, bf16>, !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %t = ttl.tile_tanh %ab into dst[%c0] : !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %s = ttl.tile_sigmoid %a_tile into dst[%c0] : !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %g = ttl.tile_mul %t, %s into dst[%c0] : !ttcore.tile<32x32, bf16>, !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
+    %f = ttl.tile_fill 3.000000e+00 into dst[%c0] : !ttcore.tile<32x32, bf16>
+    %o = ttl.tile_mul %g, %f into dst[%c0] : !ttcore.tile<32x32, bf16>, !ttcore.tile<32x32, bf16> -> !ttcore.tile<32x32, bf16>
```
brnorris03: Putting the fill between the two muls makes insert-inits re-emit `mul_binary_tile_init` here on the key change. That re-init is required for correctness, so let's check for it (and `fill_tile_init` above) -- otherwise a change to init dedup could break the kernel with this test still passing.

I'd also capture and include dst indices in the checks including lines with `fill_tile` and `copy_tile` ops.

brnorris03 (commented):
Thanks for tracking this down! Looks good -- added some fairly minor comments throughout.

Up to you, but I think it makes sense to also fix `tile_transpose` and `tile_reduce` here the same way since they have the same problem.

docs/development/DST_Utilization.md
```diff
@@ -549,7 +549,7 @@ algorithm assigns each op a depth (longest path from any root in the
 dependency graph), then sorts by a 6-component key:
 
 1. depthLevel — dependency depth (correctness constraint)
-2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/main/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L178-L187) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < CopyDst < DstIndex (no Reduce entry yet — reduction ops do not exist in TTL; when added, they would likely slot between Bcast and Transpose as full-init CB-input ops)
+2. category — [`TileOpCategory`](https://github.com/tenstorrent/tt-lang/blob/cf10242c073e73dd5589972d30400f2b73883ab1/include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h#L370-L380) enum: Bcast < Transpose < CopyTile < FPUBinary < SFPUUnary < SFPUBinary < CopyDst < DstIndex < Fill (Fill is a constant `fill_tile` write to DST with no tile operands)
```
brnorris03: Thanks for the permalink -- but `cf10242c0` is this PR's first commit, where the enum still reads `Fill = 6, CopyDst = 7, DstIndex = 8`, so the link contradicts the ordering in this sentence (and that hash won't be on `main` after the squash). I'd point at a `main` commit or just drop the line range.

lib/Dialect/TTL/IR/TTLOpsUtils.cpp
```diff
@@ -187,7 +187,9 @@ TileOpCategory classifyTileOp(Operation *op) {
   if (isa<TileMatmulBlockOp>(op)) {
     return TileOpCategory::FPUBinary;
   }
-  // TODO: add TileOpCategory::Transpose case when TTL transpose op is added.
+  if (isa<TileFillOp>(op)) {
+    return TileOpCategory::Fill;
+  }
```
brnorris03, line 761: While you are here -- can you also add the transpose case? `TileOpCategory::Transpose` is declared but never returned, so `tile_transpose` gets `Unknown` and sorts after `copy_tile` at the same depth, the opposite of what the enum doc says. I tried `if (isa<TileTransposeOp>(op)) return TileOpCategory::Transpose;` and it emits `transpose_wh_init`/`transpose_wh_tile` ahead of the `copy_tile` as intended, with all 190 lit tests still passing.

lib/Dialect/TTL/IR/TTLOpsUtils.cpp
```diff
@@ -200,6 +202,14 @@ TileOpCategory classifyTileOp(Operation *op) {
   if (op->hasTrait<TTLTileBinaryOpTrait>()) {
     return TileOpCategory::SFPUBinary;
   }
+  // TODO: tile_transpose and tile_reduce still fall through to Unknown. They
```
brnorris03: Kind of verbose, and it names a consumer pass -- I'd keep the TODO to what's missing here (`tile_transpose` and `tile_reduce` have no category, so they sort last within their depth) and leave the scheduler's behavior out of the classifier.

test/ttlang/Dialect/TTL/Transforms/ScheduleOperations/schedule_fill_hazard.mlir
```diff
@@ -0,0 +1,75 @@
...
+    ttl.tile_store %o, %result_view[%i, %j] from dst[%c0] : !ttcore.tile<32x32, bf16>, tensor<1x1x!ttcore.tile<32x32, bf16>>
+    ttl.yield
+  } -> tensor<1x1x!ttcore.tile<32x32, bf16>>
+
+  ttl.cb_push %cb2 : <[1, 1], !ttcore.tile<32x32, bf16>, 1>
+
+  func.return %result : tensor<1x1x!ttcore.tile<32x32, bf16>>
+}
```
brnorris03: Since the argument for filtering on `DstAccessOpInterface` instead of `isa<TileFillOp>` is that it generalizes to uncategorized ops (but that is untested), can you please add a transpose companion here? Swapping `%f` for `ttl.tile_transpose %a_tile, %out_tile into dst[%c0] {ttl.transpose_output_cb_index = 2 : index}` reproduces the same hoist without the fix.

lib/Dialect/TTL/Transforms/TTLScheduleOperations.cpp
```diff
@@ -244,8 +244,15 @@ struct TTLScheduleOperationsPass
...
-        TileOpCategory cat = classifyTileOp(&op);
-        if (cat != TileOpCategory::Unknown) {
+        // Schedule every op that reads/writes DST, not just the ones with a
+        // known TileOpCategory. classifyTileOp is used only to order ops within
+        // a depth level (the sort key); using it as the inclusion filter would
+        // silently drop any DST op it doesn't categorize (e.g. tile_fill,
+        // tile_transpose, tile_reduce), leaving that op out of the WAW/WAR
+        // depth analysis so it strands ahead of an op recycling its DST slot.
+        // Every TTL_TileOp implements DstAccessOpInterface, so this includes
+        // all of them; an uncategorized one just sorts last within its depth.
```
brnorris03: Not quite accurate -- an op left out here ends up early because it isn't moved with the rest, not because it's missing from the WAW/WAR analysis. Also "just sorts last" isn't harmless for `tile_transpose` (see my note on the enum), and please avoid "silently".

```suggestion
        // An op excluded here is not moved with the rest, so it lands ahead of
        // ops reusing its DST slot; classifyTileOp leaves some uncategorized.
```

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,121 @@
...
+K = 3.0
+
+_DTYPE_PARAMS = [
+    pytest.param(torch.bfloat16, id="bf16"),
+    pytest.param(torch.float32, id="f32"),
+]
+# Error is dominated by the SFPU tanh/sigmoid approximation, not the storage
+# dtype, so one generous tolerance covers both. The constant-drop bug this guards
+# collapses the result magnitude by a factor of ~k, well outside this bound.
+_TOL = 5e-2
```
brnorris03: Please use dtype-specific tolerances like `_DTYPE_TOL` in `test_tensor_recurrences.py` -- I measured max relative error 1.3e-2 for bf16 but 3.1e-7 for f32, so a shared 5e-2 is orders of magnitude too loose on f32. The comment is inaccurate too (the dtype dominates, not the SFPU approximation), so I'd just drop it.

```suggestion
_DTYPE_TOL = {
    torch.bfloat16: dict(rtol=5e-2, atol=1.0),
    torch.float32: dict(rtol=1e-3, atol=1e-3),
}
```

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @ttl.datamovement()
+    def dm_read():
+        with a_dfb.reserve() as a_blk:
+            ttl.copy(a[0, 0], a_blk).wait()
+
+    @ttl.datamovement()
+    def dm_write():
+        with out_dfb.wait() as blk:
+            ttl.copy(blk, out[0, 0]).wait()
```
brnorris03: This is duplicated test code -- the two kernels are identical apart from the fill `dtype`. Suggest a single `make_fill_inline_kernel(fill_dtype)` factory closing over the dtype, parameterized the way the dtype tests in `test_tensor_recurrences.py` are; that drops the `_FILL_INLINE` dict too.

test/python/test_dst_fill_clobber.py
```diff
@@ -0,0 +1,121 @@
...
+    a = _inputs(dtype)
+    expected = _golden(a, K).to(dtype)
+
+    a_t = to_dram(a, device)
+    out_t = to_dram(torch.zeros(32, 32, dtype=dtype), device)
+
+    _FILL_INLINE[dtype](a_t, out_t)
+
+    result = ttnn.to_torch(out_t)
+    assert_allclose(result.float(), expected.float(), rtol=_TOL, atol=_TOL)
```
brnorris03: And here to match.

```suggestion
    assert_allclose(result.float(), expected.float(), **_DTYPE_TOL[dtype])
```

docs/development/DST_Utilization.md
jackzhangTT: sorry I put the wrong one oops

test/python/test_dst_fill_clobber.py
brnorris03: I think this suggestion wasn't completely right, but something like this -- otherwise things look good in the latest, as long as CI passes.

brnorris03 (approved):
Just needs a test fix to get CI passing, otherwise LGTM!

PR #777, opened 2026-07-28
[frontend] Support statically indexed DFB collections [dfb-reuse-1]

brnorris03:
Chain: #704 -> #777 (this PR) -> #778 -> #775

### Problem description

Composed operations can create repeated groups of dataflow buffers, but the frontend requires every DFB declaration and argument to be written individually. This makes repeated composition verbose and prevents collections from using the same static resource handling as individual DFBs.

### What's changed

Support non-empty list and tuple literals and statically evaluated list comprehensions built from `ttl.make_dfb` or `ttl.make_dataflow_buffer_like`. Collection elements can be referenced by non-negative integer literal index, destructured into named DFBs, and passed as composed-operation arguments. The frontend expands each element into an ordinary static DFB capture before the kernel split and diagnoses dynamic, negative, out-of-range, rebound, and ambiguous uses.

For example, a composed operation can declare its repeated DFBs as a collection and pass statically indexed elements to another operation:

```python
@ttl.operation()
def copy_stage(source: ttl.DFB, destination: ttl.DFB):
    source_block = source.wait()
    destination_block = destination.reserve()
    destination_block.store(source_block)


@ttl.operation(grid=(1, 1))
def chain():
    stage_dfbs = [
        ttl.make_dfb("bf16", shape=(1, 1), block_count=2)
        for _ in range(3)
    ]
    copy_stage(stage_dfbs[0], stage_dfbs[1])
    copy_stage(stage_dfbs[1], stage_dfbs[2])
```

Indexed elements remain explicit during operation composition, then become distinct static captures before kernel splitting and lowering. Dynamic subscripts remain invalid.

### Checklist

### DIFF
--- a/docs/sphinx/specs/TTLangSpecification.externalized.md
+++ b/docs/sphinx/specs/TTLangSpecification.externalized.md
@@ -175,6 +175,8 @@ Shape determines the shape of a *block* returned by one of the *acquisition func
 
 A dataflow buffer is constructed in the scope of an operation function but its object functions run on threads. Acquisition functions can be used with Python `with` statement, which will automatically release acquired blocks at the end of the `with` scope. Alternatively, if acquisition functions are used without the `with` the user must explicitly call a corresponding release function on the acquired block: `pop` for `wait` and `push` for `reserve`.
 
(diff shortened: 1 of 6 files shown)


brnorris03:
Closing this PR because the supported literal-indexed DFB collections do not provide enough benefit in current operations to justify the added frontend complexity. The DFB reuse stack does not require this syntax; its tests will use explicit DFB declarations instead.

PR #780, opened 2026-07-28
[ttl][pipes] Form grouped PipeNet transports and reuse NoC state [pipes-7]

brnorris03:
PipeNet PR tree:

```text
main (includes #756, #768, #704, #782, #778, #775, #783, #765, #700,
               and [pipes-8] #754)
  |-> [pipes-7] #780 (this PR)
  `-> [pipes-fabric] #734 (restack pending)
```

### Problem description

Repeated point-to-point PipeNet transfers program and synchronize one NoC write per logical transfer. The compiler needs a proven grouped schedule that reduces synchronization while preserving DFB lifetimes, receiver capacity, and resident NoC command state.

### What's changed

- Group eligible fixed-trip-count transfer loops and retain a scalar residual when the trip count is not divisible by the selected group size.
- Build a backend-independent PipeTransport plan for packetization, bounded receiver storage, capacity credits, completion, and source reuse.
- Replace proven private source and receiver DFB lifecycles with aligned per-core transport scratch and modulo-indexed destination groups.
- Reuse resident TTKernel NoC write state only when command fields are loop invariant and no intervening operation may reprogram or consume that state.
- Reuse #754's transfer expansion and payload geometry while keeping matching static and record-selected endpoints together until complete conversion.
- Reject DFB resource calculations that overflow their unsigned representation.

### Validation

New regressions cover grouped and residual schedules, exact L1 budget boundaries, DFB ownership rejection, multi-stream scratch allocation, target-specific NoC burst limits, NoC command effects, capacity-credit overlap, mixed static and record-selected endpoint representations, DFB allocation overflow, and preservation of selected-record address and synchronization tables. Device tests cover BF16 and FP32 computed-address and capacity protocols.

### Checklist

- [x] New and existing tests provide coverage for changes.


lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -2601,6 +2813,186 @@ LogicalResult buildPipeResourcePlan(
...
+    std::optional<int64_t> dynamicSlotCounterIndex;
+    if (stream.getSchedule() == PipeTransportSchedule::Overlapped) {
+      auto sendResource =
+          llvm::find_if(pipeResourcePlan.resources, [&](const auto &entry) {
+            return isa<PipeTransferSendOp>(entry.first) &&
+                   entry.second.transferNode == stream.getTransferNode();
+          });
+      assert(sendResource != pipeResourcePlan.resources.end() &&
+             "transport stream is missing sender resources");
+      if (sendResource->second.addressStorage.computedAddress) {
```
zoecarver, line 2860: Can this potentially re-use the wrong slot? Maybe add an assert that `initialSlot == 0`?

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1639,14 +1662,16 @@ def TTL_CBPushOp : TTL_Op<"cb_push", [MemoryEffects<[MemWrite]>]> {
 }
 
 def TTL_CBWaitOp : TTL_Op<"cb_wait",
-    [MemoryEffects<[MemRead]>, DeclareOpInterfaceMethods<ViewLikeOpInterface>]> {
```
zoecarver, line 1672: why the memory effect change here?

include/ttlang/Target/TargetInfo.h
```diff
@@ -0,0 +1,67 @@
...
+
+/// Return the maximum one-packet NoC transfer size for the module target.
+///
+/// Compile-only IR may omit `ttl.target_arch`; in that case this uses the
+/// minimum supported Wormhole B0/Blackhole value.
+inline int64_t getTargetNocMaxBurstBytes(Operation *op) {
+  if (isBlackholeTarget(op)) {
+    return kBlackholeNocMaxBurstBytes;
+  }
+  return kDefaultNocMaxBurstBytes;
```
zoecarver, line 42: should we maybe just assert we are wh?

zoecarver (approved):

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: Yes. Transport-owned scratch is zero-based: transport formation requires every use of the replaced DFB to be inside the candidate loop, so the grouped stream begins at slot 0. Added an assert.

include/ttlang/Target/TargetInfo.h
brnorris03: I don't think so, e.g., compile-only IR may omit `ttl.target_arch`, and the frontend preserves unknown future architecture names. In such cases this deliberately uses the conservative 8 KiB minimum for the currently supported Wormhole B0 and Blackhole targets and in future should be changed if needed to be correct for all targets.

include/ttlang/Dialect/TTL/IR/TTLOps.td
brnorris03: `ttl.cb_wait` is blocking synchronization even when its returned view is unused, but with just MemRead canonicalization would remove it if there were no users, which I think is not correct, it should block, whether the value is used or not. `cb_wait_dce.mlir` verifies that canonicalization and compute conversion retain it.

PR #783, opened 2026-07-28
[ttl][pipes] Add computed PipeNet addresses [pipes-2]

brnorris03:
PipeNet PR tree:

```text
main (includes #756, #768, #704, and #782; all merged)
  -> [pipes-2] #783 (this PR)
    -> [pipes-3] #765
      -> [pipes-4] #700
        -> [pipes-5] #740
          -> [pipes-6] #784
            |-> [pipes-7] #780
            |-> [pipes-8] #754
            `-> [pipes-fabric] #734
```

### Problem description

Receiver Address / Receiver Post (`RA/RP`) requires every receiver to publish its reserved dataflow buffer address before each transfer. Computed Address / Receiver Post (`CA/RP`) removes that address publication, but it is correct only when the sender can prove the address selected by every dynamic receiver reservation.

Lexical order and `PipeKey` do not identify transfers across functions, threads, branches, loops, or repeated uses. They can associate a send with the wrong reservation or allow one transfer's completion to satisfy another transfer's wait.

### What's changed

- Represent each transfer, receiver post, send, and completion wait explicitly.
- Resolve transfer values through supported control flow and tensor containers with `ValueOriginAnalysis`.
- Derive each receiver's periodic DFB address sequence from producer reserve/push order and select computed addressing only when the sender can materialize it.
- Permit producer reservations to reuse a DFB block after runtime capacity becomes available, including when a different kernel thread performs the consumer pop.
- Require multicast receiver address sequences to be pointwise equal. Multicast remains supported with either computed or receiver-published addresses.
- Isolate readiness and completion state by transfer and reject ambiguous or unsupported associations before lowering mutates IR.
- Add `--ttl-pipe-computed-addresses` and document both option states.

This contains the transfer-identity, completion-isolation, multicast-equality, and receiver-address portions of the original #700 review.

### Performance

The Blackhole benchmark uses BF16 32-by-32 tiles. Each result is the median of 50 warm dispatches after 5 warmups. Outputs were bit-exact, and the benchmark checks final MLIR to confirm protocol selection.

Point-to-point sender time:

| Transfers | `CA/RP` (us) | `RA/RP` (us) | Reduction |
| ---: | ---: | ---: | ---: |
| 64 | 54.457 | 59.688 | 8.76% |
| 256 | 217.340 | 238.946 | 9.04% |
| 1024 | 868.520 | 955.681 | 9.12% |
| 4096 | 3473.896 | 3825.039 | 9.18% |

Multicast sender time with one sender and three receivers:

| Transfers | `CA/RP` (us) | `RA/RP` (us) | Reduction |
| ---: | ---: | ---: | ---: |
| 64 | 60.290 | 68.358 | 11.80% |
| 256 | 240.942 | 273.790 | 12.00% |
| 1024 | 964.074 | 1094.733 | 11.94% |
| 4096 | 3853.720 | 4375.486 | 11.92% |

Both protocols retain Receiver Post synchronization. The reduction comes from removing receiver address-table writes and sender address-table reads.

Benchmark: https://gist.github.com/brnorris03/b9c79d994dcf9a6524a5a63a9ea2b976

### Validation

New regressions cover BF16 and FP32 `CA/RP` and `RA/RP` transfers, static and
periodic receiver-address sequences, cross-thread single-block DFB reuse,
multicast equality, completion isolation, both computed-address option states,
and diagnostics for unproven addresses.

A full-pipeline sweep selected `CA/RP` in all 288 compiling cases, including 234 cases with physical DFB index reuse.

### Checklist


lib/Dialect/TTL/Transforms/PipeLowering.cpp
```diff
@@ -290,6 +281,65 @@ buildAddressTableDestinationAddress(Location loc, const AddressTableInfo &info,
...
+      addByteOffset(loc, receiverAddress, info.staticTileByteOffset, rewriter);
+
+  Value repeatStride =
+      arith::ConstantIntOp::create(rewriter, loc, info.repeatStride, 32);
+  Value blockCount =
+      arith::ConstantIntOp::create(rewriter, loc, info.blockCount, 32);
+  Value nextSlotUnwrapped =
+      arith::AddIOp::create(rewriter, loc, currentSlot, repeatStride);
+  Value nextSlot =
+      arith::RemUIOp::create(rewriter, loc, nextSlotUnwrapped, blockCount);
```
zoecarver, line 336: is there a cheaper and still correct option maybe we could use?

lib/Dialect/TTL/Transforms/PipeGraph.cpp
```diff
@@ -3,180 +3,1172 @@
...
+    return failure();
+  }
+
+  for (PipeReceiverDFBNode &node : receiverDFBNodes) {
+    PipeReceiverDFBKey receiverDFB = node.receiverDFB;
+    LaunchNodeDomain receiverDomain =
+        getSingleLaunchNodeDomain(getLaunchNodeCoord(receiverDFB.receiver));
+
+    SmallVector<PipeTransferPostOp> posts;
+    mod.walk([&](PipeTransferPostOp postOp) {
```
zoecarver: non blocking: can we remove these two full module walks? I can imagine that getting expensive for multi-card.

lib/Dialect/TTL/Transforms/PipeGraph.cpp
```diff
@@ -3,180 +3,1172 @@
...
+            << candidates.sends.size() << " send definition(s) and "
+            << postCount << " receiver post definition(s)";
+        correspondenceResult = failure();
+        return;
+      }
+
+      for (std::size_t sendIndex = 0; sendIndex < candidates.sends.size();
+           ++sendIndex) {
+        PipeTransferSendOp sendOp = candidates.sends[sendIndex];
+        PipeTransferPostOp postOp = postsIt->second[sendIndex];
```
zoecarver, line 1141: Do I understand correctly that these are matched lexically? Eg whatever order they appear in the module walk is how we match them? 

Is there a stronger matching we could do?

brnorris03: Here `module.walk` establishes definition order, not program order (program order is the order in which events can execute within one kernel thread). Definition order is the deterministic lexical IR order of definitions for one event kind and `PipeKey`. Within one `PipeKey`, definition order defines the association: the first send corresponds to the first receiver post, the second send to the second post, and so on. Definition order is not itself treated as proof that arbitrary operations correspond.

`PipeGraph` first groups operations by `PipeKey`, so independent pipes may have different sender and receiver definition orders; [`overlap_distinct_slots_reversed_order`](https://github.com/tenstorrent/tt-lang/blob/20e98d73e312291e63e911b9ad94641391293260/test/ttlang/Dialect/TTL/Transforms/convert_pipe_ops_overlap.mlir#L184-L240) tests that case. For every association, the compiler separately proves that both sides execute equally often. The [wait-for graph](https://github.com/tenstorrent/tt-lang/blob/20e98d73e312291e63e911b9ad94641391293260/docs/development/PipeNets.md#L258-L279) then checks its program-order and synchronization dependencies. After the transfer nodes are built, graph analysis does not depend on module traversal order. Definition order only selects the association defined by the current DSL; all correctness checks still apply. tt-lang just can't express a different, order-independent association between multiple transfers sharing one `PipeKey` (at this time anyway).

In tt-lang now, source and destination callbacks identify a pipe relation but cannot name one shared transfer. The callbacks may therefore produce separate pipe-transfer references for the same transfer, so normal SSA def-use identity cannot match them. Order-independent matching within one `PipeKey` would require an explicit shared identifier or another unique semantic association, similar to tags in MPI for example (not suggesting we should add such, just explaining the missing semantic information in the current syntax).

lib/Dialect/TTL/Transforms/PipeLowering.cpp
brnorris03: `block_count` is a compile-time constant, so the downstream device compiler can strength-reduce this remainder; power-of-two counts become a mask. This also executes only once per send, where synchronization and the NoC transfer dominate. I would keep the direct modulo expression unless profiling shows it is significant, rather than expanding it into a less-obvious compare/subtract sequence here.

lib/Dialect/TTL/Transforms/PipeGraph.cpp
brnorris03: Good point. These walks were inside the receiver-DFB loop, so cost scaled with receiver DFB nodes times module size. PR734 (leaf pr in the stack) already collected these operations once for device-aware analysis, but I moved the core-only collector into this PR, no reason to postpone the fix.

zoecarver (approved):

PR #789, opened 2026-07-29
[ci] re-enable Galaxy hardware tests

brnorris03:
Re-enable Galaxy hardware tests in normal CI. Galaxy is online, but full 32-chip TTNN discovery still fails for tests that only need one or two chips, so this keeps single-device smoke on `TT_VISIBLE_DEVICES=0` and the shared serial multi-device phases on `0,1`.

The examples runner uses the same `HW_SERIAL_TEST_VISIBLE_DEVICES` setting because it invokes standalone scripts that call `ttnn.open_device(0)`; without the restriction they hit the same full-Galaxy `unordered_map::at` failure. The full-Galaxy tutorial xfail still applies when all Galaxy chips are visible.

Validated by CI run `30500635280`, including Galaxy hardware on `g08blx04` and N150 hardware.

### DIFF
--- a/.github/scripts/compile-and-run-examples.sh
+++ b/.github/scripts/compile-and-run-examples.sh
@@ -23,6 +23,8 @@
 #   bash .github/scripts/compile-and-run-examples.sh
 #
 # Optional first argument: repo root (default: current directory).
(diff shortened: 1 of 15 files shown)


jackzhangTT (approved):
LGTM

PR #790, opened 2026-07-30
[ttl] Add typed external function arguments

### LINKED ISSUE #767: [ttl] call_extern_func: support semaphore and raw tensor arguments
### Feature request

`ttl.call_extern_func` (added in the externFunc PR) currently supports a
limited set of argument kinds. Two argument types are not yet supported and
should be follow-ups:

- **Semaphores** as arguments. A kernel cannot currently pass a semaphore to an
  external C/C++ function via `func_args` / `template_args`.
- **Raw tensor operands** passed directly. Only dataflow-buffer CB-index
  passing is supported today (a DFB lowers to its i32 CB index); passing a
  tensor operand directly is not handled by `OpaqueCallLowering` and will fail
  at lowering.

### Currently supported (for reference)

| Kind | Accepts |
|------|---------|
| `template_args` | integer literal / module-level int constant; `ttl.get_dfb_id(dfb)` |
| `func_args` | scalar integer; scalar float (raw bit pattern); dataflow buffer (i32 CB index) |
| `include_paths` | list of `-I` paths for the JIT compiler |

### Scope

- Define how a semaphore is represented at the `opaque_call` boundary and lower
  it in `ConvertTTLToTTKernel` / `TTKernelToEmitC`.
- Define whether/how a raw tensor operand can be forwarded (e.g. as an L1
  pointer + shape metadata) and lower accordingly.
- Extend the Python `ttl.call_extern_func` handling and add lit + e2e coverage.

arichinsTT:
### Problem description

`ttl.call_extern_func` lacked stable C++ representations for DFBs, tensors, addresses, scalars, and synchronization resources. DFB arguments also lost allocation metadata and storage dependencies.

Fixes #767.

### What's changed

- Adds `ttl.dfb_descriptor(dfb)` with the finalized DFB index, pages per block, block count, and page size while preserving the DFB dependency.
- Lowers scalars, DFB indices, NOC-thread `TensorAccessor` values, raw tensor addresses for NOC, compute, and unified multi-thread kernels, and captured `ttnn.GlobalSemaphore` values to typed C++ arguments. Float template values use unsigned IEEE-754 f32 bit-pattern literals.
- Validates the complete argument plan before IR mutation and rejects non-byte-addressable or unrepresentable DFB descriptors.
- Uses target-supported physical DFB page sizes for L1 allocation while preserving exact scalar page sizes in runtime descriptors.

Coverage includes captured constant aliases, DFB allocation and reuse, subtile and nonzero multi-tile descriptor sizing, bf16/fp32 `TensorAccessor` and raw-address interop across NOC, compute, and unified multi-thread kernels, DRAM and interleaved or sharded L1, cached invocations with changed addresses, and invalid argument diagnostics.

The complete contract is documented in `docs/development/ExternalFuncInteropLowering.md`.

### Checklist


lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
```diff
@@ -1332,6 +1351,23 @@ struct OpaqueCallLowering : OpConversionPattern<OpaqueCallOp> {
         continue;
       }
 
+      // Bare tensor args default to TensorAccessor materialization.
```
brnorris03: I think the direct DFB must remain an operand of `ttl.opaque_call`, not only become `ttl.get_dfb_id`. An `i32` template argument cannot establish a storage dependency for finalization. The proposed descriptor template argument leaves `%dfb` in the ordered template operands; conversion after finalization creates a typed `ttkernel.dfb_descriptor` with its index and configuration fields.

test/python/test_call_extern_func.py
```diff
@@ -6,9 +6,10 @@
 End-to-end tests for ttl.call_extern_func.
 
 Covers:
-- Struct-based compute header with DFB IDs as template args (negate).
+- Struct-based compute header with DFB template args (negate).
```
brnorris03: I suggest rebasing #790 on #775 without requiring a duplicate `func_args=[dfb]` argument. Use `template_args=[ttl.dfb_descriptor(dfb)]`, so the descriptor DFB itself is a direct call operand. Raw physical-index use remains `ttl.get_dfb_id(dfb)` plus `dfb` in `func_args`, because the integer alone does not declare a storage dependency.

brnorris03 (commented):
I believe the direct DFB template-argument design needs to change before this becomes the final public external interface.

Current #790 source and lowering:

    template_args=[src]
    -> ttl.get_dfb_id %src
    -> ttl.opaque_call template_args(%id)
    -> my_shim<index>()

This leaves no direct DFB operand on `ttl.opaque_call`. The allocator therefore cannot treat the custom call as a storage dependency for `src`.

Proposed source and lowering:

    template_args=[ttl.dfb_descriptor(src)]
    -> ttl.opaque_call template_args(%src)
    -> ttkernel.dfb_descriptor carrying
       {index, pages_per_block, block_count, page_size_bytes}
    -> my_shim<ttlang::DFBDescriptor<...>>()

The descriptor is a C++ template type with `static constexpr` fields, so headers receive the DFB ID and declared configuration without manually repeated constants. It remains an operand of the TTL call for liveness, then conversion materializes it before TTKernel drops pages-per-block and block-count information.

Keep one ordered template-argument sequence; do not add an independent descriptor list. Existing integer template arguments and ordinary DFB function arguments remain unchanged. The detailed design note remains here: https://docs.google.com/document/d/1N8GswBLRAuIFXqe_EtbgBYme0Crvz5Qy7c-0vSFRqHM/edit?usp=sharing

docs/development/ExternalFuncInteropLowering.md
```diff
@@ -0,0 +1,131 @@
...
+| `ttl.get_dfb_id(dfb)` in `template_args` | Physical DFB index `uint32_t` literal | When the callee accesses DFB storage, the same DFB must declare a dependency through `func_args` or a `ttl.dfb_descriptor` template argument. |
+| DFB in `func_args` | Physical DFB index `uint32_t` parameter | Declares a direct DFB dependency. |
+| Integer or boolean in `template_args` | Signed integer or boolean constant | Must be compile-time evaluable. |
+| Float in `template_args` | Unsigned IEEE-754 f32 bit-pattern constant | Must be compile-time evaluable. |
+| Scalar in `func_args` | Lowered scalar parameter | Follows the TT-Metal kernel scalar convention. |
+| Base tensor in `func_args` | `TensorAccessor` parameter | Supports tiled bf16 and fp32 tensors only in NOC kernels. |
+| `ttl.raw_addr(base_tensor)` in `func_args` | `uint32_t` runtime tensor buffer address | Supports only NOC kernels; slices and derived tensor values are rejected. |
+| Captured `ttnn.GlobalSemaphore` | `uint32_t` address literal or parameter | The address is fixed for the compiled operation. |
```
zoecarver, line 29: This is great

lib/Dialect/TTL/Transforms/DFBAllocationLimits.cpp
```diff
@@ -41,18 +43,25 @@ std::optional<uint64_t> tryBudgetFromModule(ModuleOp module) {
 } // namespace
 
 FailureOr<uint64_t> getDFBAllocationSizeBytes(CircularBufferType type) {
-  int64_t totalElements = type.getTotalElements();
-  if (totalElements < 0) {
+  FailureOr<uint64_t> pagesPerBlock = getDFBPagesPerBlock(type);
```
zoecarver, line 48: Is this for sub-tile? Are we sure `getDFBPagesPerBlock` returns the right thing?

python/ttl/_src/ttl_ast.py
```diff
@@ -1470,21 +1582,82 @@ def _float_bits(py_float: float) -> int:
...
+                        return _boolean(bool(int(value_attr.value)))
+                    return _signed_integer(int(value_attr.value))
+            self._raise_error(
+                node,
+                "ttl.call_extern_func() template_args integer values must be "
+                "compile-time constants",
+            )
+        if isinstance(resolved_type, IndexType):
+            def_op = resolved.owner if hasattr(resolved, "owner") else None
+            if isinstance(def_op, arith.ConstantOp):
```
zoecarver, line 1664: Do we have tests for this case?

lib/Dialect/TTKernel/IR/TTKernelOps.cpp
```diff
@@ -631,8 +633,34 @@ void UnpackStallOnPackOp::getCanonicalizationPatterns(
 }
 
 ::mlir::LogicalResult OpaqueCallOp::verify() {
```
zoecarver, line 635: Is this duplicated with `OpaqueCallLowering::convertTemplateArg` and `DFBDescriptorAttr::verify`?

zoecarver (approved):
Glad to see this coming together

lib/Dialect/TTL/Transforms/DFBAllocationLimits.cpp
brnorris03: Yes. Block dimensions determine the page count; tile dimensions determine bytes per page. I added a subtile test confirming `[2, 3]` gives 6 pages and a `1x16` bf16 tile gives 32 bytes per page.

python/ttl/_src/ttl_ast.py
brnorris03: Added, thanks! The test found that a captured integer used through a local alias was incorrectly rejected, so I fixed that case, too.

lib/Dialect/TTKernel/IR/TTKernelOps.cpp
brnorris03: They check different stages: `OpaqueCallOp` restricts the allowed argument kinds, `DFBDescriptorAttr` validates finalized descriptor fields, and the lowering computes those fields from the TTL DFB and diagnoses failures such as unrepresentable dimensions or invalid page sizes.

lib/Dialect/TTL/Transforms/ConvertTTLToTTKernel.cpp
brnorris03: Implemented. The DFB remains in `template_dfb_operands` through allocation finalization. Lowering then replaces it with a `ttkernel.dfb_descriptor` containing the finalized index and configuration.

test/python/test_call_extern_func.py
brnorris03: Done. `ttl.dfb_descriptor(dfb)` now includes the dependency without a duplicate function argument. `ttl.get_dfb_id(dfb)` remains an index only, so the DFB must also be a function argument when the external code accesses its storage.

PR #823, opened 2026-08-07
[ttl] Add tensor-provided dynamic read indices

brnorris03:
### Problem description
Data movement kernels sometimes receive runtime indices in f32 or bf16 tensors, such as table-selection or position values. TTL can load those scalar values from dataflow buffers, but cannot convert them to MLIR index values because generic floating-point-to-integer conversion is not supported in data movement code.

### What changed

~321 LOC implementation (tests and documentation excluded).

- Adds `ttl.read_index` to load one f32 or bf16 element from a dataflow buffer and return it as an MLIR index.
- Defines finite, nonnegative, INT32_MAX-bounded input semantics with truncation toward zero; behavior outside that contract is undefined.
- Lowers the conversion by decoding the element's IEEE-754 representation with integer operations. Format parameters come from the MLIR float type, and the lowering does not use generic `arith.fptosi` or inspect SSA producers.
- Verifies the supported element types, `ttl.cb_wait` acquisition, data movement thread placement, and coordinate count.
- Exposes the Python operator, classifies it as data movement during atom splitting, and documents the compiler design in the DFB management document.

This supersedes Zoe Carver's initial implementation in commit [2f6349dda](https://github.com/tenstorrent/tt-lang/commit/2f6349dda82580f10e8735908b67de1fa2cb8a66).

### Example
If `index_tensor` contains `2.75`, `ttl.read_index` returns index `2`, which selects the third tensor slot:

```python
with index_dfb.reserve() as index_block:
    ttl.copy(index_tensor[0, 0], index_block).wait()

with index_dfb.wait() as index_block:
    slot = ttl.read_index(index_block, 0, 0)
    with weights_dfb.reserve() as weights_block:
        ttl.copy(
            weights[slot * SLOT_TILES : (slot + 1) * SLOT_TILES, 0:1],
            weights_block,
        ).wait()
```

### Validation
- MLIR tests cover tiled f32 and row-major bf16 acceptance, each verifier error independently, and exact format constants during integer-only conversion.
- The atom-split test verifies that `ttl.read_index` remains with its dataflow buffer acquisition and transfer.
- The device pytest verifies exact output for below-one, right-shift fractional, and left-shift decode branches in bf16 and fp32, including tensor value 2.75 truncating to slot 2.

### Checklist

### DIFF
--- a/docs/development/DFBManagement.md
+++ b/docs/development/DFBManagement.md
@@ -1791,6 +1791,12 @@ The verifiers enforce:
 4. Scalar read and write types match the block element type.
 5. Only `f32` and `bf16` block elements are accepted.
 
+This is a scalar-access lowering restriction, not a DFB storage restriction.
+The lowering maps the 32-bit and 16-bit IEEE-754 representations to TTKernel
+L1 integer loads and stores. `ttl.read_index` decodes those representations
+with integer operations because noc kernels do not support generic
+floating-point-to-integer conversion.
+
 `ttl.read_index` truncates fractional values toward zero and returns an MLIR
 index. Its source must be finite, nonnegative, and no greater than INT32_MAX;
 behavior is undefined otherwise. The element-access operations carry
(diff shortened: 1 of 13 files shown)


arichinsTT (commented):
Would we be able to keep this operation in the frontend, emiting a read raw element op with arith casting and bounds checks? it would make it easier to switch to native integer tensor support later and have a simpler lowering, using existing ops

brnorris03:
> Would we be able to keep this operation in the frontend, emiting a read raw element op with arith casting and bounds checks? it would make it easier to switch to native integer tensor support later and have a simpler lowering, using existing ops

As we discussed separately, this will be a follow on change.

zoecarver (approved):
I know Alex wanted to do this without a new op and just by emitting casts in operators.py, happy either way.

arichinsTT (approved):

PR #825, opened 2026-08-08
[ttl] Add portable logical kernel selection

brnorris03:
### PR stack

```
main
  |
  `-- #825 Portable logical-kernel selection (this PR)
        |-- #827 Portable operation runtime resources
        `-- #830 Typed DFB effects for opaque calls
```

### Problem description

A unified `@ttl.operation` is written as one Python function, but the compiler emits separate compute and data-movement kernels. For ordinary TT-Lang statements, visible operation semantics and data dependencies determine which kernel executes each statement. An opaque `ttl.call_extern_func` does not provide that information, and one external call may need to execute in more than one logical kernel.

A dataflow-buffer `push()` or `pop()` has the same ambiguity when its associated block is otherwise unused. Physical processor names are not a suitable selector API because they vary by target and cannot consistently identify multiple operation-level kernels of the same kind.

This PR provides explicit, target-independent logical-kernel selectors for these cases. A logical kernel identifies an operation-level compute or data-movement kernel independently of its physical backend assignment. The identity is retained through splitting and specialization.

The runtime requires a complete backend kernel set, so the compiler can also emit empty kernels. These synthesized kernels receive stable logical identities, ensuring every runtime `KernelSpec` can be addressed without relying on descriptor order or physical processor names.

### What's changed

- Add `KernelKind` and operation-local `Kernel` selectors.
- Add `kernel=` to `call_extern_func`, `TensorBlock.push`, and `TensorBlock.pop`.
- Allow one selector, a `KernelKind` union using `|`, or a nonempty tuple for external calls; require at most one selector for push/pop.
- Split unified operations by logical kernel without changing external-call lowering or DFB synchronization operations.
- Preserve verified logical-kernel identity through specialization and expose it on `KernelSpec`.
- Bind factory-captured kernel handles in place during operation registration.
- Enforce target capacity separately for compute and data-movement kernels, including explicit operations that previously accepted unsupported kernel mixes with three total threads.
- Document external-function arguments, kernel selection, include directories, and DFB synchronization ownership in the Sphinx user guide.

### Checklist

### DIFF
--- a/python/sim/kernel.py
+++ b/python/sim/kernel.py
@@ -4,8 +4,11 @@
 
 """Inert logical-kernel selectors for simulator API compatibility."""
 
+from __future__ import annotations
+
 from dataclasses import dataclass
 from enum import Enum, auto
+from typing import Tuple
 
 
 class KernelKind(Enum):
@@ -18,6 +21,11 @@ class KernelKind(Enum):
 
     COMPUTE = auto()
     DATA_MOVEMENT = auto()
+
+    def __or__(self, other: object) -> Tuple[KernelKind, KernelKind]:
+        if not isinstance(other, KernelKind):
+            return NotImplemented
+        return (self, other)
 
 
 @dataclass(frozen=True)
--- a/python/sim/kernel.py
+++ b/python/sim/kernel.py
@@ -29,8 +29,7 @@ class Kernel:
     def __post_init__(self) -> None:
         if not isinstance(self.kind, KernelKind):
             raise TypeError(
-                "Kernel kind must be a KernelKind, got "
-                f"{type(self.kind).__name__}"
+                "Kernel kind must be a KernelKind, got " f"{type(self.kind).__name__}"
             )
 
 
--- a/python/sim/kernel.py
+++ b/python/sim/kernel.py
@@ -1,0 +1,37 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Inert logical-kernel selectors for simulator API compatibility."""
+
+from dataclasses import dataclass
+from enum import Enum, auto
+
+
+class KernelKind(Enum):
+    """A target-independent kernel class."""
+
+    @staticmethod
+    def _generate_next_value_(name, start, count, last_values):
+        del start, count, last_values
+        return name.lower()
+
+    COMPUTE = auto()
+    DATA_MOVEMENT = auto()
+
+
+@dataclass(frozen=True)
+class Kernel:
+    """A logical-kernel declaration ignored by simulator execution."""
+
+    kind: KernelKind
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.kind, KernelKind):
+            raise TypeError(
+                "Kernel kind must be a KernelKind, got "
+                f"{type(self.kind).__name__}"
+            )
+
+
+KernelSelector = KernelKind | Kernel
--- a/python/sim/kernel.py
+++ b/python/sim/kernel.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum, auto
-from typing import Tuple
 
 
 class KernelKind(Enum):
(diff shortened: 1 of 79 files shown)


test/python/explicit_kernel_selection.py
```diff
@@ -0,0 +1,83 @@
...
+compute_kernel = ttl.Kernel(ttl.KernelKind.COMPUTE)
+reader_kernel = ttl.Kernel(ttl.KernelKind.DATA_MOVEMENT)
+
+
+@ttl.operation(grid=(2, 1))
+def explicit_selected_add(inp, out):
+    input_dfb = ttl.make_dataflow_buffer_like(inp, shape=(1, 1), block_count=1)
+    output_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=1)
+
+    @ttl.compute(kernel=compute_kernel)
```
arichinsTT, line 30: this syntax feels klunky and over-verbose/duplicate information in this situation.  Can we derive the relevant and necisary information for threaded operations and keep this feature within unified kernels?

arichinsTT (commented):
having a test for mismatch kernel kind and kernel decorator

test/python/explicit_kernel_selection.py
arichinsTT: offline, this is a use case that is not required but potentially possible with atom usage

test/python/kernel_selection.py
```diff
@@ -0,0 +1,84 @@
...
+            FAKE_HEADER,
+            "reader_entry",
+            template_args=[5],
+            func_args=[ttl.raw_addr(inp)],
+            kernel=reader,
+        )
+    ttl.call_extern_func(
+        FAKE_HEADER,
+        "shared_entry",
+        kernel=(KernelKind.COMPUTE, reader),
```
zoecarver: could we spell this `KernelKind.COMPUTE | KernelKind.DATA_MOVEMENT`?

arichinsTT: ^

brnorris03: Yep, added this spelling. Tuples are still supported when the selection includes named `Kernel` handles.

brnorris03:
> having a test for mismatch kernel kind and kernel decorator

Done in `test_explicit_kernel_decorator_rejects_wrong_named_kernel_kind` in test/python/
  atom/test_explicit_kernel.py (https://github.com/tenstorrent/tt-lang/pull/825/changes#diff-f62210ce9cc294ffd9516ec2beb2819bf3cb45099dc710f8da3f448922c4ba46R19-R56)

zoecarver (approved):

python/ttl/kernel.py
```diff
@@ -0,0 +1,334 @@
...
+    try:
+        nonlocal_captures = inspect.getclosurevars(function).nonlocals
+    except (TypeError, ValueError):
+        return base_identity
+
+    encoded_captures = []
+    for name, value in sorted(nonlocal_captures.items()):
+        encoded = _encode_identity_literal(value)
+        if encoded is None:
+            continue
```
zoecarver: I think this should fail rather than dropping unecnodable? This might lead to conflicting distinct kernels

python/sim/kernel.py
```diff
@@ -0,0 +1,43 @@
...
+
+"""Inert logical-kernel selectors for simulator API compatibility."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum, auto
+
+
+class KernelKind(Enum):
```
FConstantinos: There is a KernelType class in blockstate.py. Is it possible to avoid duplication here?

arichinsTT (approved):

python/ttl/kernel.py
brnorris03: Good catch, fixed!

python/sim/kernel.py
brnorris03: Yes, thanks for catching that! `KernelKind` is now the simulator’s single enum; `KernelType` and its abbreviated `DM` member are removed.

FConstantinos (approved):

PR #827, opened 2026-08-08
[ttl] Add portable operation runtime resources [fabric-1]

brnorris03:
Stacked PR tree:

```text
main
`-> #827 operation runtime resources (this PR)
    `-> #734 multidevice fabric PipeNets
```

### Problem description

TT-Metal programs distinguish reusable program structure, including JIT definitions and program semaphore layout, from per-core runtime arguments that may change for each dispatch. Main's `@ttl.operation` runner constructs this program only from compiled kernels, DFB descriptors, tensor addresses, and compiler-owned PipeNet resources. It has no operation-level input for caller-owned program semaphores or JIT definitions, per-logical-kernel runtime arguments, or owners of resources referenced during execution.

External kernel ABIs need these inputs when their program structure or dispatch values depend on the current tensors, device, launch range, or compiler-reserved semaphore IDs. The caller interface must address logical kernels because generated descriptor indices and processor assignments are target artifacts, and core specialization may materialize one logical kernel as several descriptors.

### What's changed

This change adds a typed resource-planning boundary between operation compilation and TTNN descriptor construction. A per-execution factory returns resources keyed by logical `KernelKind` or explicit `Kernel`; the runner validates the complete plan before materialization and separates cache-affecting structure from current dispatch values and ownership.

Decorated operations and emitted Python runners use the same planner and materializer. Per-core values are routed to specialized descriptors, invalid resources fail before TTNN program descriptor construction, and retained owners are committed only after successful execution.

### Checklist

- [x] New and existing tests provide coverage for changes.
- [x] Public and developer documentation is updated.


zoecarver (approved):
Fantastic! Thank you for formalizing this!!

PR #830, opened 2026-08-08
[ttl] Describe opaque-call DFB effects [dfb-reuse-1]

### LINKED ISSUE #806: Represent external-call DFB dependencies and protocol effects
### Problem

`ttl.call_extern_func` can pass DFBs as C++ arguments, and `ttl.opaque_call` preserves them as direct DFB operands. This models opaque storage access for the duration of the call.

The DSL and IR cannot represent dependency-only DFBs, statically unknown DFB access sets, or ordered reserve, push, wait, and pop operations implemented inside an external callee.

Generic read/write memory effects are insufficient. DFB lifecycle analysis requires protocol operation ordering, tile counts, and pointer ownership.

### Kimi V3 requirement

This functionality is required for Kimi V3 production integration. Kimi V3 uses external kernels that perform DFB protocol operations internally. Without typed effects, the compiler cannot prove bounded lifecycles, validate ownership, or safely reuse physical DFB allocations.

Artificial DSL operations that imitate hidden C++ behavior are not an acceptable contract because they do not represent executed operations and can become inconsistent with the external implementation.

### Required design

- Add an explicit DFB dependency operand segment. Direct DFB function arguments are included automatically; dependency-only DFBs do not change the C++ signature.
- Add an ordered, typed protocol-effect summary. Each effect references a DFB dependency and identifies reserve, push, wait, or pop with its tile count.
- Require effects to complete before the call returns. Outstanding asynchronous work must have explicit completion semantics.
- Resolve DFB identities and tile counts before allocation. Diagnose malformed or contradictory summaries before IR mutation.
- Represent unknown DFB access explicitly and disable physical-index reuse in every affected allocation scope.
- Diagnose `ttl.get_dfb_id` use unless its source DFB is included in the dependency set.
- Define one declarative interface consumed by concurrent DFB liveness, physical allocation, and SPSC verification.
- Preserve metadata through composition, unified-operation splitting, logical-kernel selection, and core specialization.
- Never infer effects from callee names, headers, emitted C++, or library-specific markers.

### Illustrative DSL

The final spelling remains a design decision. Intended semantics:

```python
ttl.call_extern_func(
    "dfb_protocol.hpp",
    "consume_and_publish",
    func_args=[input_dfb, output_dfb],
    dfb_dependencies=[scratch_dfb],
    dfb_effects=[
        ttl.DFBEffect.wait(input_dfb, tiles=1),
        ttl.DFBEffect.reserve(output_dfb, tiles=1),
        ttl.DFBEffect.push(output_dfb, tiles=1),
        ttl.DFBEffect.pop(input_dfb, tiles=1),
    ],
)
```

`scratch_dfb` is visible to compiler analyses but is not emitted as a C++ argument. Effects describe operations executed inside the call; they do not emit duplicate protocol operations.

### Validation

- ODS round-trip tests cover dependencies, ordered effects, and unknown access.
- Invalid tests reject non-dependency references, invalid tile counts, unresolved identities, and invalid lifecycle summaries.
- Frontend tests cover direct arguments, dependency-only DFBs, typed effects, and unknown access.
- Dependency-only DFBs extend liveness without appearing in generated C++ arguments.
- Complete summaries permit reuse only when existing ordering and ownership analyses prove it safe.
- Incomplete summaries remain conservative and produce precise diagnostics when bounded lifecycle proof is required.
- Existing calls without annotations retain unchanged generated C++.
- A device test executes an external operation with hidden DFB protocol operations and validates its result against a golden implementation.

### Non-goals

- Parsing external C++ to infer effects.
- Architecture-specific processor names.
- Callee or header names as semantic identifiers.
- Replacing visible DFB protocol operations with opaque effects.

brnorris03:
### What's changed

~1,600 LOC implementation (tests and documentation excluded).

Describe ordered DFB effects for external kernels, including bounded repetition and static count expressions, so allocation and protocol analyses preserve externally implemented queue operations.

Keep strict DFB verification as the default while allowing controlled diagnostic qualification through `TTL_RELAX_DFB_SPSC`; PipeNet endpoint, correspondence, and schedule checks remain enabled.

### Validation

- Blackhole BF16/FP32 device coverage on DRAM and L1
- Focused frontend and MLIR coverage
- Complete stacked MLIR suite
- Pre-commit


python/ttl/_src/atom_split.py
```diff
@@ -346,6 +346,8 @@ def _classify_ttl_call(
         outer = func.value
         if isinstance(outer.value, ast.Name) and outer.value.id == "ttl":
             namespace = outer.attr
+            if namespace == "DFBEffect":
+                return None
```
jackzhangTT: I feel like DFBEffect shouldn't actually count as a namespace as it's more of an "effect". For now this is okay but I think we should revisit this at some point and clean it up.

arichinsTT (approved):

jackzhangTT (approved):
LGTM

python/ttl/_src/atom_split.py
brnorris03: Fixed in `4c952aa10`: `DFBEffect` now uses the shared registry as `_Placement.CONTROL`.

PR #858, opened 2026-08-11
Fixes for Blacksmith

FConstantinos:
## Summary

Unblocks kernels generated by [tt-blacksmith-codegen](https://github.com/tenstorrent/tt-blacksmith-codegen/tree/umales/initial_triton_to_ttlang/kernels) by aligning the simulator and Python frontend with the TT-Lang/TTNN behavior those workloads rely on.

This is the foundational PR for the Blacksmith fix stack. More focused compiler and simulator changes are split into the dependent PRs listed below.

### Simulator correctness and TTNN compatibility

- Track logical tensor shapes separately from padded physical storage, and run golden-backed operations on logical data.
- Preserve dtype, layout, memory configuration, and relevant storage metadata when tensors and blocks are created or transformed.
- Add explicit simulator implementations for `to_layout`, `reshape`, `copy`, `slice`, and `deallocate`, including rejection of reads from deallocated tensors.
- Make golden-operation dispatch behavior-driven: handwritten simulator operations take precedence, supported TTNN operations are wrapped dynamically, and unsupported names remain unavailable.
- Improve DFB/block handling for rank-1 and higher-rank tile grids, trailing sub-slabs of higher-rank backing tensors, and `Block.to_list`/`Block.from_list` metadata preservation.
- Give unsupported scalar/block expressions a direct diagnostic that points users to `ttl.block.fill()`.
- Allow `ttl.block.fill()` to declare its output dtype while retaining a sufficiently wide host backing type.

### Python frontend and public API

- Expose the implemented block operations through the public `ttl.block` namespace, matching specification revision 0.17.
- Treat captured scalar tuples/lists as compile-time shape or axis values, the same as equivalent inline literals.
- Support `dims=1` and `dims=2` for `ttl.node()` and `ttl.grid_size()`; the one-dimensional form follows the specification's node-flattening order.
- Keep diagnostics for unsupported grid dimensionalities explicit and covered by tests.

## Testing

The PR adds focused regression coverage for:

- the public `ttl.block` namespace and operation signatures;
- captured shape sequences in compiled kernels;
- one- and two-dimensional node/grid queries;
- DFB rank, tile-grid, storage-metadata, and scalar-operand behavior;
- logical versus padded tensor shapes;
- dtype/layout/device/memory-config propagation;
- handwritten and golden-backed simulator operations;
- slicing and tensor deallocation.

The full repository CI exercises pre-commit, compiler/build tests, simulator tests on Linux and macOS, hardware tests, wheel construction, tutorials, and security scanning.

## Follow-up PRs

The remaining Blacksmith fixes are intentionally split into a reviewable stack:

- #938 — explicit reduction shape arguments
- #940 — rank-N reductions
- #941 — module-level kernel helpers
- #948 — block squeeze/unsqueeze views
- #949 — batched block matmul
- #950 — reject mixed block data types in the simulator
- #953 — exact store-shape enforcement across compiler and simulator

Support for higher-dimensional submeshes in `ttl.node()` and `ttl.grid_size()` is tracked separately in #951.


python/ttl/operators.py
```diff
@@ -729,56 +729,72 @@ def node(*, dims):
...
+    x = ttl.core_x()
+    if dims_val == 2:
+        return (x, ttl.core_y())
+    # Flattening folds the highest-rank dimension into the one below it, so x
+    # varies fastest and the row stride is the grid's column count.
+    cols = _get_current_grid()[0]
+    ctx = x.type.context
+    stride = arith.ConstantOp(IndexType.get(ctx), cols).result
+    row_base = arith.MulIOp(ttl.core_y(), stride).result
+    return arith.AddIOp(row_base, x).result
```
zoecarver: Is this correct logic? I don't actually know how dims=1 is supposed to work. Regardless let's assert dims is one or two. Also can we maybe support dims on larger sub mesh ASAP before fabric pipes?

python/ttl/operators.py
```diff
@@ -729,56 +729,72 @@ def node(*, dims):
...
-            f"grid_size() currently only supports dims=2, got dims={dims_val}. "
+            f"grid_size() currently only supports dims=1 and dims=2, got dims={dims_val}. "
             "Multi-dimensional grids are not yet supported."
         )
     # grid is stored as (cols, rows) = (x, y), matching tt-metal convention
-    return _get_current_grid()
+    cols, rows = _get_current_grid()
+    if dims_val == 2:
+        return (cols, rows)
+    return cols * rows
```
zoecarver, line 796: same

python/ttl/operators.py
FConstantinos: fixed, thanks!

The check happens in previous lines (~740)

python/ttl/operators.py
FConstantinos: fixed

python/ttl/operators.py
FConstantinos: RE: 

> Also can we maybe support dims on larger sub mesh ASAP before fabric pipes?

I've opened an issue for that:

https://github.com/tenstorrent/tt-lang/issues/951

python/sim/block.py
```diff
@@ -22,7 +22,7 @@
 from .dfb import Block, check_same_layout, track_source_blocks, _dry_run_result
 from .blockstate import BlockAcquisition
 from .kernel import KernelKind
-from .ttnnsim import ROW_MAJOR_LAYOUT, Tensor
+from .ttnnsim import ROW_MAJOR_LAYOUT, Tensor, bfloat16, _promote_dtype
```
brnorris03: nit: `_promote_dtype` probably shouldn't be private if it's used outside its module.

python/sim/block.py
```diff
@@ -235,13 +235,15 @@ def broadcast(
 def fill(
     value: float,
     shape: Tuple[int, ...],
+    dtype: Any = None,
```
brnorris03: Would it be possible to have dtype be an enum (don't we already have that)?

python/sim/dfb.py
```diff
@@ -814,6 +870,18 @@ def _binary_op(
 
         Tracks wait() Compute blocks that contribute to the result.
         """
+        # Reject a non-Block operand here rather than letting it reach the
+        # layout check, which would surface as an attribute error naming
+        # neither the operand nor the operation.  The compiler accepts a
+        # compile-time scalar for `*` via ttl.mul_unary_const; until that
+        # form is specified and implemented here, say so. See issue #869.
```
brnorris03: Nit: kind of convoluted comment, could be simplified to a # TODO(#869): add support for non-Block operands

python/sim/ttnnsim.py
```diff
@@ -1957,6 +1965,47 @@ def __init__(
...
+
+    def device(self) -> object:
+        """Return the device handle this tensor was created on.
+
+        Mirrors ``ttnn.Tensor.device()``, which a kernel calls to route a
+        derived tensor to the same device as its input.  Only the creation
+        entry points (``from_torch`` / ``rand`` / ``empty`` / ``zeros``) carry a
+        handle; the simulator does not propagate residency through operations,
+        so a tensor produced by arithmetic or slicing has none and this raises
+        rather than name a device the simulator did not place it on.
```
brnorris03: not a blocker, but curious about the ``ttnn.Tensor.device()`` vs the usual markdown `ttnn.Tensor.device()`, why the double ``?

brnorris03 (approved):
Just a couple of small things, mostly looks good to me!

python/sim/block.py
FConstantinos: Addressed in 90f33edd4: renamed the helper to public promote_dtype and updated all callers.

python/sim/block.py
FConstantinos: Addressed in d7c3f166b: block.fill now uses the existing DType annotation and the specification default, bfloat16. The simulator models ttnn dtype values as torch.dtype plus the BFP8 sentinel, so DType is the precise existing type; introducing a separate enum would be a broader simulator API change.

python/sim/dfb.py
FConstantinos: Addressed in 15d6eb24e: simplified this to TODO(#869): Add support for non-Block operands.

python/sim/ttnnsim.py
FConstantinos: Addressed in 868de3927: removed the ambiguous inline markup and described this as the device method on ttnn.Tensor.

PR #885, opened 2026-08-12
[ttl] Specialize CBs after per-core specialization

jackzhangTT:
Annotate surviving CB uses after core specialization so runtime descriptors and backing allocations only cover cores that need each slot, with a conservative fallback when metadata is missing.

### Problem description
`ttkernel-specialize-cores` clones kernels and folds dead coordinate branches, but the host still built CB descriptors for every physical slot on the whole launch grid. Unused slots still occupied the static CB region, so specialized programs paid L1 for CBs their clone never touched.

### What's changed
- Add `ttkernel-annotate-cb-use` after specialize + canonicalize/CSE. Each function gets `ttl.used_cb_indices` from surviving `get_compile_time_arg_val` ops below `ttl.base_cta_index`, including scalar IDs passed to `opaque_call`. `func.call` uses are propagated to callers; unanalyzable callees take every CB in that range.
- Runtime reads that attribute onto `KernelSpec`, unions uses per core, and emits descriptors / pipe backing only for `(CB, core)` pairs that remain live. Unannotated kernels keep the old whole-grid behavior (`None` vs `[]`).
- Emitted runners serialize `KERNEL_USED_CB_INDICES` so offline replay matches.

CB IDs are unchanged. L1 is saved because Metal packs configured descriptors, not `0..max_index`.

### Ticket
N/A

### Checklist
```

### CODE REVIEW

File: include/ttlang/Dialect/TTL/IR/TTL.h [hunk 1]
```diff
@@ -165,6 +165,11 @@ constexpr llvm::StringLiteral
 /// CTA layout is [CBs, TAs], so this equals the number of CBs.
 constexpr llvm::StringLiteral kBaseCTAIndexAttrName("ttl.base_cta_index");
 
+/// Function attribute recording physical CB indices referenced by the final
+/// TTKernel body after per-core specialization and canonicalization.
+constexpr llvm::StringLiteral
+    kUsedCBIndicesAttrName("ttl.used_cb_indices");
+
```
brnorris03: C++ writes this attribute and Python reads it, so its name should have a single definition rather than duplicated string literals in multiple places. Please bind the renamed (see comment about CB->DFB below) `kUsedDFBIndicesAttrName` in `TTLModule.cpp` as `_ttl_ir.USED_DFB_INDICES_ATTR`, following `PIPE_COMPUTED_ADDRESS_DFB_INDICES_ATTR`, and use that constant in `ttl_api.py`.

include/ttlang/Dialect/TTL/Passes.td
```diff
@@ -1200,6 +1200,29 @@ def TTKernelSpecializeCores
...
+    : Pass<"ttkernel-annotate-cb-use", "::mlir::ModuleOp"> {
+  let summary = "Record physical CB indices used by each TTKernel function";
+  let description = [{
+    Runs after per-core specialization and canonicalization, when untaken
+    coordinate branches have been removed. For each function, records the
+    sorted physical CB indices reached through surviving
+    `ttkernel.get_compile_time_arg_val` operations below
+    `ttl.base_cta_index` in `ttl.used_cb_indices`. The compile-time ABI places
+    every CB before tensor-accessor arguments, including CB ids passed to
+    opaque externs as scalar values.
```
brnorris03: nit: "untaken coordinate branches" sounds kind of weird to me, could reword. 

Also we have switched to using DFB, not CB in all new code (old code still needs updating), so I'd change the pass and description accordingly.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
```diff
@@ -0,0 +1,119 @@
...
+              func->getAttrOfType<IntegerAttr>(kBaseCTAIndexAttrName)) {
+        moduleCBCount = std::max(moduleCBCount, attr.getInt());
+      }
+    }
+
+    // Populate usedCBs set and callees map
+    for (func::FuncOp func : module.getOps<func::FuncOp>()) {
+      Operation *key = func.getOperation();
+      int64_t cbCount = getCBCount(func, moduleCBCount);
+      func.walk([&](ttk::GetCompileArgValOp op) {
```
brnorris03: This can miss a DFB whose only surviving use on a specialized core is dprint. `ttl-lower-dprint-to-emitc` runs before specialization and replaces the typed DFB operand with `get_compile_time_arg_val(...)` text in `emitc.verbatim`; this walk cannot see that use, so the runtime can omit the DFB descriptor on that core. Should preserve a typed DFB-use marker through dprint lowering and add a frontend regression for this case. (I have often also forgotten the dprint since it's not always used)

lib/Dialect/TTL/Pipelines/TTLPipelines.cpp
```diff
@@ -97,6 +97,7 @@ void createTTLToTTKernelPipeline(OpPassManager &pm,
     pm.addPass(createTTKernelSpecializeCores());
     pm.addPass(createCanonicalizerPass());
     pm.addPass(createCSEPass());
+    pm.addPass(createTTKernelAnnotateCBUse());
```
brnorris03: Suggest registering `ttkernel-specialize-cores,canonicalize,cse,ttkernel-annotate-dfb-use` as a sub-pipeline and invoking it here and from `ttl_api.py`. Otherwise someone may reorder accidentally. And remember we have three places to update for pipeline changes -- c++, python, and me2e tests.

test/python/test_kernel_runner.py
```diff
@@ -515,6 +522,251 @@ def test_build_cb_descriptors_preserves_subtile_geometry(monkeypatch):
...
+        path="/tmp/specialized.cpp",
+        thread_type="compute",
+        tensor_indices=[],
+        config=object(),
+        core_ranges=core_ranges,
+        used_cb_indices=used_cb_indices,
+    )
+
+
+def test_specialized_cb_descriptor_follows_active_kernel_core(monkeypatch):
```
brnorris03: Can you add a hardware pytest (non-mocked) that runs a specialized kernel where a DFB is live on only a subset of cores?

include/ttlang/Dialect/TTL/Passes.td
```diff
@@ -1200,6 +1200,29 @@ def TTKernelSpecializeCores
...
+
+    Direct function calls are analyzed transitively. An unresolved call is
+    handled conservatively by recording every CB compile-time argument below
+    `ttl.base_cta_index`.
+  }];
+
+  let dependentDialects = [
+    "::mlir::func::FuncDialect",
+    "::mlir::tt::ttkernel::TTKernelDialect"
+  ];
```
brnorris03: This pass only reads func/ttkernel operations and sets a builtin attribute; it creates no dialect operations, so `dependentDialects` should be empty.

test/ttlang/Dialect/TTKernel/Transforms/specialize_cores_dfb_use_elision.mlir
```diff
@@ -0,0 +1,39 @@
...
+
+// CHECK-NOT: func.func @conditional_cb_user()
+// CHECK-LABEL: func.func @conditional_cb_user_c0_0
+// CHECK-SAME: ttl.core_coord = {{\[\[}}0, 0]]
+// CHECK-SAME: ttl.used_cb_indices = array<i32: 0>
+// CHECK: ttkernel.get_compile_time_arg_val(0)
+// CHECK: ttkernel.cb_wait_front
+// CHECK-NOT: scf.if
+// CHECK-LABEL: func.func @conditional_cb_user_c0_1
+// CHECK-SAME: ttl.core_coord = {{\[\[}}0, 1]]
```
brnorris03: Not insisting, but may be good to capture the values (mostly 0) here instead of hard-coding; wouldn't spend too much time on that if it causes headaches though

brnorris03 (commented):
I think it looks good overall, but I would really like to have an actual device pytest, too (or maybe a python lit test if you can verify this works with lit checks on the generated C++).

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return attr.getInt();
+  }
+  return fallback;
+}
+
+struct TTKernelAnnotateCBUsePass
+    : impl::TTKernelAnnotateCBUseBase<TTKernelAnnotateCBUsePass> {
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    llvm::StringMap<Operation *> symbols;
```
brnorris03: Something to check if you want -- could this use MLIR's `CallGraph` instead of building `symbols` and `callees` manually? It already handles callable lookup and unknown callees. LLVM's SCC traversal can then propagate DFB uses from callees to callers, including recursive calls.

python/ttl/kernel_runner.py
```diff
@@ -533,6 +577,78 @@ def _make_node_core_ranges(nodes: Tuple[Tuple[int, int], ...]) -> Any:
     )
 
 
+def _core_range_coordinates(core_ranges: Any, *, label: str) -> set[Tuple[int, int]]:
```
brnorris03, line 1284: Up to you: could use `ttnn.corerange_to_cores()` here and `core_ranges.num_cores()` above instead of duplicating the `CoreRangeSet` expansion. 

The test fake should match the real method-based API too.

python/ttl/kernel_runner.py
```diff
@@ -647,11 +763,177 @@ def _validate_tensor_backing_aliases(
...
+                            f"DFB[{cb_index}] storage segment claims core "
+                            f"{core} outside the program grid"
+                        )
+                    if core in candidates:
+                        raise ValueError(
+                            f"DFB[{cb_index}] has overlapping storage segments "
+                            f"on core {core}"
+                        )
+                    candidates[core] = source
+        placements.append(
```
brnorris03, line 1518: Should check that every core using this DFB is covered by one of its `storage_segments`. Right now a used core with no matching segment is left out of the descriptor, so this isn't caught when the descriptors are built.

python/ttl/ttl_api.py
```diff
@@ -970,6 +979,20 @@ def _get_optional_kernel_i32_array_attr(module, kernel_name: str, attr_name: str
     return list(attr)
 
 
+def _get_kernel_optional_i32_array_attr(module, kernel_name: str, attr_name: str):
```
brnorris03, line 1009: Could use one optional i32-array helper instead of adding a second copy. Have it return `None`, and use `or []` at the computed-address call site where missing and empty mean the same thing.

python/ttl/kernel_runner.py
```diff
@@ -141,6 +141,26 @@ def get_min_remaining_l1_for_device(device):
     return max(0, budget_bytes - max_core_bytes)
 
 
+def _get_remaining_l1_by_core_for_device(
```
brnorris03, line 167: Could share the buffer-page aggregation with `get_min_remaining_l1_for_device` instead of having two implementations. They both interpret `cb_limit` and sum the L1 pages, so this should live in one helper.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateCBUse.cpp
```diff
@@ -0,0 +1,119 @@
...
+      Operation *key = func.getOperation();
+      if (conservative.contains(key)) {
+        int64_t cbCount = getCBCount(func, moduleCBCount);
+        for (int64_t index = 0; index < cbCount; ++index) {
+          usedCBs[key].insert(static_cast<int32_t>(index));
+        }
+      }
+      SmallVector<int32_t> sorted(usedCBs[key].begin(), usedCBs[key].end());
+      llvm::sort(sorted);
+      func->setAttr(kUsedCBIndicesAttrName,
```
brnorris03: Could only set this on `ttkernel.thread` functions? Python only reads the attribute from kernel entries, so helpers can stay in the analysis and declarations do not need an empty attribute.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateCBUse.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+using CBSet = llvm::SmallDenseSet<int32_t, 8>;
+
+static int64_t getCBCount(func::FuncOp func, int64_t fallback) {
+  if (auto attr =
+          func->getAttrOfType<IntegerAttr>(kBaseCTAIndexAttrName)) {
+    return attr.getInt();
+  }
+  return fallback;
+}
```
brnorris03: Is this not already available as a utility? Seems to me we couldn't have made it this far without but maybe not? And what is the purpose of `fallback`? Definitely could use a function comment.

test/python/test_kernel_runner.py
```diff
@@ -515,6 +522,251 @@ def test_build_cb_descriptors_preserves_subtile_geometry(monkeypatch):
...
+        core_ranges=full_grid,
+        kernel_specs=[_specialized_spec(full_grid, [2])],
+    )
+
+    assert [
+        descriptor.format_descriptors[0].buffer_index for descriptor in descriptors
+    ] == [2]
+
+
+def test_specialized_cb_budget_is_computed_per_core(monkeypatch):
```
brnorris03: Could make this cover the descriptor partitioning more directly: e.g. core 0 uses `[0]`, core 1 uses `[0, 1]`, and core 2 uses `[1]`, then check each descriptor's DFB id and core set together. Right now the assertion only checks that both core sets appear, so it wouldn't catch a DFB being assigned to the wrong core.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
jackzhangTT: good catch. But instead of preserving a typed DFB-use marker wouldn't it be better to remove it entirely? For example:

suppose we start with
```
%dfb = ttl.bind_cb {cb_index = 0} : !ttl.cb<...>
ttl.dprint "input" (%dfb) {mode = "cb"} : (!ttl.cb<...>)
scf.if %is_core_0_0 {
  %tile = ttl.cb_wait %dfb ...
  // compute
}
```

it gets specialized to:

```
func @k_c0_0 {
  emitc.verbatim "ttmlir::dprint(...get_compile_time_arg_val(0)...)"
  %cb = ttkernel.get_compile_time_arg_val(0)
  ttkernel.cb_wait_front(%cb, ...)
}


func @k_c0_1 {
  emitc.verbatim "ttmlir::dprint(...get_compile_time_arg_val(0)...)"
  return
}
```

rather than keeping the `emitc.verbatim` in `k_c0_1` like:

```
func @k_c0_1 {
  %cb = ttkernel.get_compile_time_arg_val(0)   // still here — the print needs it
  emitc.verbatim "..." %cb
  return
}
```

it should be:

```
func @k_c0_1 {
  return
}
```

brnorris03: I am a bit torn between removing it or emitting something like "dfb BLAH not used". Just removing dprints that users actually rely on seeing in all threads seems worse -- what if the lack of the print is some other problem?

jackzhangTT: it wouldn't completely remove the prints the prints would still exist in the kernels where that cb is used. In the above example the print would still exist in c0_1. My thinking is just that after specialization if the kernel isn't used any where then it should be possible to print it either.

brnorris03: I get that, but what if the whole purpose of the dprint is to confirm that the CB is NOT used in a particular kernel? All the user gets is an absence of output, so it's not really possible to know whether it's because of specialization or a different problem with CB allocation. I am not saying we need to support every possible user scenario, but I would make it very clear in the little command line option doc that dprints may be eliminated.

jackzhangTT: I think that's a good suggestion I can add that as part of the command line option and emit a warning whenever I remove a dprint and report the core I'm removing it from

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
jackzhangTT: I didn't know about this this is a much cleaner solution will use LLVM's SCC traversal from now on

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateCBUse.cpp
jackzhangTT: No utility but I think this logic is actually really confusing the fallback was suppose to be the largest DFB count in the module that we previously found in the graph. The whole idea was if the function has no baseCTAIndex attribute then we conservatively give it the largest found DFB count. I removed this helper I think it should help with reading the logic

docs/sphinx/reference/compiler-options.md
```diff
@@ -140,7 +140,7 @@ ttlang-opt input.mlir -p 'ttl-to-ttkernel-pipeline{maximize-dst=true lower-to-em
 | `l1-budget-override` | uint32_t | `0` (target default) | Override the L1 allocation budget used by DFB validation and PipeTransport selection. |
 | `reuse-user-dfbs` | bool | `true` | Reuse physical DFB indices for compatible logical DFBs with proven non-overlapping concurrent lifetimes. |
 | `exact-coloring-search-limit` | uint64 | `1000000` | Maximum states examined during deterministic exact DFB allocation before reporting an inconclusive result. |
-| `specialize-cores` | bool | `false` | Clone TTKernel functions that branch on a core coordinate once per launch coordinate (`ttkernel-specialize-cores`), then run `canonicalize` / `cse`. Maps from `--ttl-specialize-cores`. |
+| `specialize-cores` | bool | `false` | Run the `ttkernel-specialize-and-annotate-dfb-use` sub-pipeline. Maps from `--ttl-specialize-cores`. |
 | `lower-to-emitc` | bool | `false` | Run the TTKernel-to-EmitC backend (produces C++ source). |
 
 The pipeline runs these passes in order:
```
brnorris03, line 147: nit: now enumerates sub-pipelines, too, so perhaps change to "The pipeline runs these passes and sub-pipelines in order"

docs/sphinx/reference/print-debugging.md
```diff
@@ -89,6 +89,16 @@ print(inp_dfb, thread="unpack")
 
 When `thread` is specified, the print is wrapped in the corresponding `DPRINT_MATH(...)`, `DPRINT_PACK(...)`, or `DPRINT_UNPACK(...)` macro. In compute kernels, the thread is automatically selected based on the print mode when no explicit `thread` is given: scalar and DST prints use `math`, DFB and tile prints use `pack`. Tensor page prints (`num_pages=`) are only supported in datamovement kernels. In datamovement kernels, no wrapping is applied when `thread` is omitted.
 
+### Core specialization
+
+`--ttl-specialize-cores` clones kernels per launch coordinate and folds
+branches that depend on the core coordinate. A DFB or tile print that
+survives only on a core with no remaining wait, reserve, or other
+non-print use of that DFB is dropped. The host does not allocate the DFB
+on that core solely to keep the print from faulting. Cores that still
```
brnorris03, line 98: Not asking for changes -- just elaborating a bit, what I meant earlier was not to allocate a DFB just enable its DPRINT user, but to modify the DPRINT itself (since you do know the index at that point, can just create a "DFB <#> not allocated or used on this node" kind of message and append to whatever the original message was. But what you do now is fine.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
```diff
@@ -0,0 +1,257 @@
...
+// True when every use (through unrealized casts) is an emitc.verbatim dprint.
+// An i32 get used as a DFB id for wait/reserve/extern is not print-only.
+static bool isPrintOnlyGet(ttk::GetCompileArgValOp op) {
+  SmallVector<Operation *> users;
+  SmallPtrSet<Operation *, 8> visited;
+  collectPrintUsers(op.getResult(), users, visited);
+  return !users.empty() && llvm::all_of(users, [](Operation *user) {
+    return isa<emitc::VerbatimOp>(user);
+  });
+}
```
brnorris03, line 61: nit: Kind of weird function name, but it's just a helper so don't feel obligated to change.

lib/Dialect/TTKernel/Transforms/TTKernelAnnotateDFBUse.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+static void warnDroppedPrint(func::FuncOp func, int32_t dfbIndex) {
+  InFlightDiagnostic diag = func.emitWarning()
+                            << "eliminating debug print of unused DFB "
+                            << dfbIndex;
+  if (auto coord = func->getAttr("ttl.core_coord")) {
+    diag << " on specialized core " << coord;
+  }
+}
```
brnorris03, line 70: This is great, thanks for adding.

python/ttl/kernel_runner.py
```diff
@@ -28,6 +28,8 @@ def _ensure_ttnn():
     try:
         import ttnn as _ttnn
 
+        if not hasattr(_ttnn, "DataType"):
+            return None
```
brnorris03, line 35: Maybe add a short comment explaining this? Not really clear what this check is for (I am guessing making sure it's a module and not something else, but isn't there a way to check that directly?)

test/python/test_specialize_cores.py
```diff
@@ -264,6 +267,102 @@ def test_specialize_cores_branch_matches_reference(device, monkeypatch, tmp_path
     _assert_reader_cloned(str(final_mlir))
 
 
+def _parse_used_dfb_indices(mlir: str, func_name: str) -> list[int]:
+    match = re.search(
+        rf"func\.func @{re.escape(func_name)}\b[\s\S]*?"
+        r"ttl\.used_dfb_indices = array<i32((?::[^\n>]*)?)>",
+        mlir,
+    )
```
brnorris03, line 275: This can break (worst case -- passes when it shouldn't) if the IR syntax changes slightly, I would much rather have a multiline string MLIR input with some specific indices and just match those directly, not use regex to extract them.

test/ttlang/Dialect/TTL/Transforms/lower_dprint_cb.mlir
```diff
@@ -0,0 +1,15 @@
...
+func.func @cb_print() {
+  %cb = ttl.bind_cb {cb_index = 0, block_count = 2} : !ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>
+  "ttl.dprint"(%cb) {fmt = "input", mode = "cb"}
+      : (!ttl.cb<[1, 1], !ttcore.tile<32x32, bf16>, 2>) -> ()
+  return
+}
+
+// CHECK-LABEL: func.func @cb_print
+// CHECK: %[[CB:.*]] = ttkernel.get_compile_time_arg_val(0) : () -> i32
+// CHECK: emitc.verbatim "ttmlir::dprint(ttmlir::CBPrinter({}));" args %[[CB]] : i32
```
brnorris03, line 15: nit: mostly we put checks above the input IR so I'd follow that convention for consistency.

brnorris03 (approved):
Looks great, minor things only -- up to you whether to address (except maybe the regex in the test one).

PR #938, opened 2026-08-18
[ttl] Accept shape argument for reductions

FConstantinos:
## Summary

- accept an explicit result shape for reduce_sum and reduce_max
- validate the supplied shape against the compiler-derived reduction shape
- preserve shape inference for existing callers that omit the argument
- add regression coverage for invalid rank and multi-tile negative dimensions

## Testing

- pre-commit run --all-files
- invalid_reduce_shape.py and invalid_reduce_shape_rank.py compiler diagnostics
- four focused multi-tile reduce_sum/reduce_max hardware cases (4 passed)


python/ttl/operators.py
```diff
@@ -913,7 +914,25 @@ def _reduce_impl(
             )
     norm_dims = sorted({d % rank for d in dims})
 
-    result_shape = [1 if i in norm_dims else s for i, s in enumerate(input_shape)]
+    expected_shape = [1 if i in norm_dims else s for i, s in enumerate(input_shape)]
+    if shape is None:
+        # Keep accepting the legacy compiler spelling while supporting the
+        # explicit result shape required by the language specification.
+        result_shape = expected_shape
```
zoecarver, line 920: emit deprecation warning?

zoecarver (approved):

python/ttl/operators.py
FConstantinos: done

PR #939, opened 2026-08-18
[ttl] A library for querying the cost of ttkernel ops on execution engines

rguoTT:
# Purpose

Nothing in tt-lang can say what anything costs. This PR
adds the missing input: a library that answers what one TTKernel operation costs
on one execution engine, from measurements taken on silicon.

Intended as a shared bottom layer rather than one feature's helper: compiler
choices made by rule today -- FPU against SFPU for binary ops, subblock shapes,
whether combining pack tiles is worth it -- could compare candidates by cost.

Two properties make it usable for that. The numbers are *measured*, from LLK perf
sweeps narrowed to the configurations tt-lang can actually emit -- not modelled
from instruction counts, and not tt-llk's nightly sweeps, which cover
configurations no tt-lang kernel produces. And the library never invents: a
configuration nothing measured has no answer, so a cost you get has a measurement
behind it, and a miss is something you can report rather than paper over.

The data itself, `lib/Analysis/CostTableBlackhole.inc`, is generated by
`llk-perf/gen_cost_table.py` in tt-lang-ops-and-models, where the measurement
pipeline lives -- it needs a device and a tt-llk harness. **Skip it in review**;
its banner records the revisions of all three repositories it was produced
against. Today: 3127 measured rows over 296 operations, blackhole only.

# The API

The whole surface is `include/ttlang/Analysis/OpCost.h`, which depends on LLVM
Support alone -- no MLIR, no dialect -- so anything can link it and no edge back
from a dialect can ever be created. Recovering the key from IR is the caller's
job.

```cpp
std::optional<Cost> lookup(StringRef op, Engine engine,
                           const OpKey &opKey, const KernelConfig &config,
                           Arch arch = Arch::Blackhole);
```

`op` is the bare operation name (`"copy_tile"`). `Engine` is `Dm`, `Unpack`,
`Math` or `Pack` -- the units that do the work, not the instruction streams:
NCRISC and BRISC both read `Dm`, because an operation does not choose which
data-movement core runs it. `Arch` is per query rather than a build-time constant,
because costs do not transfer between architectures.

```cpp
struct Cost {
  double value;            // cycles, per call or per tile
  double fixed;            // fit intercept against a block dimension; zero today
  Unit unit;               // PerCall (an init zone) or PerTile (a tile loop)
  bool isScalar() const;   // false when `fixed` is non-zero
};
```

The key is two structs, split by where each half comes from -- and it has to be
supplied in full, because a cost is only meaningful in the configuration it was
measured in: `mul_tiles` on math spans 18.95 to 84.16 cycles/tile across the four
math fidelities, and `copy_tile` on unpack reads 41.77 or 121 depending on whether
its buffer is routed straight to DST.

```cpp
struct KernelConfig {        // holds for a whole kernel; recover once, reuse
  StringRef outFormat;       // what the packer wrote
  bool destAcc = false;      // fp32_dest_acc_en
  StringRef dstSync;         // "Half" / "Full"; carried, not consulted
  unsigned faces = 4;        // 4 is a full 32x32 tile
};

struct OpKey {               // what this operation supplies; per query
  StringRef inFormat;        // format of the buffer it reads
  ArrayRef<Knob> knobs;      // {name, value}, e.g. {"math_fidelity", "HiFi4"}
};
```

`outFormat` cannot be per-operation -- `add_tiles` has no output buffer among its
operands, while the measurement was keyed on what the packer wrote. `inFormat`
cannot be per-kernel -- each operation names the buffer it unpacks from. A stale
value on the `OpKey` side is a wrong answer rather than a miss, which is why they
are separate.

`knobs` is untyped on purpose: the library does not know what any of them mean.
Some are kernel-wide (`math_fidelity`), some belong to a buffer
(`unpack_to_dest`), some are attributes of the operation asked about (`mathop`,
`reduce_pool_type`, `approx_mode`, `iterations`, `input_clamping`). Matching walks
the *row's* knobs, so supplying more than a row names costs nothing while missing
one it names makes that row unmatchable -- which is the point. A table that gains
a knob needs no change here.

The rest keeps the flavours of "no answer" apart, which one `std::optional`
cannot:

```cpp
bool isKnownOp(StringRef op, Arch = Arch::Blackhole);            // absent vs known
bool runsOnEngine(StringRef op, Engine, Arch = Arch::Blackhole);
bool runsNowhere(StringRef op, Arch = Arch::Blackhole);          // known and free
unsigned getMeasurementCount(StringRef op, Engine, Arch = ...);  // 0 = untimed slot
ArrayRef<StringRef> getOperations(Arch = Arch::Blackhole);
TableStats getTableStats(Arch = Arch::Blackhole);                // ops, measured rows
StringRef getEngineName(Engine);
void forEachMeasurement(Arch, function_ref<void(const Measurement &)>);
```

An absent operation means the table and the dialect have drifted (generation reads
the operation list out of `TTKernelOps.td`), so a caller should fail rather than
assume zero. `runsOnEngine` with a zero measurement count is missing data; a
non-zero count with no answer from `lookup` is a key this kernel cannot supply.
Consumers need that distinction to say how much of a report is measured.
`forEachMeasurement` hands back each row with the key it was measured at, which is
what makes the table checkable from outside.

# Example

```cpp
using namespace mlir::tt::opcost;

KernelConfig kernel;                                   // from the enclosing function
kernel.outFormat = "Float16_b";
kernel.destAcc   = false;
kernel.faces     = 4;

const Knob knobs[] = {{"unpack_to_dest", "false"}};    // from ttl.unpack_to_dest_fp32
OpKey key{/*inFormat=*/"Float16_b", knobs};            // from the op's own buffer

if (std::optional<Cost> cost = lookup("copy_tile", Engine::Unpack, key, kernel)) {
  // 41.77 cycles per tile on this table
} else {
  // nothing measured this operation in this configuration -- say so
}
```

# Tests

`ttlang-opcost-test` (`test/lib/Analysis/OpCost/OpCostTest.cpp`), run as a lit
test by `test/ttlang/Analysis/opcost_library.mlir`. It links `TTLangOpCost` and
`LLVMSupport` and nothing else, so if the library ever grows an MLIR dependency
the link breaks and says so.

**It pins no cycle counts, deliberately.** The table is regenerated whenever the
measurements are re-taken, and a test whose numbers churn on every re-sweep trains
its reader to update it without looking -- which is how a real loss of data gets
waved through. Every check derives its probes from whatever table is compiled in,
and asserts what the library promises about any of them:

1. **Every measured row is reachable by the key it was measured at.** Rebuilds
   each row into an `OpKey` + `KernelConfig` and requires its own cost back. This
   is also how a conflicting pair of rows is caught -- two rows sharing a key and
   disagreeing answer nothing, so a bad sweep would otherwise make a cost vanish
   from every report rather than announce itself.
2. **A row's knobs are all required and extras ignored.** Drop one, empty the
   list, or set a value nothing measured: all must miss. Add a knob no row names:
   the answer must not change.
3. **Formats and faces are part of the key; `dstSync` is not.** Perturbed to
   values no row can hold, so a miss cannot be another row matching instead.
4. **A slot nothing timed stays silent**, asked nine different ways -- the
   untimed-versus-unmatched distinction consumers depend on.
5. **The operation list, the predicates and the counts agree** with each other.
6. **An unknown operation answers nothing on every engine**, and reads as absent
   rather than known and free.
7. **An architecture with no table borrows nothing**, even for a key Blackhole
   answers.

Failures name the row they broke on:

```
  FAIL answered without a knob the row is keyed on: copy_tile/unpack Bfp8_b->Bfp8_b destAcc=0 faces=4 [unpack_to_dest=false]
FAIL knob matching needs every knob a row names, and no more
```

Verified to actually bite, by breaking the library two ways and watching the right
checks fail: treating a missing knob as a match (check 2) and answering zero for
an untimed slot (check 4). Both exit non-zero. This closes the last limitation the
library shipped with -- until now nothing in this tree called it, so a build
proved it compiled and linked, not that it answered correctly.

## Running it

```bash
cmake --build build --target check-ttlang-mlir      # the CI suite; 317/317 here

cmake --build build --target ttlang-opcost-test     # just this test
llvm-lit -v build/test/ttlang/Analysis/opcost_library.mlir

./build/bin/ttlang-opcost-test                      # or the binary directly
PASS every measured row is reachable by its own key
PASS knob matching needs every knob a row names, and no more
PASS formats and face count are part of the key, dstSync is not
PASS a slot with no measurements answers nothing
PASS the operation list, the predicates and the counts agree
PASS an unknown operation answers nothing on every engine
PASS an architecture with no table answers nothing
opcost: all checks passed
```

`TTLangOpCost` is a separate CMake target in `lib/Analysis/CMakeLists.txt` rather
than part of `TTLangAnalysis`, which links MLIR and three dialects; the split is
what keeps the no-MLIR dependency enforceable, and both targets therefore declare
`PARTIAL_SOURCES_INTENDED`.


include/ttlang/Analysis/OpCost.h
```diff
@@ -0,0 +1,257 @@
...
+/// nothing timed, or in a configuration no sweep presented, answers nothing
+/// rather than an invented number. A caller that needs a figure for every
+/// operation supplies its own fallback and knows that it did.
+///
+/// Rows that disagree by more than a hair are treated as no match: two rows
+/// matching one key means the key is missing a field the measurement depended
+/// on, which is how an unverifiable number would otherwise reach a report.
+std::optional<Cost> lookup(llvm::StringRef op, Engine engine,
+                           const OpKey &opKey, const KernelConfig &config,
+                           Arch arch = Arch::Blackhole);
```
brnorris03: Could we make `arch` required on every query? The API says callers must identify the target because costs could be different, but without the arg you'd always get Blackhole. Side effect of that -- #946 reports Blackhole costs for a Wormhole module instead of rejecting an unavailable table.

lib/Analysis/OpCost.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/// Every field has to agree, `variant` included.
+///
+/// An empty field on the row matches any value: it is how a benchmark that
+/// never varied something stays reachable, rather than demanding a value it
+/// never measured.
+bool keyMatches(const MeasuredCost &row, const OpKey &opKey,
+                const KernelConfig &config) {
+  return row.inFormat == opKey.inFormat && row.outFormat == config.outFormat &&
```
brnorris03, line 155: Mostly a nit: the empty-format contract could be more consistent. The header says an empty `KernelConfig` format is unconstrained, and this function says an empty measured format matches any value, but the implementation requires exact equality. You can encode the intended behavior and test it.

brnorris03 (approved):
Nice work! A couple of small things to fix if you agree.

include/ttlang/Analysis/OpCost.h
rguoTT: Yes, that makes sense.

Issue #863, opened 2026-08-11
reduce_sum/reduce_max: spec and simulator require a shape argument the compiler does not accept


FConstantinos:
## Problem

The spec defines both reduction functions with three parameters:

> `ttl.math.reduce_sum(expr: ttl.BlockExpr, dims: List[int], shape: ttl.Shape) -> ttl.BlockExpr`
> `ttl.math.reduce_max(expr: ttl.BlockExpr, dims: List[int], shape: ttl.Shape) -> ttl.BlockExpr`

(`TTLangSpecification.md` L1486-L1487.) The `shape` parameter is load-bearing in the spec text, which describes reduction as a two-step process driven by which dimensions of `shape` are 1 versus >1, and gives three worked examples all passing it.

The compiler does not accept `shape` at all. It is not optional -- it is absent:

```python
@syntax("reduce_sum")
def reduce_sum(input: TensorBlock, *, dims: List[int]) -> TensorBlock:
```

(`python/ttl/operators.py:838`, and `reduce_max` at L847.)

So spec-conforming code fails to compile, and code that compiles is not spec-conforming. This is the reverse of the usual direction: here the **compiler is more restrictive** than the spec, while the simulator matches it.

## Behaviour

| Surface | Verdict | Reason |
| :--- | :--- | :--- |
| Spec | **Requires `shape`** | `(expr, dims, shape)` at L1486-L1487; the result shape and the two-step tile/scalar reduction semantics are defined in terms of it. |
| Compiler | **Rejects `shape`** | Signature is `(input, *, dims)`; no `shape` parameter exists, so passing it raises `TypeError`. |
| Simulator | **Requires `shape`** | Verified signature: `(block: Block, dims: List[int], shape: Tuple[int, ...]) -> Block`. |

Verified empirically on the simulator:

```
sim reduce_sum sig: (block: sim.dfb.Block, dims: List[int], shape: Tuple[int, ...]) -> sim.dfb.Block
sim reduce_max sig: (block: sim.dfb.Block, dims: List[int], shape: Tuple[int, ...]) -> sim.dfb.Block
```

Compiler rows are source-verified rather than executed: this workspace is a sim-only build (`TTLANG_SIM_ONLY=1`), and forcing it off fails with `ModuleNotFoundError: No module named 'ttl._mlir_libs._ttlang'`.

## Consequence

Kernel 05 (`attention_scaled_dot_product_128x128`) passes two arguments, matching the compiler, and fails on the simulator with:

[...]

PR #940, opened 2026-08-18
[ttl] Support rank-N reductions

FConstantinos:
### Problem description
TT-Lang reduction verification and lowering assumed rank-2 tensors, which prevents Blacksmith rank-4 kernels from compiling. Reductions over leading dimensions also need an explicit DST accumulator because the hardware tile reduction only handles the two innermost dimensions.

### What changed
- Generalize reduction verification, affine maps, iterator types, and compute planning to rank-N tensors.
- Keep the hardware scaler as a rank-2 single tile while supporting rank-N inputs and results.
- Lower leading-dimension sum/max reductions through an initialized DST accumulator, while preserving the existing tile-reduction path for innermost dimensions.
- Teach loop lowering, TTKernel conversion, and DST subblocking about the new reduction initializer and in-place add operations.
- Add rank-4 mixed- and outer-dimension regression coverage, including the production maximize-DST pipeline.

### Ticket
Follow-up to #863 and the Blacksmith rank-4 reduction project item.

### Checklist

### Validation
- Clean out-of-tree build completed.
- Targeted llvm-lit reduction tests: 3/3 passed.
- Full TTKernel/EmitC lowering checked for rank-4 sum and max reductions.
- All pre-commit hooks passed.
- Four iterative review/fix passes completed; final pass found no issues.
- Blacksmith hardware rerun attempted, but the shared host currently has a stale/foreign sysmem mapping that prevents device initialization; results will be updated after hardware becomes available.


zoecarver (approved):
LGTM but please add a python test and @brnorris03 should probably take a look

PR #945, opened 2026-08-19
Preserve DFB lifetime safety for fused DSpark

### LINKED ISSUE #916: ttl.opaque_call declares no memory effects, so effect-only extern calls can be eliminated
## Summary

`ttl.opaque_call` declares no memory effects, so an extern call whose only purpose
is a side effect can be deleted by the compiler. The equivalent op one dialect
down, `ttkernel.opaque_call`, does declare `MemoryEffects<[MemRead, MemWrite]>`.
Both ops were added in the same commit, 5e3e926 (*[ttl] merge
arichins/externFunc: call_extern_func -> ttl.opaque_call*), so this looks like an
oversight rather than a deliberate difference between the two levels.

```tablegen
// include/ttlang/Dialect/TTKernel/IR/TTKernelOps.td
def TTKernel_OpaqueCallOp : TTKernel_Op<"opaque_call",
    [MemoryEffects<[MemRead, MemWrite]>]> { ... }

// include/ttlang/Dialect/TTL/IR/TTLOps.td
def TTL_OpaqueCallOp : TTL_Op<"opaque_call", []> { ... }
```

## Why it matters

`call_extern_func` exists to reach C/C++ that tt-lang has no construct for. A
useful class of those functions returns nothing and exists purely for its effect
on hardware state. From the compiler's position such a call has no result anybody
uses and declares no effects, which makes it indistinguishable from a no-op and
eligible for elimination. The call is opaque by construction, so there is no way
for the compiler to discover the effects by looking inside it — the declaration is
the only signal available.

Externs that return data are unaffected, because their results are consumed and
that keeps them alive. This only bites the effect-only case.

## How we hit it

We are fusing a speculative-decode drafter — a transformer backbone and a sampling
top — into a single `ttl.operation` so a block of drafted tokens costs one dispatch
rather than a pipeline of them. One program means one pool of 64 circular-buffer
indices, and the two halves together want more than 64, so the top's buffers are
coloured onto indices the backbone already used.

A reused slot arrives dirty: pointers part-way round, page counters holding what
the last push/pop left. The top needs those slots rewound to empty. A program
boundary would normally do that, and a single program has none, so we call an
extern (`ttl_reset_cb_range`) that resets the CB state registers directly. It
returns nothing. Without effects declared on `ttl.opaque_call` it is eliminated,
and the top then either waits on pages that were already consumed or blocks
reserving space in a buffer it believes is full.

To be clear about scope: this does not block single dispatch as such. Other
designs that avoid buffer reuse across the seam would not need the reset at all.
It affects this approach to it, and it will affect anyone else calling an
effect-only extern.

## Suggested fix

Declare the same effects on the TTL op as on the TTKernel one, so the two dialects
agree:

```tablegen
def TTL_OpaqueCallOp : TTL_Op<"opaque_call",
    [MemoryEffects<[MemRead, MemWrite]>]> {
```

We are running with exactly this change locally and it resolves the problem. Happy
to raise it as a PR.

A more precise design is possible — effects declared per call site, so genuinely
pure externs keep their optimisation freedom — but that would be a new policy, and
matching TTKernel is the smaller change that makes the current behaviour
consistent.

FConstantinos:
## Summary

- base the DSpark compiler dependency on ops-and-models/sep1
- retain the scope-exit API compatibility fix
- preserve bind-anchored DFB lifetimes so overlapping logical buffers cannot receive the same physical CB
- refresh DFB checks for the metadata and reset attributes emitted by the Sep 1 integration

## Why

The Aug 29 reset-barrier change accidentally removed the bind lifetime anchor from DFB resource planning. That allowed buffers with overlapping effective lifetimes to alias one physical circular buffer, corrupting the fused DSpark output. Restoring the original anchor fixes the regression while keeping reuse for genuinely disjoint lifetimes.

## Validation

- clean TT-Lang build
- ninja -C build check-ttlang: 178 MLIR passed, 1 expected failure; 3 Python binding tests passed; 116 packaging tests passed
- fused DSpark device-top golden tests: 7 passed
- exact Galaxy branch DSpark suite: 533 passed, 3 skipped
- downstream K3-spatial superset: 602 passed, 4 skipped
- git diff --check

## Target-branch CI note

The current checks also expose pre-existing drift in ops-and-models/sep1: its tt-metal gitlink is a749fcc6e69c while the version manifest expects e908c31332b6 and the former is not fetchable from the public tt-metal remote; simulator tests still assert a 32-DFB default while the branch config uses 64; and the repository-wide pre-commit hook reports formatting changes across unrelated base files. Those failures are outside this five-file compiler fix and are documented here so they are not mistaken for DSpark regressions.

### DIFF
--- a/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
@@ -594,6 +594,10 @@ struct TTLFinalizeDFBIndicesPass
         }
       };
 
+      // The bind anchors the interval; binds are hoisted to the function
+      // entry, so the start is refined to the first acquire below.
+      extendInterval(dfb, bindOp, func, *order);
+
       // Pipe destinations are slot-addressed (block_count == sender count)
       // and pipe sends read asynchronously, so pipe-attached DFBs keep
       // dedicated indices.
--- a/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
+++ b/lib/Dialect/TTL/Transforms/TTLFinalizeDFBIndices.cpp
@@ -130,8 +130,8 @@ struct LogicalDFB {
 
 enum class ResetControlFlow { Linear, Cyclic };
 
-constexpr llvm::StringLiteral kDFBResetPreservedIndicesAttrName(
-    "ttl.dfb_reset_preserved_indices");
+constexpr llvm::StringLiteral
+    kDFBResetPreservedIndicesAttrName("ttl.dfb_reset_preserved_indices");
 
 enum class HardwareDataFormat : int64_t {
   Float32 = 0,
@@ -450,8 +450,7 @@ struct TTLFinalizeDFBIndicesPass
 
     SmallVector<SmallVector<int64_t>> preservedByResetOrdinal(
         resetCount.value_or(0));
-    SmallVector<bool> initializedPreserveOrdinal(resetCount.value_or(0),
-                                                 false);
+    SmallVector<bool> initializedPreserveOrdinal(resetCount.value_or(0), false);
     llvm::DenseSet<int64_t> preservedLogicalIndices;
     for (func::FuncOp func : moduleOp.getOps<func::FuncOp>()) {
       auto callsIt = resetCallsByFunction.find(func.getOperation());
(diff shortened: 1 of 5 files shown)


brnorris03 (approved):
Thank you for catching this and the fix! You may want to try `main` based tt-lang when convenient -- it should not have this or other related issues.

Issue #800, opened 2026-08-04
[Compiler] Static cost estimator for TTKernel modules


rguoTT:
# Static cost estimator for TTKernel modules

## Purpose

Estimate the cycle cost of a device launch by analyzing TTKernel IR, so **optimization decisions can be evaluated without a device, a profiler run, or a build.** The target is ranking candidates against each other, not predicting absolute runtime.

The estimator is **read-only**: it never mutates the IR.

```
--ttl-cost-estimate                      # summary
--ttl-cost-estimate-detail               # per-lane op tables + event timeline
--ttl-cost-estimate-timeline-step=N      # fixed-step timeline
```

## Approach

**Five lanes.** `NCRISC`, `TRISC0 unpack`, `TRISC1 math`, `TRISC2 pack`, `BRISC`, each an in-order timeline. Ordering is free within one instruction stream, cross-lane handoffs are modeled in addition.

**Per-TRISC split.** The three TRISCs overlap, and how much overlap a kernel achieves depends on the CB depth, DST-half and Src-bank credits — so a single compute timeline can only sum the work, and cannot attribute cost to a lane.

**Thread affinity table.** `ThreadType` is only `{Noc, Compute, Ethernet}`, so a compute function is split three ways: for each TTKernel op, which of the three TRISCs it runs on. Affinity is read off the compute API headers and their `UNPACK()`/`MATH()`/`PACK()` wrappers (`common_globals.h:15-36`) — each expands to its argument only under the matching `TRISC_*` define, so the wrapper around an `llk_*` call is what decides which TRISC runs it. Those calls are the evidence for a lane, cited per entry in `scripts/lane_work.py`, but they are not what the table records: cost is measured per operation per lane, and a resource effect is a property of the operation rather than of any one call it makes.

`dm` is one slot rather than one per data-movement RISC, because an operation does not choose which core it runs on — `ttl.noc_index` on the enclosing function does.

| TTKernel op | dm | unpack | math | pack | evidence |
|---|---|---|---|---|---|
| `add_tiles` | — | ✓ | ✓ | — | `llk_unpack_AB`, `llk_math_eltwise_binary` |
| `binary_op_init_common` | — | ✓ | ✓ | ✓ | `llk_unpack_hw_configure` + `llk_unpack_AB_init`; `llk_math_pack_sync_init` + `llk_math_hw_configure`; `llk_pack_hw_configure` + `llk_pack_init` + `llk_pack_dest_init` |
[...]

PR #946, opened 2026-08-19
[ttl] Cost Estimator based on tracing execution on RISC cores with synchronization awareness

rguoTT:
# Purpose

Estimate the cycle cost of a device launch by reading TTKernel IR, so that
**optimization candidates can be compared without a device, a profiler run, or a
build**. The target is ranking lowerings against each other -- does subblocking
pay for the extra init zones, does an FPU path beat an SFPU one, which lane is
this kernel actually bound on -- not predicting absolute runtime. Read-only: it
never mutates the IR, and it never fails a compile.

Design and open gaps: https://github.com/tenstorrent/tt-lang/issues/800

TTKernel just before EmitC is the stage that makes this possible. Operations map
one-to-one onto the emitted compute-API calls, the constants carrying tile counts
are folded, and the circular-buffer calls are still operations rather than the
opaque `emitc.verbatim` strings they become afterwards.

```
$ ttlang-opt --ttkernel-cost-estimate test/ttlang/Analysis/cost_estimator.mlir -o /dev/null
cost estimate: 17 of 35 placements measured (49%)
  0 unmatched (measured in no configuration this kernel can supply), 18 untimed (nothing measured them); both charged 0
  measured on blackhole from 3127 rows over 296 operations; per-operation provenance in the detail view
  kernels: read compute write
  NCRISC reader: 5 ops, 5 busy, 0 stalled, 532 drained, 1% utilized
  TRISC0 unpack: 10 ops, 389 busy, 4 stalled, 144 drained, 72% utilized
  TRISC1 math: 8 ops, 257 busy, 154 stalled, 126 drained, 48% utilized
  TRISC2 pack: 9 ops, 329 busy, 205 stalled, 3 drained, 61% utilized
  BRISC writer: 3 ops, 3 busy, 534 stalled, 0 drained, 1% utilized
  latency: 537
  busiest lane: TRISC0 unpack (389 busy)
```

`--ttl-cost-estimate-detail` adds per-lane operation tables with start, end, cost
and stall per placement, and a timeline with one row per event boundary -- every
start, every finish, and the instant a lane began waiting, so nothing is aliased
away and no interval is invented. The `gap` column is the distance to the next
event on *any* lane, which is what row height does and does not mean here:

```
timeline: one row per event, '|' running, 'w' waiting, '^' ended, '.' idle

    cost    gap  NCRISC  |TRISC0  |TRISC1  |TRISC2  |BRISC   |
                 --------|--------|--------|--------|--------|
       0      1  cbresv  |w       |binit   |cbresv  |w       |
       4      1  cbpush  |cbwait  ||       ||       |w       |
       6     86  .       |binit   ||       ||       |w       |
      92      1  .       ||       |dstacq  ||       |w       |
      93     84  .       ||       |addinit ||       |w       |
     177      3  .       |addinit ||       ||       |w       |
     243     37  .       |add     |w       |w       |w       |
     280     19  .       |add     |add     |w       |w       |
     299     18  .       ||       |w       |w       |w       |
     391      1  .       |cbpop   |add     |w       |w       |
     410      1  .       |.       |dstcmt  |w       |w       |
     411      1  .       |.       |^       |dstwait |w       |
     412     30  .       |.       |.       |pack1   |w       |
     532      1  .       |.       |.       |dstrel  |w       |
     534      1  .       |.       |.       |^       |cbwait  |
     537    end  .       |.       |.       |.       |^       |
```

That is the eltwise-add module above, abridged. It is also the argument for the
credit model over a `max(lane busy)` heuristic: MATH trails UNPACK by one tile
throughout, because it waits on a Src bank the unpacker has yet to fill, and PACK
cannot start at all until `tile_regs_commit` hands over the whole four-tile block
at 410 -- so 537 of latency comes out of 389 of busiest-lane work, and a heuristic
would rank a kernel that pipelines properly identically to one that does not.

# Approach

**Five lanes.** `NCRISC`, `TRISC0 unpack`, `TRISC1 math`, `TRISC2 pack`, `BRISC`,
each an in-order blocking timeline. A compute function is split three ways
because the TRISCs overlap, and how much overlap a kernel achieves is exactly
what a single summed compute timeline cannot say.

**Placement.** Every operation lands on the lanes the cost table says it occupies
-- one source of truth for both where an operation runs and what it costs.
Traversal is structural rather than a walk, so a loop body repeats in place and a
lane holds the program in execution order. Loops are unrolled within a budget.

**Resources as credit counters.** Circular-buffer occupancy, the MATH/PACK
semaphore behind the DST halves, and the SrcA/SrcB dvalid banks. Nothing models
double buffering directly: whether an acquire blocks is an outcome of a counter,
so raising a block count changes no code. Acquires take effect when an operation
starts and releases when it retires, so a push becomes visible to its consumer at
the end of the push. A program that cannot progress is reported as a deadlock
with each lane's blocking reason, rather than as a plausible number.

**Costs are measured or absent.** From `TTLangOpCost`, which holds measurements
alone, so a placement is either measured or charged nothing -- and the report
keeps the two reasons for nothing apart: `nokey`, a measurement exists in no
configuration this kernel can supply, and `untimed`, nothing measured that engine
at all. Recovering the lookup key is this pass's job: the packed format and the
dest-accumulation and sync flags from the enclosing function, `unpack_to_dest`
from `ttl.unpack_to_dest_fp32` against the buffer an operation reads, and
`mathop`, `reduce_pool_type`, `approx_mode`, `iterations` and `input_clamping`
from the operation's own attributes. Math fidelity is the one key field TTKernel
IR does not carry -- it is a runtime compute-config value -- so it arrives as a
pass option, forwarded from the operation's `math_fidelity` on the Python path.

**Where it runs is not settled.** The pass sits last in the TTKernel stage today
because that is the earliest point where every operation exists and none has
become an `emitc.verbatim` string -- which makes it a convenient place to test,
not the right place to consume. Ranking candidates means asking about several
lowerings, which is a query from whatever code is choosing between them, not a
report printed once at the end of a compile. The pass wrapper is deliberately
thin for that reason: it owns only where the estimate happens and where the
report goes, while `mlir::tt::CostEstimator` is reusable by any caller. Expect
the placement to move once the first consumer decides what it needs.

**All or nothing.** An estimate covers every operation in a module or is not
produced, so a kernel the estimator cannot model is never mistaken for a cheap
one. Each gap is warned about at the operation responsible -- an operation the
table does not know, control flow whose outcome is unresolved, a loop whose trip
count is not static -- and the report then says only that no estimate is
available.

Against the plan in #800, this closes: pointing the estimator at the generated
table, reporting a miss instead of falling back to an invented number, per-lane
provenance in the report, computing `unpack_to_dest` per operation, answering the
SFPU knobs, and per-lane rather than per-benchmark keying. Still open there:
steady state for loops instead of unrolling, folding conditionals (`scf.if` fails
the estimate today), recovering the face count from IR, splitting the MATH lane
into FPU and SFPU, data-movement measurements, and hardware validation of the
ranking.

# Tests

Four lit tests, all driving the pass over hand-written TTKernel IR. Unlike the
library's own test, these **do** pin exact costs: the point of each is which row
the lookup matched, and a cost that moves without the table moving is the
regression they exist to catch. Regenerating the table can move them -- read the
new values out of `lib/Analysis/CostTableBlackhole.inc`, not out of the failure.

**`cost_estimator.mlir`** -- an eltwise add over two input buffers and one output,
four tiles per block, across all three kernel threads. Checks that the threads
are recognized and none deadlocks, that nothing is unmatched, that both halves of
the add are measured on both engines they occupy, and that the credit operations
report as untimed. This is the regression test for the table wiring: a lookup that
stops matching shows up as a column flipping away from `meas`.

**`cost_estimator_copy_tile.mlir`** -- the unary datacopy path, whose every row is
keyed on `unpack_to_dest`: the flag changes the unpacker's route rather than its
speed, and the rows differ by roughly 3x (42 cycles/tile unlisted against 120 for
a listed f32 buffer). This kernel lists nothing in `ttl.unpack_to_dest_fp32`, so
the knob answers false for each buffer and the test pins the unlisted figure. It
is the only test that reaches the per-buffer recovery.

**`cost_estimator_knobs.mlir`** -- one kernel run three times, without fidelity
and at HiFi4 and LoFi, exercising all three origins of a knob in one function:
the reduce attributes (`mathop`, `reduce_pool_type`), the exponential's own
attributes across three shapes (default clamping, clamp skipped, and no
attributes at all -- the last being what a plain `ttl.exp` emits), and fidelity
from the pass option. Without fidelity the four math rows keyed on it report
`nokey` rather than borrowing another fidelity's number; supplying it turns them
measured, and LoFi's multiply costs a quarter of HiFi4's. The clamp is worth a
factor of four on the approximate exponential, which is why it has to be
answered.

**`cost_estimator_sfpu_format.mlir`** -- an SFPU operation names no buffer at all
(`exp_tile` takes a DST index), so its format comes from the kernel's buffers,
split into an input and an output side by which ones the packer writes. First
chunk: a kernel reading bf16 and packing f32 resolves both exponential
placements, a pair the table measures. Second chunk: a kernel reading two formats
leaves the exponential `nokey`, because which format reached DST is no longer
decidable -- while `add_tiles`, which names its own buffer, still resolves. The
honest limit is pinned alongside the capability.

## Running them

```bash
cmake --build build --target check-ttlang-mlir      # the CI suite; 321/321 here

llvm-lit -v build/test/ttlang/Analysis              # just the analysis tests
ttlang-opt --ttkernel-cost-estimate=detail=1 test/ttlang/Analysis/cost_estimator.mlir -o /dev/null
```

End to end from Python, on any kernel:

```bash
python examples/eltwise_add.py --ttl-cost-estimate --ttl-cost-estimate-detail
```

Note what that particular example reports: the estimate is unavailable, because
the kernel contains an `scf.if` the estimator declines to model rather than
counting a guarded body as taken. That is the intended behaviour for an
unmodelled construct, and the loop-free kernels the tests use are what exercise a
complete estimate today.

### DIFF
--- a/lib/Analysis/CostEstimator.cpp
+++ b/lib/Analysis/CostEstimator.cpp
@@ -1,0 +1,1552 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttlang/Analysis/CostEstimator.h"
+#include "ttlang/Analysis/OpCost.h"
+
+#include "ttlang/Analysis/LoopIterationUtils.h"
+
+#include "ttlang/Dialect/TTKernel/IR/TTKernel.h"
+#include "ttlang/Dialect/TTKernel/IR/TTKernelOps.h"
+#include "ttlang/Dialect/TTKernel/IR/TTKernelOpsTypes.h"
+#include "ttlang/Dialect/TTKernel/IR/TTKernelTraits.h"
+#include "ttlang/Dialect/TTL/IR/TTL.h"
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/LoopLikeInterface.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringSet.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <algorithm>
+#include <cassert>
+#include <cinttypes>
+#include <cmath>
+#include <limits>
+#include <optional>
+
+namespace mlir::tt {
+
+namespace {
+
+constexpr llvm::StringLiteral kThreadAttrName("ttkernel.thread");
+constexpr llvm::StringLiteral kTTKernelDialect("ttkernel");
+
+/// Loop iterations a single kernel may unroll, shared across the whole nest.
+/// Exceeding it fails the estimate rather than truncating the program; see
+/// placeLoop.
+///
+/// Counted in iterations, which is not the quantity that matters: a body of N
+/// operations costs trip * N placements, and placements are what the report
+/// holds and what the scheduler walks, so this bounds time and memory only
+/// loosely. A 16x16 block nest over 4x4-tile blocks needs a few tens of
+/// thousands of iterations and places 59k operations, which the estimator
+/// handles in a few hundred milliseconds and no measurable memory over the rest
+/// of a compile. This value is set well above such a kernel while still
+/// bounding a pathological one, but a loop whose body is far larger can still
+/// cost more than the count suggests.
+constexpr std::uint64_t kUnrollBudget = 1ull << 20;
+
+/// Output caps. A kernel whose loops unroll to tens of thousands of operations
+/// would otherwise emit a report longer than anyone will read. Truncation is
+/// always announced, never silent.
+constexpr size_t kMaxTimelineRows = 200;
+constexpr size_t kMaxListedOpsPerLane = 30;
+
+using Lane = CostEstimator::Lane;
+using opcost::Engine;
+using opcost::KernelConfig;
+
+/// The engine a lane's work is charged to.
+///
+/// Five in-order instruction streams over four engines: NCRISC and BRISC both
+/// read `Dm`, because an operation does not choose which data-movement core
+/// runs it -- `ttl.noc_index` on the enclosing function does.
+Engine engineFor(Lane lane) {
+  switch (lane) {
+  case Lane::Ncrisc:
+  case Lane::Brisc:
+    return Engine::Dm;
+  case Lane::Trisc0Unpack:
+    return Engine::Unpack;
+  case Lane::Trisc1Math:
+    return Engine::Math;
+  case Lane::Trisc2Pack:
+    return Engine::Pack;
+  }
+  llvm_unreachable("unhandled lane");
+}
+
+using ResourceEffect = CostEstimator::ResourceEffect;
+
+/// Resource effect of one TTKernel operation, read from its operands.
+///
+/// Nothing here is hand-maintained data: the buffer identity, the tile count
+/// and the capacity all come from the module. Only the op-name-to-kind mapping
+/// is fixed, and that follows from the operation's own semantics. Operations
+/// that unpack into SrcA/SrcB without carrying `TTKernelFPUOpTrait`. The trait
+/// covers the six FPU ops; these are the datacopy and (un)tilize paths, which
+/// feed Src exactly the same way but are not FPU operations. Missing one means
+/// MATH is free to run before the data exists, so the assert below cross-checks
+/// this list against the lanes each operation runs on.
+const llvm::StringSet<> &getNonFpuSrcCouplingOps() {
+  static const llvm::StringSet<> ops = {"copy_tile",
+                                        "copy_block_matmul_partials",
+                                        "tilize_block", "untilize_block"};
+  return ops;
+}
+
+/// Whether an operation's lanes are consistent with a Src handshake: it has to
+/// fill a bank on UNPACK and drain one on MATH. Used only to validate the set
+/// above.
+///
+/// This is the half of the old per-call cross-check that survives at lane
+/// granularity, and it is the half that catches the mistake that matters: an
+/// operation named as Src coupled but not running on both lanes would leave
+/// MATH waiting on a bank nothing fills. The converse -- an operation on both
+/// lanes that only configures the unpacker -- can no longer be checked, because
+/// `add_tiles` and `add_tiles_init` are indistinguishable once the call names
+/// are gone.
+bool lanesAllowSrcCoupling(llvm::StringRef name) {
+  return opcost::runsOnEngine(name, Engine::Unpack) &&
+         opcost::runsOnEngine(name, Engine::Math);
+}
+
+/// Operations whose MATH half re-initializes the DST pipeline.
+///
+/// Their MATH half calls `llk_math_pack_sync_init`, which spins until the
+/// MATH/PACK semaphore reads zero -- every committed half drained -- before
+/// re-seeding that semaphore and resetting the DST section base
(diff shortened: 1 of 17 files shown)


lib/Analysis/CostEstimator.cpp
```diff
@@ -0,0 +1,1552 @@
...
+               << inFormat;
+        for (const opcost::Knob &knob : knobs.knobs) {
+          keyOut << '\0' << knob.name << '=' << knob.value;
+        }
+      }
+
+      auto [it, inserted] = resolved.try_emplace(key);
+      if (inserted) {
+        opcost::OpKey opKey{inFormat, knobs.knobs};
+        it->second = opcost::lookup(name, engineFor(lane), opKey, config);
```
brnorris03: Could we read `ttl.target_arch` once and pass it through every OpCost query? A Wormhole module currently receives the Blackhole 555-cycle estimate because all calls use the default architecture.

lib/Analysis/CostEstimator.cpp
```diff
@@ -0,0 +1,1552 @@
...
+
+class CostEstimator::Impl {
+public:
+  Impl(ModuleOp module, Options options) : module(module), options(options) {}
+
+  FailureOr<Report> estimate() {
+    Report report;
+    uint64_t ttkernelOps = 0;
+
+    for (auto funcOp : module.getOps<func::FuncOp>()) {
```
brnorris03, line 1032: Should this partition functions by `ttl.core_coord` and schedule each core independently? Specialized clones execute concurrently on distinct cores, but this appends every clone to one five-lane program with shared DFB state. Two identical clones report 320 cycles; either clone alone reports 186.

python/ttl/ttl_api.py
```diff
@@ -2436,6 +2436,19 @@ def _lower_program_to_kernel(
                 "canonicalize",
                 "cse",
             ]
+        # Runs last in the TTKernel stage: after the passes that create the ops
+        # it reads, and before EmitC conversion turns circular-buffer calls into
+        # opaque verbatim strings. The measurements the estimate draws on are
+        # keyed on math fidelity, which the IR does not carry, so it is passed
+        # through here rather than recovered there.
+        if compiler_options.cost_estimate:
```
brnorris03: Let's reject `--ttl-cost-estimate-detail` unless `--ttl-cost-estimate` is enabled -- I think that simplifies things.

python/ttl/ttl_api.py
```diff
@@ -2436,6 +2436,19 @@ def _lower_program_to_kernel(
...
+            detail = int(compiler_options.cost_estimate_detail)
+            estimate_options = [f"detail={detail}"]
+            if math_fidelity:
+                estimate_options.append(f"math-fidelity={math_fidelity}")
+            pipeline_passes.append(
+                "ttkernel-cost-estimate{" + " ".join(estimate_options) + "}"
```
brnorris03: Could this also be wired through `ttl-to-ttkernel-pipeline`? That pipeline is documented as equivalent to the Python pipeline, but only `_lower_program_to_kernel` can enable this pass. Ideally would also add enabled/disabled pipeline tests and update the compiler-options docs accordingly.

test/ttlang/Analysis/cost_estimator.mlir
```diff
@@ -0,0 +1,97 @@
+// RUN: ttlang-opt --ttkernel-cost-estimate=detail=1 %s -o /dev/null 2>&1 | FileCheck %s
```
brnorris03: Could use one device validation that compares two real lowering candidates and checks each RISC against profiler data. The current lit tests validate table lookup and internal scheduling only, so they do not establish the ranking produced through the estimator. I am fine with just deferring that with a TODO.

brnorris03 (approved):
LGTM, a few minor optional things.

python/ttl/ttl_api.py
rguoTT: `--ttl-cost-estimate-detail` alone now fails with a message.

python/ttl/ttl_api.py
rguoTT: Wired. Doc not updated yet

PR #947, opened 2026-08-20
[ttl] Fuse binary + broadcast into a single FPU op

kwerblinskiTT:
### Problem description

A row/col/scalar broadcast feeding `add`/`sub`/`mul` materialized the broadcast into DST via `unary_bcast` and then ran a separate SFPU binary op. The FPU can apply the broadcast to its SRCB unpack source instead, so the pair can collapse into one instruction with no DST round-trip.

### What's changed

Adds `ttl.tile_binary_bcast` and `ttkernel.binary_bcast` (+ init), lowering to the metal `any_tiles_bcast`/`init_bcast` APIs, which already exist on the current tt-metal pin. The fold is decided in the compute-op planner alongside the matmul accumulator fold rather than as a peephole, so DST assignment and scheduling see the fused op from the start.

Falls back to the existing `tile_bcast` + binary path when the FPU cannot express the expression: `broadcast(B) - A` (sub is not commutative and only SRCB is broadcast), inter-tile-only broadcasts, multi-use broadcast results, and operands already living in DST rather than a CB.

Five existing tests encoded the pre-fusion lowering. Three plus the softmax DSL test now expect the fused op. The DST-capacity and subblocking tests depended on scratch slots the fold removes, so they switch to the broadcast-on-the-left subtraction the FPU cannot fold, preserving their original assertions.

Measured on `flash_shard`, ~1.5x fewer device cycles with PCC unchanged, across three interleaved runs at under 0.15% variance:

| tokens | main | branch | speedup |
| --- | --- | --- | --- |
| 1k | 141,537 | 96,436 | 1.47x |
| 32k | 4,333,421 | 2,876,957 | 1.51x |
| 64k | 8,659,610 | 5,747,039 | 1.51x |

### Checklist


zoecarver (approved):
Awesome! LGTM!

lib/Dialect/TTL/Transforms/ComputeOpCreationPlanning.cpp
```diff
@@ -498,6 +498,47 @@ static LogicalResult buildFusedOperationPlans(ComputeOpCreationPlan &plan,
...
+
+    Value other = user->getOperand(0) == broadcast.getResult()
+                      ? user->getOperand(1)
+                      : user->getOperand(0);
+    if (!plan.trace.rootInputs.contains(broadcast.getInput()) ||
+        !plan.trace.rootInputs.contains(other)) {
+      continue;
+    }
+
+    foldedBroadcastByBinary.try_emplace(user, broadcast);
```
brnorris03, line 538: Should we keep the broadcast separate when a signpost region or tile `dprint` occurs between it and the binary op? The matmul-accumulator fold handles this with `hasInstrumentationBetween`, could do something similar if you agree.

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -3212,6 +3225,13 @@ class ConvertTTKernelToEmitCPass
         TTKernelToEmitCOpaqueRewriter<ttkernel::TensorAccessorOp>>(
         typeConverter, context);
 
+    // The fused binary broadcast ops keep dialect-level names while metal
+    // exposes them as init_bcast/any_tiles_bcast, so the callee is overridden.
+    patterns.add<TTKernelToEmitCOpaqueRewriter<ttkernel::BinaryBcastInitOp>>(
```
brnorris03, line 3230: Could we lower this to `bcast_init` and remove `out_cb` from the per-op init? `ttkernel-insert-inits` already emits `binary_op_init_common`, so `init_bcast` repeats common setup and is also deprecated at the pinned tt-metal revision.

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1422,6 +1422,30 @@ def TTL_TileBcastOp : TTL_TileOp<"tile_bcast", "Broadcast",
...
+    otherwise cost a broadcast into DST followed by an SFPU binary op.
+
+    Because the hardware only broadcasts the second unpack source, `rhs` is
+    always the broadcast operand. Commutative operations (add, mul) may swap
+    their operands to satisfy this; sub only supports `lhs - broadcast(rhs)`.
+
+    The output operand carries the output CB for init-time PACK configuration
+    only. Maps to ttkernel.binary_bcast_init + ttkernel.binary_bcast.
+  }];
+  let arguments = (ins AnyType:$lhs, AnyType:$rhs, AnyType:$output,
```
brnorris03, line 1442: Suggest using `TTL_TileType` and `AllTypesMatch` here. This would catch wrong IR such as `(i32, f32, index) -> i64` (now would make it to TTL-to-TTKernel lowering).

include/ttlang/Dialect/TTL/TTLElementwiseOps.def
```diff
@@ -115,6 +121,9 @@ TTL_UNARY_TILE_OP(Trunc, TruncTileOp, RoundingTileInitOp, TruncTileOp)
 // CB -> DST ops with attribute
 TTL_CB_UNARY_ATTR_OP(Bcast, TileBcastOp, UnaryBcastInitOp, UnaryBcastTileOp)
 
+// CB, CB -> DST ops with attributes
+TTL_CB_BINARY_ATTR_OP(BinaryBcast, TileBinaryBcastOp, BinaryBcastInitOp, BinaryBcastTileOp)
```
brnorris03, line 125: I think this entry and `TTL_CB_BINARY_ATTR_OP` can be removed now? Also `TTL_CB_UNARY_ATTR_OP`?

test/ttlang/Dialect/TTL/Transforms/subblock_matmul.mlir
```diff
@@ -107,7 +108,7 @@ func.func @scaled_acc_matmul_one_output_tile_subblock(
   %old_attached = ttl.attach_cb %o_old, %cb3 : (tensor<1x2x!ttcore.tile<32x32, f32>>, !ttl.cb<[1, 2], !ttcore.tile<32x32, f32>, 2>) -> tensor<1x2x!ttcore.tile<32x32, f32>>
   %reserve = ttl.cb_reserve %cb4 : <[1, 2], !ttcore.tile<32x32, f32>, 2> -> tensor<1x2x!ttcore.tile<32x32, f32>>
   %alpha_bcast = ttl.block.broadcast %alpha_attached dims = [-1], shape = [1, 2] : tensor<1x1x!ttcore.tile<32x32, f32>> -> tensor<1x2x!ttcore.tile<32x32, f32>>
-  %scaled = ttl.mul %alpha_bcast, %old_attached : tensor<1x2x!ttcore.tile<32x32, f32>>, tensor<1x2x!ttcore.tile<32x32, f32>> -> tensor<1x2x!ttcore.tile<32x32, f32>>
+  %scaled = ttl.sub %alpha_bcast, %old_attached : tensor<1x2x!ttcore.tile<32x32, f32>>, tensor<1x2x!ttcore.tile<32x32, f32>> -> tensor<1x2x!ttcore.tile<32x32, f32>>
```
brnorris03, line 111: Changing the existing `mul` cases to non-foldable `sub` means neither DST test exercises the new fold, so its reduced scratch use and resulting subblock selection are untested. Could we keep these `sub` cases and restore the `mul` cases with updated expectations?

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1422,6 +1422,30 @@ def TTL_TileBcastOp : TTL_TileOp<"tile_bcast", "Broadcast",
...
 }
 
+def TTL_TileBinaryBcastOp : TTL_TileOp<"tile_binary_bcast", "Broadcast",
+    [TTL_CBInputTileOpTrait, TTL_DstResultOpTrait]> {
+  let summary = "Fused tile-level binary operation with a broadcast operand";
+  let description = [{
+    Computes `lhs <eltwise_binary_type> broadcast(rhs)` in a single FPU pass.
+    Both operands are read from circular buffers and the broadcast is applied
+    to `rhs` as it is unpacked, so it is never materialized into DST. This
+    replaces the ttl.tile_bcast + ttl.tile_{add,sub,mul} pair, which would
```
brnorris03, line 1432: I think the part starting with "This replaces" is not something that should be in the op description since it refers to a pass (the planning pass can include that perhaps)? I'd just include semantics and a usage example in op descriptions.

brnorris03 (commented):
Thank you for adding this!

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -3214,6 +3227,13 @@ class ConvertTTKernelToEmitCPass
     patterns.add<TTKernelToEmitCOpaqueRewriter<ttkernel::PackWaitedTileOp>>(
         typeConverter, context, "pack_tile");
 
+    // The fused binary broadcast ops keep dialect-level names while metal
+    // exposes them as init_bcast/any_tiles_bcast, so the callee is overridden.
+    patterns.add<TTKernelToEmitCOpaqueRewriter<ttkernel::BinaryBcastInitOp>>(
+        typeConverter, context, "init_bcast");
```
brnorris03: `init_bcast` is deprecated for removal after September 15. You could initialize hardware once at kernel entry with `compute_kernel_hw_startup`, then emit the required data-format reconfiguration and `bcast_init` when the input DFB pair changes. Also please add a test with two different DFB pairs in one region to verify that each receives the correct configuration.

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
```diff
@@ -542,6 +543,21 @@ static LogicalResult buildFusedCompute(Operation *sinkOp,
       tileResult = matmul;
       break;
     }
+    case FusedOperationRecipe::BinaryBroadcast: {
+      std::optional<EltwiseBinaryType> eltwiseBinaryType =
+          getFusedEltwiseBinaryType(operationPlan.source);
```
brnorris03: You can record the binary kind in `FusedOperationPlan` with the broadcast kind, so this doesn't have to be rediscovered here.

brnorris03 (changes requested):

lib/Dialect/TTL/Transforms/ConvertTTLToCompute.cpp
kwerblinskiTT: Done

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
kwerblinskiTT: Done

docs/development/broadcast-initialization.md
```diff
@@ -0,0 +1,65 @@
+# Broadcast initialization
+
+The init insertion pass emits one `compute_kernel_hw_startup` at function entry,
```
brnorris03, line 3: Can you also update `DST_Allocation.md`, `DST_Utilization.md`, and `ComputeOpCreation.md` and link this page from `implementation.md`? Those docs still describe the old full-init behavior.

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1654,6 +1654,41 @@ def TTL_TileBcastOp : TTL_TileOp<"tile_bcast", "Broadcast",
   let assemblyFormat = "$input `,` $output $bcast_type `into` `dst` `[` $dst_index `]` attr-dict `:` `(` type($input) `,` type($output) `)` `->` type($result)";
 }
 
+def TTL_TileBinaryBcastOp : TTL_TileOp<"tile_binary_bcast", "Broadcast",
+    [TTL_CBInputTileOpTrait, TTL_DstResultOpTrait,
+     AllTypesMatch<["lhs", "result"]>]> {
+  let summary = "Fused tile-level binary operation with a broadcast operand";
+  let description = [{
+    Computes `lhs <eltwise_binary_type> broadcast(rhs)` in a single FPU pass.
```
brnorris03, line 1662: Minor: Up to you, but since `mul` moves from SFPU to the fidelity-controlled FPU here, its numerics now depend on the math fidelity option -- perhaps mention that in the PR overview which becomes the commit comment.

include/ttlang/Dialect/TTL/IR/TTLOps.td
```diff
@@ -1654,6 +1654,41 @@ def TTL_TileBcastOp : TTL_TileOp<"tile_bcast", "Broadcast",
...
+    Example:
+    ```mlir
+    %r = ttl.tile_binary_bcast %lhs, %rhs, %out 0 : i32 2 : i32 into dst[%c0]
+           : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>,
+              !ttcore.tile<32x32, f32>)
+             -> !ttcore.tile<32x32, f32>
+    ```
+  }];
+  let arguments = (ins TTL_TileType:$lhs, TTL_TileType:$rhs,
+                       TTL_TileType:$output,
```
brnorris03, line 1685: Dead code -- I think `output` is not needed any more, lowering converts to `bcast_output_cb_index`, neither used by init insertion pass; should remove the cb index association walk, too.

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -1203,9 +1221,20 @@ class TTKernelToEmitCOpaqueRewriter : public OpConversionPattern<SourceOp> {
...
-        op, resultTypes, getOpName(op), getCallArgs(rewriter, op),
-        getTemplateArgs(rewriter, op), adaptor.getOperands());
+    auto call = emitc::CallOpaqueOp::create(
+        rewriter, op.getLoc(), resultTypes, getOpName(op),
+        getCallArgs(rewriter, op), getTemplateArgs(rewriter, op),
+        adaptor.getOperands());
+    if constexpr (std::is_same_v<SourceOp, ttkernel::CopyTileInitOp>) {
+      // Format reconfiguration restores the FPU zero-substitution default.
+      // Datacopy must preserve signed zero, as the legacy init_sfpu did.
+      emitc::VerbatimOp::create(rewriter, op.getLoc(),
```
brnorris03, line 1231: `CopyTileInitOp` currently lowers to two state changes: `copy_tile_init` and an extra call to the private `_configure_unary_preserve_zero_flag_state_` LLK helper. Only the first is represented in TTKernel IR. Represent the extra state change with a typed TTKernel operation, or include it in a public tt-metal copy-init API contract. Generated code should not contain hardware state changes that are invisible in the IR.

lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
```diff
@@ -119,21 +107,27 @@ static llvm::DenseMap<mlir::TypeID, InitOpInfo> buildComputeToInitMap() {
 #define TTL_FPU_BINARY_TILE_OP(TTL_OP, TILE_OP, TTK_INIT, TTK_COMPUTE)         \
   map[mlir::TypeID::get<ttk::TTK_COMPUTE>()] = {                               \
       [](OpBuilder &b, Location l, Operation *computeOp) {                     \
+        ttk::ReconfigDataFormatOp::create(b, l, computeOp->getOperand(0),      \
```
brnorris03, line 110: Could track the configured source DFB pair separately and reset it at each tile-register region. Right now `add` forllowed by `mul` with the same DFBs repeats `reconfig_data_format` even though the formats are unchanged.

test/python/test_binary_bcast_configuration.py
```diff
@@ -0,0 +1,111 @@
...
+        @ttl.datamovement()
+        def writer():
+            ttl.copy(out_dfb.wait(), out[0:1, 0:8])
+
+    return normalized_broadcast_pairs
+
+
+@pytest.mark.parametrize("dst_full_sync_en", [False, True])
+@pytest.mark.parametrize(
+    "torch_dtype,fp32_dest_acc_en", [(torch.bfloat16, False), (torch.float32, True)]
```
brnorris03, line 70: Would be good to add device coverage for add/sub/mul across row/column/scalar broadcast, with bf16/fp32 and DRAM/L1 tensors? This test covers only row multiply and uses one tolerance for both dtypes (there are test utils that can be used for dtype-specific validation).

lib/Dialect/TTKernel/Transforms/TTKernelInsertInits.cpp
```diff
@@ -2,23 +2,10 @@
...
-//      the compute category (FPU binary vs SFPU/copy/bcast) and derives
-//      input/output CBs from compute and pack ops.
-//   2. Per-op inits: emitted in linear block order whenever the op type
-//      changes (unary SFPU, binary SFPU, minmax, FPU binary). The init
-//      key is (init op TypeID, operand values). An init is inserted only
-//      when the key changes. Tracking resets at sync boundaries.
-//
-// TODO(#329): Emit init_short variants for cheaper re-inits on type switches.
+// Hardware startup is planned once per function before any compute API call.
+// Region and operation changes reconfigure formats and use short pipeline
```
brnorris03, line 6: Thank you for tackling this! So this is a bigger change than I imagined -- affects copy, binary operations, matmul, reduction, transpose, and broadcast, not just binary-broadcast fusion alone, so it may be good to extract in its own PR  that can be merged before this one, WDYT?

There is also some duplicate instructions being added -- for L1 recurrence loops, source reconfiguration and copy initialization are emitted before the loop by region planning, and the per-op traversal emits them inside every iteration (see `accumulate_add_recurrence_pipeline.mlir`). I think they shouldn't be emitted per-iteration in this case.

brnorris03 (approved):
A few more suggestions, I don’t feel strongly about the separate PR if you prefer to keep this together.

PR #948, opened 2026-08-20
[ttl] Support block squeeze and unsqueeze views

### LINKED ISSUE #917: [ttl] ttl.block.squeeze and ttl.block.unsqueeze are specified but unimplemented
## Problem

`ttl.block.squeeze` and `ttl.block.unsqueeze` are specified but not implemented by the compiler. Kernels using them fail at compile time:

```text
TTLangCompileError: error: Unknown function: ttl.block.squeeze
```

The spec defines both (`TTLangSpecification.md` L1517-L1518):

> `ttl.block.squeeze(expr: ttl.BlockExpr, dims: List[int]) -> ttl.BlockExpr` -- Remove shape dimension at positions specified by `dims`. Removed shape dimension must be 1.
> `ttl.block.unsqueeze(expr: ttl.BlockExpr, dims: List[int]) -> ttl.BlockExpr` -- Add shape dimension of 1 at positions specified by `dims`.

They were added in spec revision 0.16 and moved to `ttl.block` in 0.17 (L52-L53), and the spec's own worked examples use them (L674, L694).

The absence is deliberate and already documented in `python/ttl/ttl_block.py` L13-L14:

```python
"""Only the operations the compiler implements appear here. ``mask``,
``mask_posinf``, ``where``, ``squeeze`` and ``unsqueeze`` are specified but not
yet implemented, and are absent rather than bound to a stub."""
```

## Why this should be cheap

The simulator implements both as pure shape reinterpretation (`python/sim/block.py` L341 and L394) -- the tile list is preserved and only the shape tuple changes:

```python
new_shape = tuple(s for i, s in enumerate(block_shape) if i not in norm_dims)
tiles = block.to_list()
result = Block.from_list(tiles, new_shape)
```

No data movement is required. The open question is whether the compiler can express this as a metadata-only operation on a CB-attached block, or whether it needs a new op with a lowering that is a no-op on the data path.

## Affected kernels

14 of 100 in the tt-blacksmith codegen corpus:

01, 02, 08, 23, 25, 26, 27, 28, 35, 45, 47, 51, 56, 60

Full names: `01_attention_scaled_dot_product_attention_128x128`, `02_softmax_s_d_p_attention_128x512`, `08_batched_matmul_feat_exemplar_avg_block_256x256`, `23_attention_attention_128x128`, `25_attention_scaled_dot_product_attention_256x256`, `26_attention_scaled_dot_product_attention_256x256`, `27_attention_scaled_dot_product_attention_256x256`, `28_attention_scaled_dot_product_attention_256x256`, `35_softmax_scalar_mix_128x128`, `45_batched_matmul_simple_bmm_module_256x256`, `47_batched_matmul_l2_distance_attention_256x256`, `51_batched_matmul_feat_exemplar_avg_block_256x256`, `56_matmul_attention_128x128`, `60_matmul_simple_matmul_module_128x128`

FConstantinos:
### Problem description
Generated Blacksmith kernels call `ttl.block.squeeze`, but the block namespace does not currently expose squeeze/unsqueeze shape views. These kernels stop during tracing before reaching their later compiler or runtime behavior.

### What changed
- Add `ttl.block.squeeze` and `ttl.block.unsqueeze` as zero-copy block shape views.
- Support positive and negative dimensions, multiple dimensions, duplicate handling, and empty-dimension no-ops.
- Validate squeeze dimensions and preserve element type, encoding, and tile order.
- Add focused namespace, simulator, and hardware regression coverage.

### Ticket
Fixes #917.

### Stack
Depends on #941. The base branch is `kostas/unsupported-fill`.

### Checklist

### Validation
- Focused namespace and hardware tests: 41 passed.
- Squeeze/unsqueeze simulator subset: 14 passed.
- `check-ttlang`: 314 MLIR, 174 packaging, and 3 bindings tests passed.
- Four iterative review/fix passes completed; the final pass found no actionable issues.
- Pre-commit and `git diff --check` passed.
- Blacksmith sweep improved from 22/100 to 23/100 hardware passes with no regressions; no kernel still reports an unknown squeeze/unsqueeze function.


python/ttl/operators.py
```diff
@@ -1052,6 +1052,107 @@ def broadcast(input: TensorBlock, *, dims: List[int], shape) -> TensorBlock:
...
 
+def _block_shape_view(input: TensorBlock, result_shape: List[int]) -> TensorBlock:
+    """Create a zero-copy block view with a different grid rank."""
+    from ttl.dialects import builtin
+
+    input_type = input.type
+    result_type = RankedTensorType.get(
+        result_shape, input_type.element_type, input_type.encoding
+    )
+    return builtin.UnrealizedConversionCastOp([result_type], [input]).result
```
brnorris03: I don't think unrealized conversion casts are the best way to implement this. It works because of implicit compiler conventions, which could change and so this could easily break downstream.

The problem is that `builtin.unrealized_conversion_cast` does not express the semantics of this operation (MLIR defines it as a temporary bridge between type systems and gives it no representation or execution semantics). Its verifier does not require identical encodings, identical element types, shared storage, or dimension sequences that differ only by unit dimensions. It would be easy to produce invalid IR that would fail nontrivially in later analyses or conversions.

I would define an explicit operation, something like this:

```tablegen
def TTL_BlockViewOp : TTL_Op<"block.view", [
    Pure, SameOperandsAndResultElementType, ViewLikeOpInterface]> {
  let arguments = (ins AnyStaticShapeTensor:$input);
  let results = (outs AnyStaticShapeTensor:$result);
  let assemblyFormat =
      "$input attr-dict `:` type($input) `->` type($result)";
  let hasVerifier = 1;
  let hasFolder = 1;
  let extraClassDeclaration = [{
    ::mlir::Value getViewSource() { return getInput(); }
  }];
}
```

The verifier should require identical encodings and identical dimension sequences after removing unit dimensions. Both frontend functions should create this operation instead of unrealized casts. Also the folder should remove identity views. Implementing `ViewLikeOpInterface` lets the existing DFB provenance logic follow the view without adding operation-specific handling.

test/python/test_squeeze.py
```diff
@@ -0,0 +1,305 @@
...
+    def dm_read():
+        pass
+
+    @ttl.datamovement()
+    def dm_write():
+        pass
+
+
+def _make_tensors(device):
+    source = torch.arange(6 * 32 * 32, dtype=torch.bfloat16).reshape(1, 1, 64, 96)
```
brnorris03: Can you also add BF16/FP32 and DRAM/L1 coverage (because these views preserve dtype and backing storage)?

brnorris03 (changes requested):
Thank you for adding this! I think a new op is warranted here, see inline comments (unrealized casts are not really appropriate, IMO).

python/ttl/operators.py
brnorris03: Actually, thinking a bit more: this does not require a new TTL op. `tensor.expand_shape` and `tensor.collapse_shape` already express this kind of ranked tensor reassociation, with upstream verification, folding, and canonicalization. `unsqueeze` should lower to `tensor.expand_shape` and `squeeze` to `tensor.collapse_shape`. TT-Lang still needs to recognize the DFB-preserving subset: static shapes, identical element types and encodings, and only singleton dimensions inserted or removed. That avoids both a custom op and using `builtin.unrealized_conversion_cast` for semantics it does not define.

brnorris03: On second thought, this should not require a new TTL op. You can replace `_block_shape_view` and its `builtin.unrealized_conversion_cast` with `tensor.expand_shape` for `unsqueeze` and `tensor.collapse_shape` for `squeeze`. Update the shared DFB provenance utility, including `getAttachedCB`, to trace these operations only when the source and result are static, retain the same element type and encoding, and differ only by inserted or removed size-one dimensions. Also should add IR tests showing that these views preserve DFB association while a general reshape does not.

FConstantinos: done

test/python/test_squeeze.py
FConstantinos: done

docs/development/BlockShapeViews.md
```diff
@@ -0,0 +1,49 @@
...
+Inserting or removing a size-one dimension does not change the row-major
+linear index of any tile. The predicate therefore preserves tile order,
+element representation, and the acquired DFB slot without allocation or data
+movement. MLIR's expand/collapse verifier additionally checks the reassociation
+between source and result dimensions. Rank-zero views are accepted only when
+the other shape consists entirely of size-one dimensions.
+
+`traceDFBShapeViews` composes checked shape views with identity casts and
+one-to-one CB conversion bridges with matching element types. Arbitrary
+tensor-to-tensor reinterpretation casts do not establish storage identity.
```
brnorris03, line 27: Nit: these couple of sentences are kind of cryptic, I am not sure about what "bridges" are? I think it can be made a bit more precise (less shorthandy) fairly easily.

include/ttlang/Dialect/TTL/IR/TTLOpsUtils.h
```diff
@@ -200,16 +200,40 @@ inline mlir::Value traceUnrealizedCasts(mlir::Value value) {
...
+/// operations terminate the search because their results may not alias the
+/// acquired slot. Returns null when the chain does not end at `ttl.cb_wait` or
 /// `ttl.cb_reserve`.
 inline mlir::Operation *findCBAcquireOp(mlir::Value tensor) {
   while (true) {
-    tensor = traceUnrealizedCasts(tensor);
+    tensor = traceDFBShapeViews(tensor);
     if (auto attach = tensor.getDefiningOp<AttachCBOp>()) {
       tensor = attach.getTensor();
       continue;
```
brnorris03: Would be nice to express this as an op interface rather than a hand-maintained chain -- `attach_cb` already describes itself as an identity on the tensor value, so a `getAliasSource()` on it and the shape view/slice/extract ops would collapse this into one uniform peel and pick up future alias-preserving ops without another clause here. I would use a TTL-local interface along the lines of `DFBAccessOpInterface` rather than upstream `ViewLikeOpInterface`, which is buffer-oriented and unused by the Tensor dialect. Thanks!

brnorris03, line 239: Thank you for reusing the tensor ops! Now that you have them, it would be nice to express this as an op interface rather than having manual tracing whenever needed. I would define `TTL_StorageAliasOpInterface` in `include/ttlang/Dialect/TTL/IR/TTLInterfaces.td` right next to `TTL_DFBAccessOpInterface` and in the same shape -- `cppNamespace = "::mlir::tt::ttl"` plus a single `InterfaceMethod` returning `::mlir::Value getStorageAliasSource()`, documented as the operand that preserves the acquired slot identity. `attach_cb` can then declare it directly in `TTLOps.td` (it currently carries `[]`) returning `getTensor()`, and both this loop and `traceDFBShapeViews` collapse into one peel:

```cpp
while (auto alias = tensor.getDefiningOp<StorageAliasOpInterface>())
  tensor = alias.getStorageAliasSource();
```

The wrinkle is that four of the five cases are upstream ops -- the checked expand/collapse views, `tensor.extract_slice`, `tensor.extract`, and the cast bridge -- so those would need `attachInterface` external models registered in the TTL dialect's `initialize()`, which we do not do anywhere yet. If you would rather not introduce that, folding the two nested fixed-point loops into a single `getStorageAliasSource(Operation *)` free function with one loop here would still remove the alternation and give one place to add a case. I think I actually would prefer this now (if you agree), maybe add a TODO for potentially adding an interface later.

brnorris03 (approved):
An optional small refactoring suggestion, otherwise LGTM! Thank you for all the updates!

docs/development/BlockShapeViews.md
FConstantinos: done on #949

PR #949, opened 2026-08-20
[ttl] Support batched block matmul

FConstantinos:
### Problem description
TT-Lang block matmul accepts only rank-2 operands, while the specification's leading `I...` dimensions permit equal-rank batched operands. Generated Blacksmith kernels using rank-3 or rank-4 matmul therefore fail before lowering.

### What changed
- Accept equal-rank matmul operands of rank 2 or greater with matching leading batch dimensions.
- Preserve the specification's non-broadcasting batch contract.
- Lower one matmul block per batch with distinct input offsets and DST slots.
- Support transposed RHS operands, fused accumulators, and nested matmul slice offsets.
- Add Python, verifier, TTL-to-compute, and TTKernel lowering coverage.

### Stack
Depends on #948. The base branch is `kostas/block-squeeze`.

### Checklist

### Validation
- Rank-3 standalone and rank-4 fused batched matmul hardware tests passed.
- `check-ttlang`: 314 MLIR, 174 packaging, and 3 bindings tests passed.
- Four iterative review/fix passes completed.
- Focused verifier and lowering tests and all pre-commit checks passed.
- The 21 kernels formerly rejected by the rank-2 restriction all advance beyond rank validation.


lib/Dialect/TTL/Transforms/LowerMatmulCompute.cpp
```diff
@@ -253,6 +255,50 @@ static Value createConstantIndex(OpBuilder &builder, Location loc,
   return arith::ConstantIndexOp::create(builder, loc, value);
 }
 
+static SmallVector<int64_t> delinearizeStaticIndex(int64_t linearIndex,
```
brnorris03: You can reuse the upstream `mlir::delinearize(linearIndex, mlir::computeStrides(shape))` instead of defining this.

test/python/test_batched_matmul.py
```diff
@@ -0,0 +1,121 @@
...
+    for coordinates, factor in zip(
+        torch.cartesian_prod(*(torch.arange(size) for size in batch_shape)),
+        factors,
+    ):
+        index = tuple(int(coordinate) for coordinate in coordinates)
+        result[index] = identity * factor
+    return result
+
+
+def test_rank3_batched_matmul(device):
```
brnorris03: Can you also add batched `transpose_rhs=True` coverage and parameterize BF16/FP32 and DRAM/L1 to cover the new transposed-stride and nested-offset logic?

brnorris03 (commented):
Looks good in general, some more test ideas:
  1. Nontrivial rank-3 hardware `matmul`, such as `[2,2,3] x [2,3,2]`.
  2. Batched transpose with `K > 1` and `N > 1`.
  3. BF16/FP32 and L1/DRAM parameterization.
  4. One valid non-default subtile configuration.
  5. A TTKernel lit case for batched transpose offsets.

lib/Dialect/TTL/Transforms/LowerMatmulCompute.cpp
FConstantinos: done

FConstantinos:
> Looks good in general, some more test ideas:
> 
> 1. Nontrivial rank-3 hardware `matmul`, such as `[2,2,3] x [2,3,2]`.
> 2. Batched transpose with `K > 1` and `N > 1`.
> 3. BF16/FP32 and L1/DRAM parameterization.
> 4. One valid non-default subtile configuration.
> 5. A TTKernel lit case for batched transpose offsets.

done

test/python/test_batched_matmul.py
FConstantinos: done

test/python/test_batched_matmul.py
```diff
@@ -0,0 +1,140 @@
...
+def test_rank4_batched_matmul_fused(device):
+    a_torch = _diagonal_batches((2, 2), (1, 2, 3, 4))
+    b_torch = _diagonal_batches((2, 2), (5, 6, 7, 8))
+    bias_torch = torch.arange(4, dtype=torch.bfloat16).reshape(2, 2, 1, 1)
+    bias_torch = bias_torch.expand(2, 2, 32, 32).clone()
+    expected = torch.matmul(a_torch, b_torch) + bias_torch
+    a = to_l1(a_torch, device)
+    b = to_l1(b_torch, device)
+    bias = to_l1(bias_torch, device)
+    out = to_l1(torch.zeros_like(expected), device)
```
brnorris03: Can you parameterize this one as well similarly to `test_rank3_batched_matmul`, thanks!

docs/development/BlockShapeViews.md
```diff
@@ -0,0 +1,49 @@
+# Block Shape Views
+
+`ttl.block.squeeze` and `ttl.block.unsqueeze` remove or insert size-one block
+grid dimensions. The frontend emits `tensor.collapse_shape` and
+`tensor.expand_shape`, respectively, with explicit reassociation indices.
+An empty dimension list returns the original value. The public dimension
+normalization and diagnostic rules remain unchanged.
+
```
brnorris03: Does this need to be its own file -- it is 49 lines and the content is really about DFB storage identity, so folding it into `DFBManagement.md` or `ComputeOpCreation.md` would keep it next to the material it depends on. If it does stay standalone it needs an entry in `docs/sphinx/implementation.md`, which is the index for these docs and is not updated here. Worth noting that index has drifted independently of this PR: it lists 7 of the 13 files currently in `docs/development/`, and the six it misses include `DFBManagement.md`, `PipeNets.md`, and `PipesOnFabric.md`. Thanks!

brnorris03 (approved):
One test parameterization request, otherwise LGTM!

PR #953, opened 2026-08-22
[ttl] Materialize singleton shape-view stores through DFBs

FConstantinos:
### Problem description

Compiler DFB materialization must recognize stores through zero-copy block shape views without treating arbitrary equal-tile-count reshapes as valid views.

The exact-shape simulator enforcement, specification wording, and simulator regressions originally carried by this branch landed independently in #952 (`42dbc08724`). They are therefore intentionally absent from the current PR diff and remain covered on `main`.

Depends on #950.

### What changed

- Materialize `ttl.store` sources through DFB-backed singleton-dimension shape views.
- Restrict zero-copy DFB shape-view handling to insertion or removal of singleton dimensions.
- Preserve the compiler requirement that store source and destination ranks and extents match exactly.
- Add compiler regressions for valid singleton-dimension views and invalid equal-tile-count reshapes.
- Add compute-planning and disabled-materialization regression coverage.
- Document singleton-view DFB materialization behavior.

### Tests retained

- Simulator rank-mismatch and same-rank extent-mismatch coverage remains on `main` in `test/sim/test_dfb.py` from #952.
- This PR retains its focused MLIR coverage:
  - `compute_op_creation_materialization.mlir`
  - `dfb_materialization_shape_views.mlir`
  - `insert_intermediate_dfbs_disabled_invalid.mlir`

### Validation

- `git diff --check`: passed.
- `pre-commit run --all-files`: passed.
- Build: passed.
- MLIR suite: 483 passed, 2 unsupported.
- Focused PR Python tests: 43 passed.
- Focused simulator tests: 249 passed.
- Packaging tests: 182 passed.

The broader Python lit target is currently blocked on the validation host because it drops `USER` and attempts to write generated kernels beneath another user's unwritable `/tmp/default` directory.

### Checklist


FConstantinos:
Validation and Blacksmith rerun complete.

- Review iterations: 2/4. Iteration 1 found and fixed overly broad equal-element-count shape-view handling; iteration 2 found no further issues.
- Regression test: confirmed fail-before/pass-after.
- `check-ttlang`: 323 MLIR + 174 packaging + 3 Python binding tests passed.
- `pytest test/sim`: 1191 passed, 8 skipped, 2 xfailed.
- `pre-commit run --all-files`: passed.
- Blacksmith simulator: 86/100 pass; failures are 13 exact store-shape mismatches (4, 8, 13, 25, 28, 31, 36, 45, 47, 51, 71, 75, 89) plus pre-existing numerical mismatch 58.
- Blacksmith hardware/compiler, isolated with reset-after-failure: 30/100 pass. The compiler rejects exactly the same 13 store-shape-invalid kernels as the simulator.

The initial multi-worker hardware attempt encountered stale/wedged card state. Kernel 2 passed 3/3 immediately after `tt-smi -r --no_reinit`; the reported hardware results therefore use isolated per-kernel runs so a stalled kernel could not contaminate subsequent outcomes.

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
```diff
@@ -8,11 +8,42 @@
...
+                                          RankedTensorType resultType) {
+  auto isNotSingleton = [](int64_t extent) { return extent != 1; };
+  return llvm::equal(
+      llvm::make_filter_range(inputType.getShape(), isNotSingleton),
+      llvm::make_filter_range(resultType.getShape(), isNotSingleton));
+}
+
+Value getDFBMaterializationStoreSource(Value intermediate) {
+  Value source = intermediate;
+  while (auto cast = source.getDefiningOp<UnrealizedConversionCastOp>()) {
```
brnorris03: This infers view semantics from `builtin.unrealized_conversion_cast`, which is a generic conversion placeholder and can misclassify unrelated casts. Here I'd change this to work with the `tensor.expand_shape` and `tensor.collapse_shape` operations introduced by #948 instead if you agree. You could possibly use a shared helper to verify static, encoding-preserving singleton-dimension changes, and record the accepted view chain in the materialization plan rather than rediscovering it during application.

docs/development/ComputeOpCreation.md
```diff
@@ -568,6 +571,13 @@ attached storage may be released before the consumer. With compiler DFBs
 disabled, either condition produces a diagnostic instead of changing the
 lifetime result.
 
+Singleton-rank `squeeze` and `unsqueeze` operations lower to zero-copy
+`unrealized_conversion_cast` shape views. A direct store of a computed shape
```
brnorris03: This is implemented in #948 so should be documented there (and perhaps updated if you agree with the suggested implementation change there).

docs/development/DFBManagement.md
```diff
@@ -459,6 +459,12 @@ attach_cb                            // tensor SSA view of the waited slot
 cb_pop                               // occupancy: 1 -> 0
 ```
 
+For a zero-copy tensor shape view, the producer-side reserve may use the
+producer rank with an explicit tile count while `attach_cb` exposes the same
+published DFB at the consumer-visible rank. This is valid only when the view
+inserts or removes singleton dimensions while preserving the static element
+type and encoding.
```
brnorris03: I think this would be clearer with a concrete example and without describing rank as a property of the DFB storage, e.g.

```suggestion
A compiler-created DFB may expose the same tile sequence through tensor types
that differ only by singleton dimensions. For example, a producer may store
`tensor<2x2xtile>` into a four-tile reservation that the consumer attaches as
`tensor<1x2x2xtile>`. The compiler accepts this without data rearrangement only
for static types with identical element types and encodings. The compiler does
not treat arbitrary reshapes, such as `tensor<2x2xtile>` to
`tensor<1x4xtile>`, as zero-copy DFB views, even though both types contain four
tiles.
```

include/ttlang/Dialect/TTL/Passes.td
```diff
@@ -486,6 +488,10 @@ def TTLInsertIntermediateDFBs
     producer materializes the exact consumer operand. Multiple reserve
     transactions for one result, and direct stores of one value in multiple
     blocks, materialize all relevant uses through one shared compiler DFB.
+    A direct store of a computed singleton-dimension tensor shape view
+    materializes its producer tiles through the producer-rank reserve view and
+    preserves the store's consumer-visible tensor rank when attaching the
+    compiler DFB.
```
brnorris03: "computed singleton-dimension tensor shape view" is kind of a mouthful -- maybe break it up some?

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
```diff
@@ -231,6 +231,23 @@ static void applyStandaloneMaterializationPlan(
...
+    auto cast = view.getDefiningOp<UnrealizedConversionCastOp>();
+    assert(cast && cast.getInputs().size() == 1 &&
+           "materialization view trace must contain single-input casts");
+    Value input = cast.getInputs()[0];
+    if (!cast->use_empty()) {
+      break;
+    }
+    cast.erase();
+    view = input;
+  }
```
brnorris03: This mutation-time traversal rediscovers view policy after the IR has changed and assumes that every accepted view is an `UnrealizedConversionCastOp`. You could extend the `StandaloneDFBMaterializationPlan` to record the producer value to store and the exact `tensor.expand_shape`/`tensor.collapse_shape` operations that are dead after the planned operand rewrites. So first validate those records before mutating IR, then erase the recorded views outermost-first with `IRRewriter` after rewriting the operands. In general here I don't think the view chain should be reconstructed or the planning semantic checks repeated.

test/ttlang/Dialect/TTL/Transforms/dfb_materialization_shape_views.mlir
```diff
@@ -0,0 +1,73 @@
...
+      : <[2, 2], !ttcore.tile<32x32, bf16>, 2>
+        -> tensor<2x2x!ttcore.tile<32x32, bf16>>
+  %input = ttl.attach_cb %input_wait, %input_dfb
+      : (tensor<2x2x!ttcore.tile<32x32, bf16>>,
+         !ttl.cb<[2, 2], !ttcore.tile<32x32, bf16>, 2>)
+        -> tensor<2x2x!ttcore.tile<32x32, bf16>>
+  %negative = ttl.neg %input
+      : tensor<2x2x!ttcore.tile<32x32, bf16>>
+        -> tensor<2x2x!ttcore.tile<32x32, bf16>>
+  %reshape = builtin.unrealized_conversion_cast %negative
```
brnorris03: This negative test does not cover a view applied directly to a DFB-attached value. I reproduced `tensor<2x2xtile> -> tensor<1x4xtile>` from `%input` through this pass pipeline; it succeeds because `findCBAcquireOp`/`getAttachedCB` traces every one-input `UnrealizedConversionCastOp`, so `addShapeViewStoreRequirements` considers the value already DFB-backed. Please apply the shared singleton-dimension view check in DFB provenance and add a `*_invalid.mlir` full-pipeline test for this direct case. The test should require a diagnostic, not only the absence of `reason=store-input-shape-view`.

test/ttlang/Dialect/TTL/Transforms/compute_op_creation_materialization.mlir
```diff
@@ -260,3 +260,105 @@ func.func @two_transactions_and_elementwise_use()
...
+// A zero-copy singleton-rank view has no tile recipe of its own. Materialize
+// its computed input through a reserve view with the producer's rank, then
+// attach the same DFB with the consumer-visible rank for the final passthrough.
+// PLAN-LABEL: ComputeOp creation plan @computed_shape_view_store
+// PLAN:       unassigned-store
+// PLAN:       operand=0
+// PLAN-NEXT:  reason=store-input-shape-view
+// FULL-LABEL: func.func @computed_shape_view_store
+// FULL:       %[[INTERMEDIATE_DFB:.*]] = ttl.bind_cb{{.*}} {ttl.compiler_allocated}
+// FULL:       %[[INTERMEDIATE_RESERVE:.*]] = ttl.cb_reserve %[[INTERMEDIATE_DFB]] {num_tiles = 4 : i64}
```
brnorris03: This test covers only computed insertion of singleton dimensions. Please add computed removal and a device pytest for both directions. The inherited `test_squeeze.py` cases apply the view directly to a `cb_wait` result, so they never execute the compiler-created reserve/store/push/wait lifecycle added here. Parameterize the device test over bf16 and fp32; the view operation is not dtype-restricted.

include/ttlang/Dialect/TTL/Transforms/DFBMaterialization.h
```diff
@@ -17,6 +17,13 @@
 
 namespace mlir::tt::ttl {
 
+/// Returns the value whose tiles must be published when `intermediate` is
+/// materialized. Unrealized tensor casts that only insert or remove singleton
+/// dimensions are zero-copy block-shape views, so their input can be stored
+/// through a reserve view with the input rank while consumers retain the
+/// original attached tensor type.
+Value getDFBMaterializationStoreSource(Value intermediate);
```
brnorris03: This seems like too much detail (in the comment). How about:

```suggestion
/// Returns the producer value underlying zero-copy singleton-dimension tensor
/// views, or `intermediate` when there are none.
Value getDFBMaterializationStoreSource(Value intermediate);
```

lib/Dialect/TTL/Transforms/DFBMaterialization.cpp
FConstantinos: done

docs/development/ComputeOpCreation.md
FConstantinos: done

docs/development/DFBManagement.md
FConstantinos: done

include/ttlang/Dialect/TTL/Passes.td
FConstantinos: done!

lib/Dialect/TTL/Transforms/TTLInsertIntermediateDFBs.cpp
FConstantinos: Removed the custom bypass/cleanup path. Now uses the planned producer operand, made by existing typed plans

test/ttlang/Dialect/TTL/Transforms/dfb_materialization_shape_views.mlir
FConstantinos: done

test/ttlang/Dialect/TTL/Transforms/compute_op_creation_materialization.mlir
FConstantinos: done

include/ttlang/Dialect/TTL/Transforms/DFBMaterialization.h
FConstantinos: simplified

docs/development/ComputeOpCreation.md
```diff
@@ -568,6 +571,17 @@ attached storage may be released before the consumer. With compiler DFBs
 disabled, either condition produces a diagnostic instead of changing the
 lifetime result.
 
+The [block shape-view contract](BlockShapeViews.md) defines the checked views
+accepted by shared DFB provenance. A direct store of a view of a computed
+expression cannot become a tile recipe because the view itself performs no
+compute. Intermediate DFB planning records the innermost view's source operand
```
brnorris03: The "A direct store of a view of a computed
expression cannot become a tile recipe because the view itself performs no
compute." sentence is kind of weird, not sure what is meant by "tile recipe" here? A view obviously is not compute, so stating that seems superfluous, too.

PR #957, opened 2026-08-25
Add optional compiler validation before simulation

FConstantinos:
### Problem description

The functional simulator can accept and execute kernels that the compiler later rejects. This weakens the simulator's role as an early correctness gate, especially for generated kernels, because compiler frontend and TTL pipeline diagnostics are not available until a separate compilation attempt.

Compiler validation remains optional so the standalone `tt-lang-sim` package continues to work without the compiler's compiled MLIR Python extensions, TT-NN, or Tenstorrent hardware.

### What's changed

- Add a host-only compiler validation entry point based on static tensor descriptors.
- Split the compiler flow into explicit TTL IR construction, validation, runtime lowering, and artifact-construction phases. Simulator validation returns after the shared validation phase; normal compilation continues through TTKernel and EmitC lowering.
- Add simulator validation modes:
  - `off` keeps simulator-only behavior and remains the default.
  - `auto` validates when the compiler is available and otherwise continues with one warning.
  - `required` fails when compiler validation cannot run. Passing bare `--compiler-validation` selects this mode.
- Add offline Blackhole and Wormhole B0 target selection for architecture-dependent checks. Target selection does not probe hardware.
- Keep mode and target configuration in the simulator's per-run context. No new public environment variables are introduced.
- Preserve compiler/simulator import isolation, direct and captured tensor aliases, successful-analysis caching, and thread safety.
- Support explicit multi-kernel and unified-body operations without threading validation booleans through normal compiler decorators.
- Propagate all compiler-relevant tensor metadata modeled by the simulator: logical and padded shape, dtype, tiled versus row-major layout, tile geometry and byte size, DRAM versus L1 placement, memory layout, regular and ND shard metadata, core ranges, and mesh distribution.
- Include the complete descriptor and tensor aliasing in successful-validation cache keys.
- Reject compiler-unsupported metadata explicitly instead of silently substituting tiled/interleaved defaults. This currently includes row-major TTNN interop and ND sharding, which the TTL dialect does not represent.
- Preserve requested memory configuration in the simulator's `rand`, `empty`, and `zeros` factories.

### Production hardening

- Define and check a versioned simulator/compiler bridge contract. Mixed incompatible installations fail with an actionable diagnostic instead of attempting validation with a mismatched descriptor format.
- Validate tensor descriptor invariants at construction, including shapes, padding, tiling, placement, sharding, core ranges, orientation, distribution, and mesh dimensions.
- Identify simulator tensors with an explicit private protocol marker rather than module-name heuristics, while excluding class objects.
- Restrict TTL symbol rebinding to non-scalar objects so interned strings, numbers, booleans, tuples, and `None` cannot cause accidental global substitutions.
- Reset CLI context before applying options so repeated embedded invocations have process-equivalent lifecycle behavior.
- Centralize supported modes and targets between validation configuration and CLI parsing.
- Document the full metadata contract, cache lifetime, compatibility behavior, and remaining limitations.
- Rebase the complete feature onto current `main` before final verification.

The simulator currently models only the standard 32x32 tile and cannot know the amount of L1 available on a live device. Runtime behavior such as deadlocks, numerical disagreement, and data-dependent failures remains the responsibility of simulation and device testing.

### Validation

- `pre-commit run --all-files`: all hooks passed.
- `cmake --build build --target check-ttlang` on current `main`:
  - 392 supported MLIR tests passed; 1 unsupported.
  - 3 Python binding tests passed.
  - 174 packaging tests passed.
- Focused static-analysis, cache, tensor-backed DFB, and bridge tests: 106 passed.
- `pytest -q test/sim`: 1115 passed, 8 skipped, 2 expected failures.
- Sphinx HTML documentation build succeeded with only the 51 pre-existing warnings.
- `examples/eltwise_add.py` passed with bare required compiler validation, including a height-sharded L1 tensor case.
- Normal compilation and execution of `examples/eltwise_add.py` passed on Blackhole silicon.
- A 100-kernel differential sweep with validation disabled produced the same result on the baseline and this branch: 95 passes and the same five existing simulator failures.
- With validation required, 90 kernels were rejected by compiler diagnostics before simulation, 8 passed validation and simulation, and 2 passed validation but retained independent simulator failures.
- A 40-case metadata sweep covered tiled/row-major layout, L1/DRAM placement, all simulator memory layouts, and aliased/distinct arguments. All 16 compiler-supported tiled configurations passed. The 20 row-major and 4 ND-sharded configurations were explicitly rejected instead of being incorrectly validated under defaults.
- Four requested review iterations were completed with no open findings, followed by the production-hardening audit above.

### Checklist

- [x] New and existing tests provide coverage for success, failure, compatibility, caching, and lifecycle behavior.
- [x] User-facing documentation is included.
- [x] The standalone simulator remains independent of the compiler by default.
- [x] The branch is rebased onto current `main` and verified on the requested Blackhole host.


python/sim/compiler_validation.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+def _to_compiler_value(value: Any, backend: _CompilerBackend, memo: dict[int, Any]):
+    """Convert a direct simulator tensor while preserving argument aliases."""
+    if not _is_sim_tensor(value):
+        return value
+    identity = id(value)
+    if identity not in memo:
+        memo[identity] = backend.tensor_spec_type(
+            value.shape,
```
brnorris03, line 293: Suggest deriving per-device dimensions from `mesh_shard_info` instead of passing the simulator’s global `value.shape`. Also reject non-divisible mappings and test 1D and 2D sharding.

python/ttl/ttl_api.py
```diff
@@ -2894,6 +3100,62 @@ def _wrapper(*args, **kwargs):
...
+        resolved_grid = _resolve_grid(grid, runtime_args, {})
+
+        env_opts = os.environ.get("TTLANG_COMPILER_OPTIONS")
+        if env_opts:
+            opts_str = f"{opts_str or ''} {env_opts}".strip() or None
+        compiler_options = CompilerOptions.from_string(opts_str).merge(
+            CompilerOptions.from_argv()
+        )
+        l1_budget_override = _resolve_l1_budget(runtime_args, compiler_options)
+        cache_key = _make_cache_key(
```
brnorris03, line 3133: Maybe should include capture/argument aliasing in this cache key. Otherwise, a successful aliased call can skip validation of a later non-aliased call with the same metadata. Also need a test.

python/ttl/ttl_api.py
```diff
@@ -552,6 +570,33 @@ def _detect_memory_space_from_tensor(tensor, default: str) -> str:
...
+    for index, arg in enumerate(args):
+        if not is_tensor_value(arg):
+            continue
+        memory_space = _detect_memory_space_from_tensor(arg, "unknown")
+        if memory_space not in ("L1", "DRAM"):
+            raise ValueError(
+                "TTNN interop requires L1 or DRAM memory space, but tensor "
+                f"{index} is in {memory_space}."
+            )
+        if hasattr(arg, "layout") and "TILE" not in str(arg.layout):
```
brnorris03, line 592: Validate `memory_config.memory_layout` for every argument here?

python/ttl/ttl_api.py
```diff
@@ -552,6 +570,33 @@ def _detect_memory_space_from_tensor(tensor, default: str) -> str:
...
+            f"TTNN interop requires all tensors to be the same type. "
+            f"Got {ttnn_count} TTNN tensors and {len(args) - ttnn_count} host tensors. "
+            f"Mixed tensor types would generate extra bounce kernels."
+        )
+
+    for index, arg in enumerate(args):
+        if not is_tensor_value(arg):
+            continue
+        memory_space = _detect_memory_space_from_tensor(arg, "unknown")
+        if memory_space not in ("L1", "DRAM"):
```
brnorris03, line 587: Are mixed DRAM/L1 arguments allowed? I am not completely sure, but I think DRAM/L1 args shouldn't be mixed here since compilation uses the first tensor’s memory space for every tensor (that may change in future), so later tensors may use the wrong addressing.

python/ttl/ttl_api.py
```diff
@@ -2480,297 +2525,421 @@ def _lower_program_to_kernel(
...
+        compiler_options=compiler_options,
+        l1_budget_override=l1_budget_override,
+        kernel_source_file=kernel_source_file,
+        kernel_line_offset=kernel_line_offset,
+        logical_kernels=logical_kernels,
+    )
+    runtime_passes = _runtime_lowering_pipeline_passes(
+        compiler_options, l1_budget_override
+    )
+    _run_ttl_pipeline(program_ir, runtime_passes)
```
brnorris03, line 2879: Do regular compilations need to run these two pass lists through separate pass managers? Host-only validation only needs the validation prefix; regular compilation could still run `validation_passes + runtime_passes` in one `_run_ttl_pipeline` call. This would preserve the existing pass-manager behavior while allowing host-only validation to stop before runtime lowering.

python/ttl/ttl_api.py
```diff
@@ -2851,10 +3058,9 @@ def _wrapper(*args, **kwargs):
                     l1_budget_override,
                     runtime_resource_cache,
                 )
-                if compiled_kernel is not None:
-                    cache[cache_key] = compiled_kernel
+                cache[cache_key] = compiled_kernel
```
brnorris03, line 3061: `_compile_ttnn_kernel` still returns `None` when TTNN is unavailable, then caches that value as a compiled kernel and later calls it -- but `None` is not a valid kernel; I'd keep the guard or replace the `None` return with an explicit error.

brnorris03 (commented):
Thank you for enabling this!

Issue #819, opened 2026-08-07
[ttl.exp] Reject unsupported iterations before kernel lowering


brnorris03:
### Problem description

PR #699 exposes the tt-metal `exp_tile` `iterations` template argument through
`ttl.exp`, but neither the TTL operation nor the Python frontend validates it.
Values including `0`, `-1`, `4`, and `2147483647` compile without a diagnostic
and reach `ttkernel.exp_tile` unchanged.

The merged operation defines `iterations` as a default-valued `i32` without a
verifier:

https://github.com/tenstorrent/tt-lang/blob/823f555e4944c92ef236744a191009ce6ff663df/include/ttlang/Dialect/TTL/IR/TTLOps.td#L1150-L1164

The Python implementations only require a constant integer. The specification
added by #699 describes the argument as `ttl.PositiveInt`, but positivity is not
enforced either.

This is unsafe at the hardware boundary. At the tt-metal revision used by the
merge:

- The API documents `iterations` as a positive integer.
- Blackhole and Wormhole replay implementations accept only 8 or 32 for one
  approximate strategy and fail with a `static_assert` otherwise.
- Other strategies iterate `dst_reg` exactly `ITERATIONS` times. Too few
  iterations leave part of the tile unevaluated; too many can advance beyond
  the tile.

Relevant tt-metal sources:

- https://github.com/tenstorrent/tt-metal/blob/e908c31332b60860ed0d4186452dc880cdd5a81d/tt_metal/hw/inc/api/compute/eltwise_unary/exp.h#L45-L72
- https://github.com/tenstorrent/tt-metal/blob/e908c31332b60860ed0d4186452dc880cdd5a81d/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_exp.h#L641-L715
- https://github.com/tenstorrent/tt-metal/blob/e908c31332b60860ed0d4186452dc880cdd5a81d/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_exp.h#L480-L513

The simulator independently discards the argument and asserts that
`iterations=4` returns exact `torch.exp`:

- https://github.com/tenstorrent/tt-lang/blob/823f555e4944c92ef236744a191009ce6ff663df/python/sim/math.py#L199-L220
[...]

PR #976, opened 2026-08-28
[ttl.exp] reject unsupported iterations in exp

vtsilytskyiTT:
### Problem description
No validation for iterations parameter in exp.

### What's changed

1. Allow only 8 as a valid parameter.
An alternative would be to entirely remove parameter. I decided to keep it, because in future we may decide to add 32, which is also valid in some rare cases.
2. `dprint/runtime.py` checks now use CHECK-DAG with runtime thread prefixes.

### Ticket
[[ttl.exp] Reject unsupported iterations before kernel lowering](https://github.com/tenstorrent/tt-lang/issues/819)

### Checklist
- [X] New & Existing tests provide coverage for changes

1. cmake --build build --target check-ttlang check-ttlang-python-lit
2. pytest test/sim/test_math.py
3. python test/python/invalid/invalid_exp_iterations.py (should fail)

### DIFF
--- a/docs/sphinx/specs/TTLangSpecification.md
+++ b/docs/sphinx/specs/TTLangSpecification.md
@@ -1424,7 +1424,7 @@ def matmul_read():
 | `ttl.BlockExpr.__abs__(self) -> ttl.BlockExpr`<br><br>`ttl.math.abs(expr: ttl.BlockExpr) -> ttl.BlockExpr` | Absolute value. Example: `abs(a)`, `ttl.math.abs(a)`. |
 | `ttl.BlockExpr.__neg__(self) -> ttl.BlockExpr`<br><br>`ttl.math.neg(expr: ttl.BlockExpr) -> ttl.BlockExpr` | Negation. Example: `-a`, `ttl.math.neg(a)`. |
 | `ttl.BlockExpr.__pow__(self, exponent: ttl.NaturalInt) -> ttl.BlockExpr`<br><br>`ttl.math.pow(expr: ttl.BlockExpr, exponent: ttl.NaturalInt) -> ttl.BlockExpr` | Power with scalar unsigned integer exponent. Example; `a ** 2`, `ttl.math.pow(a, 2)`. |
-| `ttl.math.exp(expr: ttl.BlockExpr, *, approx: bool = False, scale: Optional[float] = None, skip_clamp_check: bool = False, iterations: ttl.PositiveInt = 8) -> ttl.BlockExpr` | Natural base exponential. Computes `e^x` by default. When `scale` is provided, computes `e^(scale * x)`. |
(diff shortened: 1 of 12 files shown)


docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1440,8 +1440,9 @@ flag enables approximate exponential evaluation. The `scale` argument applies
 a compile-time scalar multiplier before the exponential. The
 `skip_clamp_check` flag disables clamping of very negative inputs in
 approximate mode. This is faster, but inputs below approximately `-88.5` can
-produce incorrect negative outputs. The `iterations` argument controls the
-number of SFPU lane iterations and defaults to 8.
+produce incorrect negative outputs. The `iterations` argument is retained for
```
brnorris03: The language specification should state valid programs, not compiler stages or TT-Metal API compatibility. State that `iterations` must be 8 and other values are invalid; keep the rationale in the PR description.

test/python/dprint/runtime.py
```diff
@@ -141,8 +141,8 @@ def dm_write():
 # CHECK-DAG: cb_id
 # CHECK-DAG: bf16 dm hello
 # CHECK-DAG: core:
-# CHECK-DAG: math thread
-# CHECK-DAG: unpack thread
+# CHECK-DAG: TR1: math thread
```
brnorris03: This dprint expectation change is unrelated to exp iteration validation. Please move it to a separate PR.

docs/sphinx/specs/TTLangSpecification.md
```diff
@@ -1440,8 +1440,9 @@ flag enables approximate exponential evaluation. The `scale` argument applies
...
+hardware API parity, but only 8 is supported because this count evaluates one
+complete TT-Lang tile. Other values are rejected before kernel lowering.
```
brnorris03, line 1445: I was a bit confused here, this seems to imply no subtile support? But `iterations=8` does not mean "one complete TT-Lang tile", right? It means eight 32-lane SFPU iterations per face. For example, the existing subtile exp tests use `8` for every supported tile size from `1x32` through `32x32`.

Looking at the LLKs: some `exp` implementations are hard-unrolled for eight iterations and don't seem to accept other values, so I think it would be better to describe `8` as the only currently supported value, not (implied) as a logical tile size. Also probably at some point the compiler should derive this from the physical tile and target instead of exposing it as a user parameter.

test/python/dprint/runtime.py
```diff
@@ -155,11 +155,11 @@ def dm_write():
 # XXX: === after add ===
 # XXX: DST[0] (ttl.tile_add)
 # XXX: DST[1] (ttl.tile_exp)
-# CHECK: pack thread
+# CHECK-DAG: TR2: pack thread
 
 # Tensor pages print in dm_write (bf16 page from L1 CB)
 # The page output starts with page number followed by BF16 values.
-# CHECK: 0:
+# CHECK-DAG: BR: 0:
```
brnorris03, line 162: seems like these are not related to this PR? Perhaps revert?

brnorris03 (approved):
A couple of minor things, otherwise LGTM!

PR #977, opened 2026-08-28
[ttl][pipes] Optimize static PipeNet reduction lowering

brnorris03:
### Problem description

Static PipeNet reductions retain runtime record selection, cumulative synchronization state, full-block transfers, and repeated packer configuration even when topology and execution counts are compile-time constants. These operations dominate small-reduction latency and prevent compact row-prefix results from remaining in DST through the reduction.

### What's changed

~3,000 LOC implementation.

- Add per-node PipeNet destination counts and byte-counted DFB/PipeNet transfers.
- Restrict local PipeNet analysis to each allocation domain and index local records without receiver-address tables.
- Plan static PipeNet transfers before rewriting, unroll constant record loops, batch static receives, and lower one-shot point-to-point transfers to posted one-packet payload and completion writes.
- Add Blackhole contiguous block packing and direct BF16/FP32 full-DST-tile to compact row-prefix DFB packing.
- Preserve type-changing accumulation state in DST, reuse compatible row-pack configuration across loops, and support one receiver-only reduction output.
- Preserve constant-table folding for execution-count analysis while retaining canonicalization.

### Stack

This draft contains the original combined implementation. The reviewable component stack begins at #979 and separates the same responsibilities into independently testable compiler layers.

### Performance

The [TreeReduce benchmark](https://github.com/tenstorrent/tt-lang-ops-and-models/pull/55) measures TT-Lang at 2.232963 us versus 2.246296 us for the manually optimized comparator: ratio 0.994064 with interval [0.989657, 0.996854] across 100 balanced pairs. The measured BF16 outputs are bit-identical.

### Checklist


brnorris03:
Superseded by the reviewable component stack rooted at #979 and ending at #992. This combined draft will not merge.

PR #978, opened 2026-08-29
[ttl][pipes] Optimize static fabric transfer dispatch [fabric-4]

brnorris03:
### Problem description

Static one-shot fabric transfers retain runtime routing, readiness, and program-placement work even when their destinations and execution counts are compile-time constants. Multi-link routes also use one connection, limiting available fabric bandwidth.

### What's changed

~1,200 LOC implementation.

- Add explicit `mesh_program_placements` for sparse program descriptor materialization while preserving full-domain placement by default.
- Preconfigure one-packet fabric writes, use posted payload submission, and omit receiver readiness only for one-shot fabric transfers with computed destinations.
- Bind pairs of same-direction connections and stripe eligible payloads across two links while tracking completion by exact byte count.
- Preserve the standard readiness protocol and single-connection behavior when the static conditions are not proven or only one link is available.

For example, an operation with a 4 x 8 device domain can select individual coordinates or inclusive `ttl.MeshProgramPlacement` ranges instead of materializing a program descriptor on every device.

### Stack

This draft is stacked on #977. The replacement component stack isolates sparse placement in #990 and the other responsibilities in their respective reviewed layers.

### Performance

The complete generated [opposite-corner D2D benchmark stack](https://github.com/tenstorrent/tt-lang-ops-and-models/pull/55) measures 16.284815 us for TT-Lang versus 27.272963 us for the manually optimized comparator, a 0.597105 ratio. The measured BF16 outputs are bit-identical.

### Checklist


brnorris03:
Superseded by the component stack rooted at #979, including sparse placement in #990 and final integration through #992. This combined draft will not merge.

PR #1038, opened 2026-09-12
[ttl] Share equivalent core-specialized kernel descriptors

brnorris03:
### Problem description

`--ttl-specialize-cores` creates one TTKernel function per launch coordinate so later passes can remove coordinate-dependent control flow. The Python kernel builder registered every resulting function separately, including functions with identical code and descriptor requirements. A `12x10` ReduceToAll specialization produced 360 functions for 20 distinct kernel contracts; constructing programs for 32 devices attempted 11,520 TT-Metal kernel registrations and did not reach device launch within 25 minutes.

### What's changed

This PR separates per-coordinate specialization from TT-Metal descriptor registration. The Python kernel builder snapshots each function's complete descriptor contract before EmitC conversion, translates the TTKernel module once in function order, and registers one descriptor for specialized functions whose generated C++ and immutable runtime requirements are identical.

Descriptor comparison includes runtime argument metadata and every function attribute except the symbol name and assigned launch coordinates. Processor and coordinate ownership is validated independently of descriptor equivalence, so conflicting assignments fail before runtime resource planning. Canonical TTL constants now define the launch-grid and specialized-coordinate attributes used across the compiler and Python bindings.

For a `2x2` grid, a reader whose only coordinate-dependent branch tests `x` still produces four specialized functions. The two `x=0` functions share one descriptor for `{(0,0), (0,1)}`, and the two `x=1` functions share another for `{(1,0), (1,1)}`.

### Validation

Unit and lit coverage verifies ordered batch translation, complete descriptor metadata, mixed specialized and whole-grid functions, descriptor grouping, and invalid processor-coordinate assignments. Blackhole tests cover BF16 and f32 with DRAM and L1 tensors, six specialized functions becoming four descriptors, and correct first execution from the emitted runner.


lib/Dialect/TTKernel/Transforms/TTKernelSpecializeCores.cpp
```diff
@@ -42,10 +43,9 @@ namespace {
 // Attribute names. These are part of the frontend / runtime contract and keep
 // the `ttl.` prefix even though this pass runs at the TTKernel level:
 // `ttl.launch_grid` (the launch extent) is set on the module by the Python
-// frontend, and `ttl.core_coord` is read back by the ttnn runtime bridge for
-// dispatch.
+// frontend, and Python kernel construction reads `ttl.core_coord` to assign
+// each specialized function's launch coordinates.
 constexpr llvm::StringLiteral LaunchGridAttrName = "ttl.launch_grid";
```
jackzhangTT: if it's not too much to ask could I ask you to also do the same for this variable like you did with `CoreCoordAttrName` for consistency while you're already here

python/ttlang/TTKernelModule.cpp
jackzhangTT, line 27: should this also be wrapped like below? 

```
.def_static("cast", [](MlirType &ty) {
  return wrap(mlir::cast<tt::ttkernel::CBType>(unwrap(ty)));
});
```

jackzhangTT (approved):
Just some minor comments rest look good! Thanks for cleaning up the hardcoded variable names!

PR #1061, opened 2026-09-18
[fix] Fix routing-plane packet header include order (post-uplift fix)

brnorris03:
### Problem description
Routing-plane kernels include `fabric_config.h` before `PACKET_HEADER_TYPE` is defined, causing fabric kernel compilation to fail with the v0.79 toolchain.

### What's changed
Add the packet-header definition as a direct routing-plane dependency and verify the required include order.

### Checklist
- [x] Existing tests provide coverage for changes


vtsilytskyiTT (approved):
Looks good.

PR #1063, opened 2026-09-18
[sim] Add a compiler-backed tt-emule backend

FConstantinos:
### Problem description

The Python simulator and the compiler/runtime stack can evolve independently.
A kernel that passes in the Python simulator is not guaranteed to match what
the compiler generates and tt-metal executes. This limits its usefulness as
an oracle for developing or generating kernels without silicon.

This PR adds an opt-in path that uses the real TT-Lang compiler and executes
its generated kernels through tt-metal and tt-emule. The Python backend remains
the default. No compiler transformations or emulator implementation are changed.

### What's changed

- `./bin/tt-lang-sim --backend=emule PROGRAM.py` runs on one emulated Blackhole
  P150 through a Linux/amd64 Docker environment.
- `./scripts/install-tt-lang-emule.sh` installs the environment separately.
  Normal execution does not download dependencies or build the compiler.
- A repository manifest pins the emulator, Metal, base image and target, with
  a required compiler baseline. Independent dependency/manifest overrides are
  rejected; the emulator repository location only locates the pinned source.
- Installation checks source revisions, target availability and LLVM
  compatibility. Execution rejects an absent or stale compiler installation.
- Docker integration preserves working-directory paths, stdin and linked Git
  worktrees. Image provenance excludes emulator source credentials.
- Existing pytest/lit device detection and environment propagation recognize
  emulation. A lightweight PR workflow runs stack and launcher regression tests.
- Installation, normal test commands, a compiler-backed smoke example and
  known limitations are documented.

#### Install and run

```bash
export TTLANG_EMULE_RUNTIME_SOURCE_URL=REPOSITORY_URL
./scripts/install-tt-lang-emule.sh
./bin/tt-lang-sim --backend=emule examples/eltwise_add.py
```

The approved repository URL is supplied by the tt-emule owners. This backend
requires a source checkout and Docker with Linux/amd64 support; it is not
included in the PyPI simulator wheel. See the
[getting-started guide](https://github.com/tenstorrent/tt-lang/blob/kostas/tt-lang-sim-emule-consolidated/docs/sphinx/simulator-getting-started.md).

#### Validation

Split validation at `5b10a58e`, September 21, 2026:

- Host simulator/focused packaging pytest: 1,195 passed, 76 skipped, 2 xfailed.
- Launcher/container BATS: 62 passed.
- Docker `check-ttlang`: 528 IR, 5 binding and 193 packaging tests passed.
- Installation and real emulator execution of `compiler_only_external_call.py`
  and `eltwise_add.py`: passed, including output checks.
- All pre-commit hooks, Bash syntax and whitespace checks passed.

Integration validation used the existing pinned image in an isolated build
volume on macOS. The full device suite and a fresh Metal/runtime image build
were not rerun for this split. The earlier full-suite run had 165 failures;
known gaps include low-level host-JIT incompatibilities, a missing RMSNorm
header, incorrect results in some buffer/collective cases and missing DPRINT
output. There is no claim of full compiler-suite, full-model or multi-device
support, nor that all remaining failures belong to tt-emule.

Dependency-update tooling, experimental configurations and the detailed
[historical report](https://github.com/tenstorrent/tt-lang/blob/kostas/tt-lang-sim-emule-maintenance/docs/sphinx/compiler-emule-test-status.md)
are reviewed separately in #1072.

### Checklist

- [x] New and existing tests cover the retained backend and failure paths.
- [x] Installation, execution, testing and known limitations are documented.


brnorris03:
Suggest restacking this onto #1043 so the actual new changes are visible/reviewable, thanks!

FConstantinos:
> Suggest restacking this onto #1043 so the actual new changes are visible/reviewable, thanks!

After discussing this offline, I understand that this comment was aimed at a different PR.

scripts/install-tt-lang-emule.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit 2
+fi
+
+_SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+_REPO_ROOT="$(dirname "$_SCRIPT_DIR")"
+readonly _SCRIPT_DIR _REPO_ROOT
+
+export TTLANG_EMULE_INSTALL=1
+exec "${_SCRIPT_DIR}/tt-lang-emule-container.sh" \
+    "${_REPO_ROOT}/examples/compiler_only_external_call.py"
```
brnorris03: The entrypoint exits after the build in install mode, so this example is never run and installation reports success without executing a kernel. Suggest running it after the build so the installer verifies emulator execution -- or drop the argument and stop requiring a script in install mode.

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+_HOST_CWD="$(pwd -P)"
+_SCRIPT_ABSOLUTE="$(realpath "$_SCRIPT_ARGUMENT")"
+_SCRIPT_DIR_HOST="$(dirname "$_SCRIPT_ABSOLUTE")"
+_SCRIPT_BASENAME="$(basename "$_SCRIPT_ABSOLUTE")"
+
+_RUN_ARGS=(
+    run
+    --rm
+    --platform "$_PLATFORM"
+    --entrypoint /workspace/scripts/tt-lang-emule-entrypoint.sh
```
brnorris03: This overrides the entrypoint copied into the image, so the image copy never runs; the entrypoint is still an image-identity input (line 74), so any edit to it forces a tt-metal rebuild. Suggest keeping one: run the image copy, or stop copying it and drop it from the identity inputs.

scripts/tt-lang-emule-entrypoint.sh
```diff
@@ -0,0 +1,156 @@
...
+    exit 2
+fi
+if [ ! -f "$1" ]; then
+    echo "tt-lang emule container: script not found: $1" >&2
+    exit 2
+fi
+
+export TT_METAL_EMULE_MODE=1
+export TT_METAL_SLOW_DISPATCH_MODE=1
+export TT_METAL_MOCK_CLUSTER_DESC_PATH="${TT_METAL_MOCK_CLUSTER_DESC_PATH:-${CLUSTER_DESCRIPTORS}/blackhole_P150_unharvested.yaml}"
```
brnorris03: These defaults duplicate the manifest target, and the launcher always sets them. I think it's better to require them and fail when missing, so the manifest stays the only source.

docs/sphinx/simulator-getting-started.md
```diff
@@ -0,0 +1,241 @@
...
+the installer:
+
+```bash
+runtime_image=IMAGE_NAME
+compiler_volume=BUILD_VOLUME_NAME
+runtime_cache_volume=CACHE_VOLUME_NAME
+source_dir="$(pwd -P)"
+git_dir="$(git rev-parse --path-format=absolute --git-common-dir)"
+
+docker run --rm -it --platform linux/amd64 \
```
brnorris03: This restates the launcher's mounts and eight environment settings and will drift from the manifest. Would be nice to add a launcher shell mode that reuses `_RUN_ARGS` and document that one command here.

scripts/tt-lang-emule-stack.py
```diff
@@ -0,0 +1,215 @@
...
+        "TTLANG_EMULE_BASE_IMAGE": base_image,
+        "TTLANG_EMULE_PLATFORM": require_string(runtime, "platform", "runtime"),
+        "TTLANG_EMULE_TARGET": require_string(target, "name", "target"),
+        "TTLANG_EMULE_CLUSTER_DESCRIPTOR": descriptor,
+        "TTLANG_EMULE_MESH_DEVICE": require_string(target, "mesh_device", "target"),
+        "TTLANG_EMULE_ALLOCATOR_MODE": require_string(
+            target, "allocator_mode", "target"
+        ),
+    }
+    if values["TTLANG_EMULE_PLATFORM"] != "linux/amd64":
```
brnorris03: `platform`, `allocator_mode` and `repository_environment` each accept exactly one value, and `TTLANG_EMULE_REPOSITORY` is always emitted empty, so the fallback at container line 57 never applies. Suggest removing them from the manifest and keeping the constants in the launcher -- `allocator_mode` is also unused (the launcher hard-codes `TT_METAL_ALLOCATOR_MODE_HYBRID=1`).

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+# Run a TT-Lang program through the compiler and an emule-enabled tt-metal
+# runtime. macOS uses Docker's linux/amd64 virtualization because tt-emule and
+# tt-metal are Linux/x86-64 components.
+
+set -euo pipefail
+
+_SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+_REPO_ROOT="$(dirname "$_SCRIPT_DIR")"
+readonly _SCRIPT_DIR _REPO_ROOT
+for _OVERRIDE in \
```
brnorris03: Nothing reads these six variables, so this block and its two BATS cases only preserve an earlier interface. Suggest removing them, along with the `unset` in the BATS `setup()`.

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+
+_BUILD_IMAGE=0
+if [ "${TTLANG_EMULE_REBUILD:-0}" = "1" ]; then
+    if [ "${TTLANG_EMULE_INSTALL:-0}" != "1" ]; then
+        echo "tt-lang-sim: runtime rebuilding is only available during installation." >&2
+        echo "Run scripts/install-tt-lang-emule.sh instead." >&2
+        exit 1
+    fi
+    _BUILD_IMAGE=1
+elif _IMAGE_INSPECT_ERROR="$("$_DOCKER" image inspect "$_IMAGE" 2>&1 >/dev/null)"; then
```
brnorris03: Matching Docker's error text is version-dependent, and Podman differs. `docker images -q "$_IMAGE"` prints nothing for a missing image and fails only on daemon errors, which removes the parsing and its four tests.

.github/scripts/tests/test_ttlang_emule_container.bats
```diff
@@ -0,0 +1,956 @@
...
+EOF
+    chmod +x "$target"
+}
+
+setup() {
+    MOCK_DOCKER="$BATS_TEST_TMPDIR/docker"
+    MOCK_DOCKER_LOG="$BATS_TEST_TMPDIR/docker.log"
+    export MOCK_DOCKER_LOG
+    make_mock_docker "$MOCK_DOCKER"
+    unset TTLANG_EMULE_STACK_MANIFEST TTLANG_EMULE_RUNTIME_COMMIT \
```
brnorris03: The entrypoint case at line 825 fails when the caller has `TT_METAL_CACHE` set. Suggest unsetting every variable the entrypoint defaults here: `TT_METAL_CACHE`, `TT_EMULE_JIT_CACHE_DIR`, `MESH_DEVICE`, `EMULE_FABRIC8`, `TT_METAL_ALLOCATOR_MODE_HYBRID`, `TT_METAL_MOCK_CLUSTER_DESC_PATH`.

test/sim/test_examples.py
```diff
@@ -192,6 +192,24 @@ def test_example_cli(script_name: str, scheduler: str) -> None:
...
+) -> None:
+    """The Python simulator rejects the compiler's external-C++ escape hatch."""
+    script = EXAMPLES_DIR / "compiler_only_external_call.py"
+    code, out = run_script_in_process(script, scheduler)
+
+    assert code != 0, (
+        "compiler_only_external_call.py unexpectedly ran on the Python "
+        f"simulator. Output:\n{out}"
+    )
+    assert "no attribute 'call_extern_func'" in out, (
```
brnorris03: This asserts an `AttributeError`, so the test depends on the API being absent rather than rejected. Suggest a simulator `call_extern_func` that raises a message naming the compiler backend, and assert that message.

examples/compiler_only_external_call.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Add two tensors while calling user-provided C++ from a compiled kernel.
+
+The external marker is intentionally a no-op: the tensor golden checks the
+normal kernel path while the generated C++ proves that ``ttl.call_extern_func``
```
brnorris03: `compiler_only_marker` is empty and nothing inspects the generated C++, so the example passes whether or not the call is emitted. Suggest giving the marker an effect the golden check observes, or removing this claim.

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+
+readonly _TT_EMULE_COMMIT="$_MANIFEST_EMULE_COMMIT"
+readonly _TT_METAL_COMMIT="$_MANIFEST_METAL_COMMIT"
+readonly _TT_EMULE_SOURCE_URL="${TTLANG_EMULE_RUNTIME_SOURCE_URL:-$_MANIFEST_EMULE_REPOSITORY}"
+readonly _TT_METAL_SOURCE_URL="$_MANIFEST_METAL_REPOSITORY"
+readonly _BASE_IMAGE="$_MANIFEST_BASE_IMAGE"
+readonly _REQUIRED_EMULE_FILE="$_MANIFEST_CLUSTER_DESCRIPTOR"
+readonly _PLATFORM="$_MANIFEST_PLATFORM"
+
+for _COMMIT in "$_TT_EMULE_COMMIT" "$_TT_METAL_COMMIT"; do
```
brnorris03: `require_sha` in the stack tool already enforces this. Suggest removing the duplicate check.

scripts/tt-lang-emule-entrypoint.sh
```diff
@@ -0,0 +1,156 @@
...
+    exit 1
+fi
+
+# activate is generated for bash and may legitimately reference variables that
+# are absent from a non-interactive container shell.
+set +u
+source "${TTLANG_BUILD_DIR}/env/activate"
+set -u
+
+export TT_METAL_EMULE_MODE=1
```
brnorris03: `activate` does not modify any of these, so the re-export has no effect. Suggest removing the block.

test/python/conftest.py
```diff
@@ -107,7 +106,7 @@ def ttnn_device():
         _ttnn_import_failed = True
         raise
 
-    if os.environ.get("TT_METAL_SIMULATOR"):
+    if os.environ.get("TT_METAL_SIMULATOR") or os.environ.get("TT_METAL_EMULE_MODE"):
```
brnorris03: The same emule-or-simulator condition appears in `ttlang_test_utils.py` as two branches with one result. Suggest one helper there, used here as well.

docs/sphinx/simulator.md
```diff
@@ -43,6 +62,99 @@ If you have already built the full TT-Lang compiler (`source build/env/activate`
...
+
+The Docker backend uses the same launcher interface as the Python backend.
+The [Docker getting-started guide](simulator-getting-started.md) describes
+installation. From an installed source checkout:
+
+```bash
+./bin/tt-lang-sim --backend=emule examples/eltwise_add.py
+```
+
+Install the complete supported compiler, emulator, and tt-metal environment
```
brnorris03: This repeats the install, validation and provenance text from the getting-started guide. Suggest keeping it there and linking from here.

scripts/tt-lang-emule-entrypoint.sh
```diff
@@ -0,0 +1,156 @@
...
+fi
+
+export TT_METAL_EMULE_MODE=1
+export TT_METAL_SLOW_DISPATCH_MODE=1
+export TT_METAL_MOCK_CLUSTER_DESC_PATH="${TT_METAL_MOCK_CLUSTER_DESC_PATH:-${CLUSTER_DESCRIPTORS}/blackhole_P150_unharvested.yaml}"
+export TT_METAL_ALLOCATOR_MODE_HYBRID="${TT_METAL_ALLOCATOR_MODE_HYBRID:-1}"
+export EMULE_FABRIC8="${EMULE_FABRIC8:-1}"
+export TT_METAL_CACHE="${TT_METAL_CACHE:-/tt-metal-cache}"
+export TT_EMULE_JIT_CACHE_DIR="${TT_EMULE_JIT_CACHE_DIR:-${TT_METAL_CACHE}/emule-jit}"
+export MESH_DEVICE="${MESH_DEVICE:-P150}"
```
brnorris03: These defaults duplicate the manifest target, and the launcher always sets them. I think it's better to require them and fail when missing, so the manifest stays the only source.

docs/sphinx/simulator-getting-started.md
```diff
@@ -0,0 +1,241 @@
...
+  --workdir /workspace \
+  --env TT_METAL_EMULE_MODE=1 \
+  --env TT_METAL_SLOW_DISPATCH_MODE=1 \
+  --env TT_METAL_MOCK_CLUSTER_DESC_PATH=/opt/tt-emule/cluster_descriptors/blackhole_P150_unharvested.yaml \
+  --env TT_METAL_ALLOCATOR_MODE_HYBRID=1 \
+  --env EMULE_FABRIC8=1 \
+  --env TT_METAL_CACHE=/tt-metal-cache \
+  --env TT_EMULE_JIT_CACHE_DIR=/tt-metal-cache/emule-jit \
+  --env MESH_DEVICE=P150 \
+  --entrypoint /bin/bash "$runtime_image" --noprofile --norc
```
brnorris03: Lines 123-135 restate the mounts that `scripts/tt-lang-emule-container.sh` builds in `_RUN_ARGS` (lines 195-210) and the environment that `tt-lang-emule-entrypoint.sh` exports (lines 24-31), including the P150 target that comes from the manifest. Suggest a launcher option that opens a shell with the same arguments -- then this section becomes one command and cannot drift.

scripts/tt-lang-emule-stack.py
```diff
@@ -0,0 +1,215 @@
...
+        raise StackError("runtime.platform must be linux/amd64")
+    if values["TTLANG_EMULE_ALLOCATOR_MODE"] != "hybrid":
+        raise StackError("target.allocator_mode must be hybrid")
```
brnorris03: `platform`, `allocator_mode` and `repository_environment` each accept exactly one value, and `TTLANG_EMULE_REPOSITORY` is always emitted empty, so the fallback at container line 57 never applies. Suggest removing them from the manifest and keeping the constants in the launcher -- `allocator_mode` is also unused (the launcher hard-codes `TT_METAL_ALLOCATOR_MODE_HYBRID=1`).

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+    TTLANG_EMULE_RUNTIME_COMMIT \
+    TTLANG_EMULE_RUNTIME_METAL_COMMIT \
+    TTLANG_EMULE_RUNTIME_METAL_SOURCE_URL \
+    TTLANG_EMULE_RUNTIME_BASE_IMAGE \
+    TTLANG_EMULE_PLATFORM; do
+    if [ -n "${!_OVERRIDE:-}" ]; then
+        echo "tt-lang-sim: ${_OVERRIDE} is not supported; use the repository's pinned stack manifest." >&2
+        exit 2
+    fi
+done
```
brnorris03: Nothing reads these six variables, so this block and its two BATS cases only preserve an earlier interface. Suggest removing them, along with the `unset` in the BATS `setup()`.

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+    _IMAGE_IS_MISSING=0
+    if [ "$_IMAGE_INSPECT_STATUS" -eq 1 ]; then
+        case "$_IMAGE_INSPECT_ERROR" in
+            "Error response from daemon: No such image: ${_IMAGE}"*|\
+            "Error response from daemon: {\"message\":\"No such image: ${_IMAGE}\"}"|\
+            "{\"message\":\"No such image: ${_IMAGE}\"}")
+                _IMAGE_IS_MISSING=1
+                ;;
+        esac
+    fi
```
brnorris03: Matching Docker's error text is version-dependent, and Podman differs. `docker images -q "$_IMAGE"` prints nothing for a missing image and fails only on daemon errors, which removes the parsing and its four tests.

examples/compiler_only_external_call.py
```diff
@@ -0,0 +1,94 @@
...
+    row_tiles = a_in.shape[0] // TILE_SIZE
+    col_tiles = a_in.shape[1] // TILE_SIZE
+
+    a_dfb = ttl.make_dataflow_buffer_like(a_in, shape=(1, 1), block_count=2)
+    b_dfb = ttl.make_dataflow_buffer_like(b_in, shape=(1, 1), block_count=2)
+    out_dfb = ttl.make_dataflow_buffer_like(out, shape=(1, 1), block_count=2)
+
+    @ttl.compute()
+    def compute():
+        ttl.call_extern_func(EXTERNAL_HEADER, "compiler_only_marker")
```
brnorris03: `compiler_only_marker` is empty (`compiler_only_marker.hpp`) and nothing inspects the generated C++, so this example passes whether or not the call is emitted. The docstring (lines 7-9) says the generated C++ proves it was lowered -- suggest giving the marker an effect the golden check observes, or removing that claim.

scripts/tt-lang-emule-container.sh
```diff
@@ -0,0 +1,389 @@
...
+readonly _BASE_IMAGE="$_MANIFEST_BASE_IMAGE"
+readonly _REQUIRED_EMULE_FILE="$_MANIFEST_CLUSTER_DESCRIPTOR"
+readonly _PLATFORM="$_MANIFEST_PLATFORM"
+
+for _COMMIT in "$_TT_EMULE_COMMIT" "$_TT_METAL_COMMIT"; do
+    if [ "${#_COMMIT}" -ne 40 ] || [[ "$_COMMIT" == *[!0-9a-f]* ]]; then
+        echo "tt-lang-sim: emulator revisions must be full lowercase commit SHAs." >&2
+        exit 2
+    fi
+done
```
brnorris03: `require_sha` in the stack tool already enforces this. Suggest removing the duplicate check.

scripts/tt-lang-emule-entrypoint.sh
```diff
@@ -0,0 +1,156 @@
...
+
+export TT_METAL_EMULE_MODE=1
+export TT_METAL_SLOW_DISPATCH_MODE=1
+export TT_METAL_MOCK_CLUSTER_DESC_PATH
+export TT_METAL_ALLOCATOR_MODE_HYBRID
+export EMULE_FABRIC8
+export TT_METAL_CACHE
+export TT_EMULE_JIT_CACHE_DIR
+export MESH_DEVICE
+unset TTLANG_COMPILE_ONLY TTLANG_SIM_ONLY
```
brnorris03: `activate` does not modify any of these, so the re-export has no effect. Suggest removing the block.

docs/sphinx/simulator.md
```diff
@@ -43,6 +62,99 @@ If you have already built the full TT-Lang compiler (`source build/env/activate`
...
+the image is built. These records identify the supported manifest and exact
+runtime inputs used to build the image. Inspect an artifact without running a
+workload with:
+
+```bash
+docker image inspect tt-lang-emule:TAG \
+  --format '{{json .Config.Labels}}'
+docker run --rm --entrypoint cat tt-lang-emule:TAG \
+  /opt/tt-emule-runtime/stack.json
+```
```
brnorris03: This repeats the install, validation and provenance text from the getting-started guide. Suggest keeping it there and linking from here.

docs/sphinx/simulator-getting-started.md
```diff
@@ -0,0 +1,241 @@
...
+docker info
+```
+
+## Obtain TT-Lang
+
+Start from a TT-Lang source checkout containing the compiler-backed backend
+and its submodules:
+
+```bash
+git clone --recurse-submodules https://github.com/tenstorrent/tt-lang.git
```
brnorris03: I don't encourage general recurse submodules -- see other docs for recommended shallow clone. This takes a very long time (with llvm and tt-metal submodules) and consumes a lot of space needlessly.

docs/sphinx/simulator-getting-started.md
```diff
@@ -0,0 +1,241 @@
...
+llvm-lit -v /ttlang-build/test/python/simple_add.py
+```
+
+These commands use the runtime configured for that compiler build. Compiler-only
+tests exercise compiler behavior; device execution tests exercise tt-emule in
+the installed Linux environment. See
+[`test/TESTING.md`](https://github.com/tenstorrent/tt-lang/blob/main/test/TESTING.md)
+for the suite boundaries, device requirements, pytest selection, lit paths, and
+output locations. See [Testing](testing.md) for the short command reference.
+
```
brnorris03: There is nothing unique (emulator-specific) -- I think instead of replicating the instructions (different versions may end up diverging), this should point to the existing build docs and just specify what build directory to use.

docs/sphinx/simulator.md
```diff
@@ -43,6 +62,99 @@ If you have already built the full TT-Lang compiler (`source build/env/activate`
 tt-lang-sim examples/eltwise_add.py
 ```
 
+### Compiler-backed emulation
+
+The Docker backend uses the same launcher interface as the Python backend.
+The [Docker getting-started guide](simulator-getting-started.md) describes
+installation. From an installed source checkout:
+
```
brnorris03: what is the "Docker backend"? Above there are two backends described -- python and emule (I am guessing this should be emule)?

docs/sphinx/simulator.md
```diff
@@ -43,6 +62,99 @@ If you have already built the full TT-Lang compiler (`source build/env/activate`
...
+```bash
+brew install colima docker
+softwareupdate --install-rosetta --agree-to-license
+colima start --vm-type vz --vz-rosetta --cpus 8 --memory 12
+```
+
+The Colima command uses Rosetta to run amd64 containers in an Apple
+Virtualization.framework VM. If Rosetta cannot be installed, an x86-64 QEMU VM
+also works but is substantially slower and requires `brew install qemu`.
+
```
brnorris03: I would just point to the official docker website or this package, I don't think it's needed to document a specific tool -- to me this suggests that my existing non-colima docker may not work (which I am pretty sure is not the case).

python/sim/ttlang_sim.py
```diff
@@ -250,11 +250,24 @@ def main() -> None:
...
+    parser.add_argument(
+        "--backend",
+        choices=["python", "emule"],
+        default="python",
+        help=(
+            "Execution backend. Compiler-backed emule is available only from "
+            "a source checkout through ./bin/tt-lang-sim."
+        ),
+    )
+
```
brnorris03: Here and everywhere they are used -- I would define named constants for the backends in a common module and not have string literals everywhere (for "python", "emule").

python/sim/ttlang_sim.py
```diff
@@ -374,6 +387,12 @@ def main() -> None:
     args.target = first
     args.script_args = script_args
 
+    if args.backend == "emule":
+        parser.error(
+            "the emule backend requires a TT-Lang source checkout; "
+            "run ./bin/tt-lang-sim SCRIPT.py --backend emule"
```
brnorris03: Does this have to be run from the directory containing `bin/tt-lang-sim`? If so state that explicitly above if not already (right now it's mostly implied by this error and the example)

scripts/install-tt-lang-emule.sh
brnorris03: Please have your agent add BATS tests for all new shell scripts using existing project practices.

brnorris03 (commented):
Thank you for creating this! Mostly good, just some cleanup suggestions, biggest missing thing is no bash shell unit tests (BATS) for new scripts.

scripts/tt-lang-emule-container.sh
brnorris03: Correction: these are not an earlier interface -- #1072 deletes this block and both BATS cases and accepts the same names as overrides. The suggestion stands for a different reason: adding them here only to remove them in the next PR is churn, so suggest leaving them out of this PR.

brnorris03: Correction: these are not an earlier interface -- #1072 deletes this block and both BATS cases and accepts the same names as overrides. The suggestion stands for a different reason: adding them here only to remove them in the next PR is churn, so suggest leaving them out of this PR.

PR #1064, opened 2026-09-19
[ttl][pipes][fix] Require receiver pushes in the control context of their post

brnorris03:
### Problem description

Computed receiver addresses require the sender and receiver to advance through the same DFB slots. The sender advances once per receiver post, but the existing proof did not verify that each receiver push executes in the same dynamic control context as its post. A looped or conditionally executed push could therefore advance the receiver write pointer without advancing the sender's slot counter, causing later transfers to write the wrong block.

### What's changed

`PipeGraph::proveReceiverProducerStreams` now compares the receiver control context of each push with every post owned by the same reservation. A missing or different context rejects computed addressing and pipe-only producer classification. Local pipes then use published receiver addresses; fabric pipes report why computed addressing could not be proven.

Full-capacity pushes are not exempt. Advancing by one DFB capacity returns the write pointer to the same slot, but slot equality does not prove that the receive post and push execute the same number of times. The analysis therefore requires both operations to have the same known control context.

In this example, `ttl.copy` posts the receive and returns a request. `ttl.wait` completes that request before `ttl.cb_push` publishes the data. The invalid condition is that the post executes once while the push executes twice:

```mlir
%request = ttl.copy %pipe, %reserved : (...) -> !ttl.receive_request
ttl.wait %request : !ttl.receive_request
scf.for %iteration = %zero to %two step %one {
  ttl.cb_push %dst : <[1, 1], !ttcore.tile<32x32, f32>, 2>
}
```

### Checklist

- [x] New tests cover local published-address fallback, fabric rejection, and full-capacity pushes in a different receiver control context.


jackzhangTT (approved):
LGTM

FConstantinos (approved):

